PR #19, opened 2026-05-26
Render targets: build only enabled + auto build-on-enable

### LINKED ISSUE #18: Re-implement enabled-only target builds + build-on-enable (reverted from #7 — broke TargetControlTest)
## Goal (the feature we still want)
- Disabling a render target should **skip rebuilding it** on subsequent saves (save build time).
- Enabling a target with no current render should **auto fetch-or-build it** so it appears.

This was implemented in **PR #7** (`enabledTargets` filtering on the build requests + build-on-enable in `setTargetEnabled`, in `KonstructionViewModel`) and **deployed**, but has now been **reverted** (in PR #17, revert commit `2bbc6bd`) because it introduced a regression.

## Why it was reverted
PR #7 broke **`TargetControlTest`** — all 3 cases time out in `setupWithTwoTargets`, waiting for `globalThis.__konstructor.state.targets.length === 2` (the konstruction load → compile → `mergeTargets` → target-display population flow never completes). Confirmed by bisect: reverting #7's `KonstructionViewModel` makes the e2e suite pass; restoring it fails 3/3. (`ThemeAndKeymapTest.testThemeSwitching`'s 30s timeout may be the same issue.)

## Root cause — not yet pinpointed
Reading #7's full diff didn't reveal it: `onInfoChanged` still calls `mergeTargets` *first*, so target population *should* survive. The break is a subtle runtime/state interaction (likely the `enabledTargets` read / build-on-enable coroutine colliding with the load→compile flow). Pinpointing it needs runtime instrumentation (browser-console / bridge `state` dumps), not just code reading.

## How it slipped in (process)
#7 was merged over an already-red CI without inspecting its *own* CI run; the e2e suite — which catches exactly this — never gated it.

## Requirement for the redo
Re-implement on a **fresh PR that must pass `:e2e:test` green (especially `TargetControlTest`) before merge.** Local e2e now works on the dev box via `__EGL_VENDOR_LIBRARY_FILENAMES=/usr/share/glvnd/egl_vendor.d/50_mesa.json` (Mesa-EGL override for the Xvfb/NVIDIA crash), so the regression can be reproduced and debugged locally.

monkopedia-coder:
## Summary

Re-implements #7 (reverted in `2bbc6bd` for breaking e2e), this time with the root cause fixed and the full e2e suite verified green locally.

- **Build only enabled targets** — `requestKonstructs` (in `save()` and the `onInfoChanged` NEEDS_EXEC branch) now requests only the user-enabled targets, so disabling a target skips rebuilding it. Untracked targets default to enabled.
- **Build-on-enable** — enabling a previously-disabled target with no current render auto fetch-or-builds it (`konstructed()` if it exists, else `konstruct()`), deduped per-target via `inFlightEnables`, guarding a null service and skipping on disable.

## Root cause of the #7 regression

#7 broke `TargetControlTest` (and `SaveFlowTest`/`ThemeAndKeymapTest`): the `setupWithTwoTargets` wait for `state.targets.length === 2` timed out at 120s.

Targets are only discovered by the server's **execute** pass — `ExecuteTask` calls `script.listTargets()` and `render()` returns that list as the konstruction's targets. `set()` and `compile()` only re-map the *already-known* `info.targets`. So a **fresh** konstruction reaches its first `NEEDS_EXEC` with `info.targets` empty.

#7 filtered `requestKonstructs(info.targets)` to the enabled subset and, when that subset was empty, short-circuited to `UiState.DEFAULT` without requesting a build. At the first `NEEDS_EXEC` the unfiltered list is *already* empty, so #7 skipped the only build that discovers targets. `targetDisplayRepo.mergeTargets` then never saw any targets, `targetDisplays` stayed empty, and `state.targets` (sourced from it in `JsBridge`) never reached 2.

## How this avoids it

`buildTargets(known)` distinguishes "nothing discovered yet" from "everything disabled":

- `known` empty -> return an **empty list** (still request the build, so discovery runs);
- at least one enabled -> return the enabled subset;
- targets known but all disabled -> return `null` (settle the UI, don't spin).

Once targets are discovered and CLEAN, later edits carry a populated `info.targets`, so the enabled-filter then correctly skips disabled targets.

## Test plan

- [x] `./gradlew clean :e2e:test -Pe2e --tests "*TargetControlTest"` — all 3 cases pass
- [x] `./gradlew clean :e2e:test -Pe2e` — full suite green (0 failures / 0 errors); previously-regressing `TargetControlTest` 3/3, `SaveFlowTest` 1/1, `ThemeAndKeymapTest` 3/3 all pass
- [x] `:frontend:spotlessKotlinCheck` clean

Fixes #18

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/viewmodel/KonstructionViewModel.kt
+++ b/frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/viewmodel/KonstructionViewModel.kt
@@ -88,6 +88,10 @@ class KonstructionViewModel(
     private var konstructionService: KonstructionService? = null
     private var listenerKey: String? = null
 
+    // Targets currently being auto-fetched/built via setTargetEnabled, to avoid
+    // kicking off duplicate work for the same target.
+    private val inFlightEnables = mutableSetOf<String>()
+
     init {
         // Re-derive enabled-rendered targets whenever displays change
         viewModelScope.launch {
@@ -193,11 +197,19 @@ class KonstructionViewModel(
                     targetDisplayRepo.mergeTargets(info.targets.map { it.name })
                     when (info.dirtyState) {
                         DirtyState.NEEDS_EXEC -> {
-                            _state.value = UiState.EXECUTING
-                            try {
-                                ks.requestKonstructs(info.targets.map { it.name })
-                            } catch (_: Exception) {
+                            val all = info.targets.map { it.name }
+                            val targets = buildTargets(all)
+                            if (targets == null) {
+                                // Every known target is disabled — nothing to
+                                // build, settle the UI instead of spinning.
                                 _state.value = UiState.DEFAULT
+                            } else {
+                                _state.value = UiState.EXECUTING
+                                try {
+                                    ks.requestKonstructs(targets)
+                                } catch (_: Exception) {
+                                    _state.value = UiState.DEFAULT
+                                }
                             }
                         }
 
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
Re-implementation of #18 reviewed to merge standard: buildTargets() empty-known case still triggers the discovery build (the #7 regression cannot recur), build-on-enable guards sound, scope is the single ViewModel. CI green (run 26465742265 — e2e under Xvfb incl. TargetControlTest passing). User signed off.

PR #21, opened 2026-05-30
Fix stale render from previous workspace/konstruction shown after switch

### LINKED ISSUE #12: Stale render from a previous workspace/konstruction shown after switch
## Symptom
After switching workspace/konstruction, a 3D model from a **different** workspace briefly appears loaded by accident, then goes away after a regenerate (a fresh build). Confirmed via investigation; matches the reported sighting.

## Root cause (high confidence)
`KonstructionViewModel` is a Koin **singleton** (one long-lived instance reused across every switch), so `onCleared()` — the only place that calls `ks.unregister(...)` — never fires.

`loadKonstruction` (`frontend/src/wasmJsMain/.../viewmodel/KonstructionViewModel.kt`) resets `_renderPath`/`renderPaths` and overwrites `konstructionService = ks` then `listenerKey = ks.register(...)`, but it **never unregisters the previous konstruction's listener first**, and the listener callbacks have **no `if (ks !== konstructionService) return` guard**.

Result: a late callback from the *old* konstruction — `onRenderChanged` / the `onInfoChanged`→CLEAN branch — still fires (its captured `ks` is the de-selected konstruction) and writes the old konstruction's full render URL (`model/{oldWs}/{oldKon}/renders/{target}.stl?v=N`) into the now-shared `renderPaths`/`_renderPath`/`enabledRenderedTargets`. The GL scene then loads the **old** workspace's mesh. A regenerate on the new konstruction emits fresh URLs that overwrite it → "goes away."

## Contributing factors
- **Auto-build (PR #7 / `feat/render-target-build-behavior`)** widens the race: `loadKonstruction`→auto-compile/build and `setTargetEnabled` kick off async work whose completion lands via the captured `ks`. Switching away from a konstruction with an in-flight build is the concrete trigger. `setTargetEnabled` also writes `renderPaths`/`_renderPath` from a coroutine without re-checking the current selection.
- **Renderer has no per-konstruction identity** (`threejs/ThreeJsRenderer.kt` `setTargets` reconciles by target *name*): it faithfully displays whatever name→URL map it's handed, so a stale write becomes a visible stale mesh until a freshened URL for that name overwrites it.
- **`inFlightEnables`** (singleton, keyed by target name, not cleared on switch) can silently drop a same-named enable right after a switch — minor, same subsystem.

## Ruled out (clean)
- Render URLs are unique per workspace+konstruction (`PathController` / `KonstructionServiceImpl`): no URL aliasing, so the renderer's URL cache can't serve the wrong mesh by collision — the leak is via stale *writes*, not URLs.
- `TargetDisplayRepository.activate(wsId, konId)` correctly re-scopes displays/colors per konstruction (no bleed).
- The workspace-selection cascade and the PR #6 renderer/layout changes are not involved.

## Fix direction
In `loadKonstruction`: `unregister` the previous service's listener (and cancel/ignore its in-flight callbacks) **before** wiring the new one; add a `ks !== konstructionService` guard in the listener callbacks so a late callback from a de-selected konstruction can't mutate the shared flows; clear `inFlightEnables` on switch.

## Repro
Open konstruction A in workspace 1, trigger (or catch mid-) a build, then immediately switch to konstruction B in another workspace — A's mesh flashes into B's scene until B renders. (Instrument: log the captured `ks` CallSign in `onRenderChanged`/`onInfoChanged` vs. the currently-selected konstruction.)

monkopedia-coder:
## Summary
Fixes #12 — after switching workspace/konstruction, a 3D model from a different workspace would briefly flash into the new scene until a fresh render overwrote it.

## Root cause
`KonstructionViewModel` is a Koin singleton reused across every switch, so `onCleared()` — the only caller of `ks.unregister(...)` — never fires. `loadKonstruction` wired a new listener (`konstructionService = ks`, `listenerKey = ks.register(...)`) without unregistering the previous konstruction's, and the listener callbacks had no identity guard. A late callback from the de-selected konstruction (`onRenderChanged` / the `onInfoChanged`→CLEAN branch — captured `ks` is the old konstruction) wrote the old render URL into the shared `renderPaths`/`_renderPath`/`enabledRenderedTargets` flows, so the renderer showed the previous workspace's mesh until a fresh render landed.

## Fix (scoped to `KonstructionViewModel.kt`)
- **Unregister the previous listener before wiring the new one** in `loadKonstruction`: capture the old service/key, null them out, then `oldKs.unregister(oldKey)` (exception-guarded).
- **Clear `inFlightEnables` on switch** so a leftover target-name entry can't suppress a legitimate enable after switching.
- **Identity guards** `if (ks !== konstructionService) return` at the top of the mutating listener callbacks (`onInfoChanged`, `onDirtyStateChanged`, `onTargetChanged`, `onRenderChanged`, `onContentChange`, `onTaskComplete`).
- **Post-await re-checks**: the CLEAN paths (in `loadKonstruction` and `onInfoChanged`) and the `setTargetEnabled` coroutine re-verify `ks === konstructionService` after the suspending `konstructed()` calls before writing the shared flows, closing the switch-during-await window.

Normal (non-switch) behavior is preserved: the first load has no previous listener to unregister, and `ks === konstructionService` holds throughout.

## Verification
- Full e2e suite (`clean :e2e:test -Pe2e`, Mesa-EGL override under Xvfb): **BUILD SUCCESSFUL**, 39 tests, 0 failures, 0 errors (6 pre-existing skips). Includes `TargetControlTest` (3/3) and the navigation/workspace/konstruction-switch coverage.
- `:frontend:spotlessKotlinCheck`: passing (Apache header + ktlint).

### DIFF
--- a/frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/viewmodel/KonstructionViewModel.kt
+++ b/frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/viewmodel/KonstructionViewModel.kt
@@ -104,6 +104,24 @@ class KonstructionViewModel(
             _state.value = UiState.LOADING
             _renderPath.value = null
             renderPaths.value = emptyMap()
+            // Drop any in-flight enable bookkeeping from the previous
+            // konstruction — keyed by target name on this singleton, a leftover
+            // entry could otherwise suppress a legitimate enable after switching.
+            inFlightEnables.clear()
+            // Unregister the previous konstruction's listener before wiring the
+            // new one, so a late callback from the de-selected konstruction can't
+            // keep mutating the now-shared flows. Capture the old refs first,
+            // clear them, then unregister.
+            val previousKs = konstructionService
+            val previousKey = listenerKey
+            konstructionService = null
+            listenerKey = null
+            if (previousKs != null && previousKey != null) {
+                try {
+                    previousKs.unregister(previousKey)
+                } catch (_: Exception) {
+                }
+            }
             targetDisplayRepo.activate(konstruction.workspaceId, konstruction.id)
             recomputeEnabledTargets()
             val service = serviceHolder.service.value
@@ -148,16 +166,20 @@ class KonstructionViewModel(
                             }
                         }.awaitAll()
                     }
-                    val newPaths = renderPaths.value.toMutableMap()
-                    for ((name, path) in paths) {
-                        if (path != null) {
-                            val fresh = freshen(path)
-                            newPaths[name] = fresh
-                            _renderPath.value = fresh
+                    // Re-check selection: a switch may have landed while the
+                    // konstructed() calls above were in flight.
+                    if (ks === konstructionService) {
+                        val newPaths = renderPaths.value.toMutableMap()
+                        for ((name, path) in paths) {
+                            if (path != null) {
+                                val fresh = freshen(path)
+                                newPaths[name] = fresh
+                                _renderPath.value = fresh
+                            }
                         }
+                        renderPaths.value = newPaths
+                        recomputeEnabledTargets()
                     }
-                    renderPaths.value = newPaths
-                    recomputeEnabledTargets()
                 } else {
                     autoCompileAndBuild(ks)
                 }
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
Tier-2 UX-path fix, user-signed-off. Stale-render leak on switch fixed via unregister-before-rewire + identity guards on listener callbacks; CI verified green. (User noted the implementation is functional but on the verbose/ugly side — tracking a cleanup follow-up separately.)

PR #30, opened 2026-06-02
Bump kodemirror 0.2.0 → 0.3.2 + enable editor soft-wrap

### LINKED ISSUE #10: Very long lines overflow the editor — no wrap, no horizontal scroll
## Symptom
Very long source lines **neither soft-wrap nor produce a horizontal scrollbar** — the content past the right edge just overflows/clips and is unreachable. (Initial guess was "weird wrapping"; corrected: it's not wrapping *or* scrolling.)

This looks like the **horizontal analog of the recently-fixed vertical overflow** (#6 / the Scaffold change): there, the editor wasn't given a bounded *height*, so kodemirror never scrolled vertically; here it appears not to get a bounded *width*, so it never scrolls horizontally and long lines run off / get clipped.

## Repro
Open a konstruction with a very long physical line — the Kodee sample's `listOf(v(...), v(...), …)` point arrays (hundreds of entries on one line) are a clean case. The tail of the line is unreachable.

## Investigate the Konstructor side first
1. **Width bounding.** The layout is HTML/CSS-driven: `MainScreen` renders `ContentPane(Modifier.fillMaxSize())`, the 3D is a separate `#gl-pane` div, and `setCodeOnLeft` (JsBridge) arranges the split. The editor itself is `KodeMirror(session, Modifier.fillMaxSize())` in a `Box(fillMaxSize())` (`EditorPane.kt` ~276). Verify a correct **bounded width** actually propagates down to CodeMirror's scroller — if the editor/container sizes to content (longest line) instead of the visible code-pane width, CM won't show a horizontal scrollbar and the line overflows. (Same shape of bug as the vertical case, on the x-axis.)
2. **Overflow handling.** Check the HTML container that hosts the Compose canvas / editor has proper `overflow` (scroll) and a fixed width matching the code half, so content can't silently spill under the `#gl-pane` overlay.
3. **No wrapping extension.** `EditorPane.kt:238` assembles `basicSetup + … + lintGutter()` with no `EditorView.lineWrapping`. Adding it would make long lines wrap within the bounded width and sidestep horizontal scrolling entirely — but first understand why there's currently *neither* wrap nor scroll (a width-bounding fix may be the root cause).

## If Konstructor's width/container is correct
Escalate to kodemirror — likely its Compose wrapper's width handling / scroller (related to the long-line view cluster there, e.g. kodemirror #20). Cross-reference with a minimal repro.

monkopedia-coder:
## Summary
Resolves **#10** — long lines overflowed the editor (neither wrapped nor horizontally scrolled). Two changes in one file each:
- `gradle/libs.versions.toml`: `kodemirror = "0.2.0"` → `"0.3.2"`.
- `EditorPane.kt`: wire the opt-in `lineWrapping` extension into the editor so long lines **soft-wrap** to the editor width (the user's chosen behavior for #10).

## Why 0.3.2 (not 0.3.0/0.3.1)
0.3.0 introduced an editor-mount regression (kodemirror **#92**): `HoverTooltipPlugin` read the session `coroutineScope` before it was attached, crashing every editor mount that uses `linter` (we do, via lint's `hoverTooltip`). That failed 7 of our e2e tests. kodemirror shipped the fix in **0.3.2** (0.3.1 failed to publish). Transitively moves `com.monkopedia.lsp:lsp` RC → **1.0.0 stable**.

## Validation
- Full **clean `:e2e:test` green** on 0.3.2 — including the 7 editor-mount tests (Theme/Keymap, Target, Save) that failed on 0.3.0.
- `:frontend:spotlessKotlinCheck` clean.

Fixes #10

### DIFF
--- a/frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/ui/editor/EditorPane.kt
+++ b/frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/ui/editor/EditorPane.kt
@@ -70,6 +70,7 @@ import com.monkopedia.kodemirror.themonedark.oneDark
 import com.monkopedia.kodemirror.view.KodeMirror
 import com.monkopedia.kodemirror.view.editorContentStyle
 import com.monkopedia.kodemirror.view.keymapOf
+import com.monkopedia.kodemirror.view.lineWrapping
 import com.monkopedia.kodemirror.view.onSave
 import com.monkopedia.kodemirror.view.saveKeymap
 import com.monkopedia.konstructor.common.KonstructionType
@@ -236,7 +237,7 @@ private fun EditorContent(
             config = LintConfig(delay = 0)
         )
         val extensions = basicSetup + themeExt + fontExt + kotlinLang +
-            keymapExt + saveKeymap + saveExt + lintExt + lintGutter()
+            keymapExt + saveKeymap + saveExt + lintExt + lintGutter() + lineWrapping
         val config = com.monkopedia.kodemirror.state.EditorStateConfig(
             doc = content.asDoc(),
             extensions = extensions
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -15,7 +15,7 @@ shadow = "8.3.6"
 javafx = "0.1.0"
 compose-multiplatform = "1.10.3"
 lifecycle = "2.10.0"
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
Tier-2 UX signoff approved by Monkopedia. kodemirror 0.3.2 + editor soft-wrap (fixes #10); e2e green incl. the 7 editor-mount tests that failed on 0.3.0.

PR #31, opened 2026-06-02
Vim display-line motion setting (j/k → gj/gk, vim-only)

### LINKED ISSUE #29: Setting: remap j→gj / k→gk in Vim mode (display-line motion, vim-only)
## Request
Add a persisted setting that remaps `j`→`gj` and `k`→`gk` in the editor's Vim mode — i.e. vertical motion moves by **visual (wrapped) line** rather than logical line. The setting should only be **available/visible when the Vim keymap is selected** (no effect for Default/Emacs keymaps).

Motivation: pairs with editor soft-wrap (issue #10 / the kodemirror 0.3.0 bump enabling `lineWrapping`) — with wrapped lines, plain `j`/`k` skip whole logical lines; `gj`/`gk` move within the wrapped display rows.

## Feasibility — confirmed, fully doable konstructor-side with kodemirror 0.3.0
kodemirror's vim engine exposes a public mapping API and the display-line motion already exists:
- `com.monkopedia.kodemirror.vim.Vim.map(lhs: String, rhs: String, ctx: String? = null)` (Vim.kt:203) — `ctx` scopes the mapping (e.g. normal/visual). Also `unmap(lhs, ctx?)` and `noremap(...)`.
- `moveByDisplayLines` motion exists (VimMotions.kt:279) — it IS the gj/gk behavior. The source even documents the intended usage: `Vim.map("j", "gj")` / `Vim.map("k", "gk")`.

So implementation is: when the setting is ON and the Vim keymap is active, call `Vim.map("j", "gj")` and `Vim.map("k", "gk")` (consider normal + visual contexts); when OFF (or keymap changes away from Vim), `Vim.unmap(...)` to restore. Verify whether the mapping is global vs per-editor-instance and place the call accordingly (likely in/near `EditorPane`'s vim keymap wiring, `EditorPane.kt:209-213`).

## Implementation sketch
1. `SettingsViewModel`: add a persisted `PersistedStateFlow.boolean` (e.g. `vimDisplayLineMotion`, prefix `konstructor.`), default false. Expose flow + setter (mirror the existing `keymap`/`showFps` etc.).
2. `SettingsPane`: add a toggle row, **conditionally rendered only when `keymap == KeymapName.VIM`** (so it doesn't show for Default/Emacs).
3. `EditorPane`: when keymap is VIM, apply/remove the `Vim.map`/`unmap("j"/"k")` based on the setting (react to both the setting and keymap changes; clean up on dispose / keymap switch).

## Dependencies / sequencing
- Requires **kodemirror 0.3.0** (has `Vim.map` + `moveByDisplayLines`). That bump is in flight (PR for #10). This should land **after** that PR to get the API and to avoid `EditorPane.kt` edit conflicts.

## Constraints
- Vim-only (gated on keymap == VIM). Behavior-preserving when off. Verify the full e2e suite green locally (Mesa-EGL override); consider a small e2e asserting the toggle remaps motion under wrapped lines if feasible without scope creep. Apache header / Spotless.

monkopedia-coder:
## Summary
Adds a persisted **Vim display-line motion** setting that remaps `j`→`gj` and `k`→`gk` when the Vim keymap is active, so vertical motion follows wrapped **display** lines rather than logical lines. Pairs with the editor soft-wrap enabled in #10. The toggle is **only shown when the keymap is Vim** — no visibility or effect for Default/Emacs.

## Changes (3 files, frontend only)
- **`SettingsViewModel.kt`** — new persisted `PersistedStateFlow.boolean("vimDisplayLineMotion", false)` with exposed flow + setter, mirroring the existing `keymap`/`showFps`/etc. settings.
- **`SettingsPane.kt`** — a `SettingSwitchRow` rendered only when `keymap == KeymapName.VIM`.
- **`EditorPane.kt`** — a `DisposableEffect` keyed on `(keymap, vimDisplayLineMotion)` that applies/removes the mapping imperatively.

## Vim mapping scope & cleanup
The kodemirror Vim engine (`Vim.map`/`Vim.unmap`, verified on 0.3.2) is a **global singleton** — mappings live in a shared keymap, not per-editor. So the mapping is applied imperatively from a `DisposableEffect` rather than as an editor extension:
- ON + Vim active → `Vim.map("j","gj")`, `Vim.map("k","gk")`.
- `onDispose` (which re-runs whenever the keys change) → `Vim.unmap("j")`, `Vim.unmap("k")`.

A **null mapping context** is used, which the engine applies in normal **and** visual modes (insert is excluded by the engine), matching the documented `Vim.map("j","gj")` idiom — so a single map covers both modes. Cleanup runs when the setting turns off, the keymap leaves Vim, or the editor disposes, so the global mapping never leaks across keymap switches or remounts.

## Verification
- Full e2e suite green locally (Mesa-EGL override): **39 tests, 0 failures, 0 errors** (6 skipped as usual). The keymap/settings tests (`ThemeAndKeymapTest` incl. `testKeymapSwitching`, `testSettingsPaneShowsDropdowns`) stay green with the new vim-gated toggle.
- `:frontend:spotlessKotlinCheck` passes; frontend WasmJS compiles clean (only pre-existing unrelated threejs opt-in warnings).

No e2e was added for the toggle: asserting it would require new TestBridge state/action plumbing (`JsBridge.kt`/`BridgeStateSnapshot.kt`), which is out of the issue's tight 3-file scope.

Fixes #29

### DIFF
--- a/frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/ui/editor/EditorPane.kt
+++ b/frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/ui/editor/EditorPane.kt
@@ -23,6 +23,7 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
@@ -176,6 +177,27 @@ private fun EditorContent(
     val settingsVm = koinInject<SettingsViewModel>()
     val themeName by settingsVm.editorTheme.collectAsState()
     val keymapName by settingsVm.keymap.collectAsState()
+    val vimDisplayLineMotion by settingsVm.vimDisplayLineMotion.collectAsState()
+
+    // The Vim mapping engine is a global singleton, so apply/restore the j/k →
+    // gj/gk remapping imperatively (rather than as an editor extension) and tie
+    // its lifetime to this effect. A null context covers normal + visual modes
+    // (insert is excluded by the engine), so vertical motion follows wrapped
+    // display lines. Cleanup runs when the keymap leaves Vim, the setting turns
+    // off, or the editor disposes — keeping it from leaking across switches.
+    DisposableEffect(keymapName, vimDisplayLineMotion) {
+        val mapped = keymapName == KeymapName.VIM && vimDisplayLineMotion
+        if (mapped) {
+            com.monkopedia.kodemirror.vim.Vim.map("j", "gj")
+            com.monkopedia.kodemirror.vim.Vim.map("k", "gk")
+        }
+        onDispose {
+            if (mapped) {
+                com.monkopedia.kodemirror.vim.Vim.unmap("j")
+                com.monkopedia.kodemirror.vim.Vim.unmap("k")
+            }
+        }
+    }
 
     val kotlinLang = remember { StreamLanguage.define(kotlin).extension }
 
--- a/frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/ui/settings/SettingsPane.kt
+++ b/frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/ui/settings/SettingsPane.kt
@@ -50,6 +50,7 @@ fun SettingsPane(modifier: Modifier = Modifier) {
     val showCameraWidget by settingsVm.showCameraWidget.collectAsState()
     val editorTheme by settingsVm.editorTheme.collectAsState()
     val keymap by settingsVm.keymap.collectAsState()
+    val vimDisplayLineMotion by settingsVm.vimDisplayLineMotion.collectAsState()
 
     Column(
         modifier = modifier.padding(16.dp),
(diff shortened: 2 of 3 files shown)


monkopedia-reviewer (approved):
Tier 1 (feature_internal; tier-2 UX-signoff for #29 pre-approved by the user, so handled to full tier-1 standard).

Correctness / leak-safety (the crux): verified against kodemirror 0.3.2 `Vim.map`/`unmap` (VimHelpers.kt). `map("j","gj")` inserts a `KeyToKeyCommand(keys="j", context=null)` at the front of the global `defaultKeymap`; `unmap("j")` (ctx=null) removes the first reverse-scan entry matching `keys=="j" && context==null` — a precise inverse. The `DisposableEffect` captures a `mapped` boolean and unmaps iff it mapped, so apply/cleanup are strictly symmetric. DisposableEffect runs onDispose before re-firing on key change, and the keys `(keymapName, vimDisplayLineMotion)` cover all transitions: setting off, keymap leaving Vim, and editor dispose all trigger unmap. No stale global j→gj leak across konstructions or keymap switches. null context covering normal+visual (insert excluded by engine) matches the documented idiom.

Vim-gated visibility (`if (keymap == KeymapName.VIM)`) doesn't affect the always-rendered keymap dropdown, so the settings e2e (ThemeAndKeymapTest) is unaffected — confirmed green in CI.

Quality: matches the existing persisted-flow idiom (store + flow + setter mirroring keymap/showFps), `SettingSwitchRow` call matches its signature, and the fully-qualified `com.monkopedia.kodemirror.vim.Vim.*` is consistent with the existing fully-qualified `...vim.vim()` in the same file. Good comments, no dead code.

CI: build-and-test green (10m6s, includes the e2e Playwright/Xvfb gate). MERGEABLE/CLEAN. Approving + squash-merging.

PR #41, opened 2026-06-03
LSP Phase 1: flag-gated ksrpc transport pipe (stub server)

monkopedia-coder:
First phase of the LSP epic (#35). Proves the LSP message pipe end-to-end with a **stub** server — no real Kotlin LSP engine yet. **Additive + flag-gated**: with `lspEnabled` OFF the editor behaves byte-for-byte as today.

## protocol/
- Add `com.monkopedia.lsp:lsp-ksrpc:1.0.0` (as `api`, since the `@KsService` LSP interfaces appear in the public `KonstructionService` surface). Built against ksrpc 1.1.0 / Kotlin 2.3.20 = our versions.
- `KonstructionService` gains the nested LSP sub-service:
  ```kotlin
  @KsMethod("/lsp")
  suspend fun lsp(client: KsrpcLanguageClient): KsrpcLanguageServer
  ```
  Passing `KsrpcLanguageClient` as a **param** lets ksrpc multiplex the reverse `publishDiagnostics` channel onto the same WebSocket (no second connection, no `connectAsLspServer`). `KonstructionService` already extended `RpcBidiService` (required because `lsp()` both returns and accepts a sub-service), so **the tier bump did not ripple** into `Konstructor`/`Workspace`.

## backend
- `StubLanguageServer` (subclass `DefaultLanguageServer`): canned `InitializeResult` advertising hover + completion; on `textDocument/didOpen` pushes one canned `publishDiagnostics` (a single fake warning) back through the held `KsrpcLanguageClient`.
- `KonstructionServiceImpl.lsp(client)` stashes the client and returns the stub, served through the existing ksrpc graph.

## frontend (wasmJs) — behind the `lspEnabled` flag, default OFF
- Add kodemirror `lsp-client` + `lsp-ksrpc` deps.
- `EditorLspClient` (a `KsrpcLanguageClient`) routes server-pushed diagnostics into kodemirror's `LSPClient.languageClient`.
- `EditorPane` builds `LSPClient(server = lsp(EditorLspClient()))` and appends `languageServerSupport(...)` to the editor extensions **only when the flag is on and the session is ready** — OFF leaves the extensions chain identical to before.
- `SettingsViewModel.lspEnabled` (`PersistedStateFlow`, default `false`).

## e2e
- `LspPipeTest`: with the flag ON, opening a konstruction round-trips the canned `publishDiagnostics` back to the editor (`count=1`, asserted via a JS-bridge hook).

## Verification
[...]

Issue #40, opened 2026-06-03
LSP Phase 5: harden — lifecycle, warmth, multi-doc, ops; module-extraction decision


Monkopedia:
Part of #35. Blocked on Phase 4. Subprocess lifecycle + keep-warm + recovery, multi-open-konstruction multiplexing, error handling, latency tuning, adolin deploy/ops. Then decide whether the translation/proxy earns extraction into a standalone `csgs-lsp` module vs staying in the backend.

monkopedia-coder:
Teardown/leak specifics (from the lsp-kotlin maintainer) are captured on #37 — the key item for this harden phase is the **sub-service leak test** (open+close editor N× → channel count returns to baseline, mirroring ksrpc's RpcSubserviceLeakTest) plus tying the lsp() sub-service to konstruction lifecycle and close()-ing it (the shared long-lived socket leaks 2 sub-channels per open→close otherwise).

monkopedia-coder:
## Diagnostics teardown correctness — fix in this harden phase (from the lsp-kotlin maintainer's review of #46's `PullDiagnosticsPublisher`)
The event-driven publisher (#46) is sound, but three close-time correctness items belong here (they're lifecycle/teardown, and this phase's leak test should validate #1):

1. **uri-space key consistency (potential real leak).** The publisher keys `openDocs`/`resultIds`/`changeJobs` by the string passed in. `onOpen` uses `params.textDocument.uri` (didOpen) while `onClose` uses the translated `frontendUri` — if those are different uri spaces, `onClose` won't match the `onOpen` key → the doc never leaves `openDocs`/`resultIds` → it keeps getting re-pulled on every `workspace/diagnostic/refresh` after close, and leaks. Ensure onOpen/onChange/onClose all key by ONE consistent uri space; the pull/publish seams translate at the boundary. (Contract for the eventual extraction: "all lifecycle callbacks use the same uri space; seams translate.")
2. **Clear-on-close.** `onClose` forgets state but doesn't emit `publish(uri, emptyList())` — if kodemirror doesn't clear on its own didClose, the last squiggles linger after the doc closes. Emit a clearing publish on close (a `clearOnClose: Boolean = true` knob in the eventual extracted version).
[...]

Issue #38, opened 2026-06-03
LSP Phase 3: diagnostics through the ±3-line translation layer


Monkopedia:
Part of #35. Blocked on Phase 2. Implement the csgs↔kotlin position translation (fixed 3-line header + footer offset; full-document sync for v1), so real kcsg diagnostics from kotlin-lsp map back onto the user's csgs lines as live squiggles. Align/share with the #33 cursor-line footer + #32 error-surfacing so they use one mechanism. Prototype the translation in the backend.

monkopedia-coder:
Maintainer flag for this phase: the ±3-line csgs↔wrapped-kt translation interacts directly with diagnostic **ranges** — an off-by-N silently lights up the wrong line. Add a **round-trip test**: engine range (wrapped-kt) → translated → asserted against the known kcsg error line in the csgs. (Phase 2 already proved 1 real error / 0 false positives end-to-end, so the engine + classpath are sound; this phase is purely getting the line math right.)

Issue #35, opened 2026-06-03
[Epic] LSP support in the editor (kcsg-aware diagnostics/completion/hover)


Monkopedia:
## Goal
Bring real language intelligence to the editor for user-written kcsg "csgs" scripts: diagnostics, completion, hover, signatureHelp — kcsg-DSL-aware. Built **additively and flag-gated** so the existing compile/render + lint path keeps working untouched until LSP is proven.

## Architecture (designed; spikes complete)
**Transport — nested ksrpc, both directions over the existing WebSocket** (confirmed with the lsp-kotlin agent):
- `lsp-ksrpc` already ships bidirectional `KsrpcLanguageServer`/`KsrpcLanguageClient` `@KsService` interfaces (all features modeled, wasmJs+jvm, ksrpc 1.1.0 — our version). kodemirror's `LSPClient` takes a plain `LanguageServer` (which `KsrpcLanguageServer` is). So konstructor writes ~no transport code.
- Carry it as a nested sub-service: `@KsMethod("/lsp") suspend fun lsp(client: KsrpcLanguageClient): KsrpcLanguageServer` — pass the editor client as a PARAM (ksrpc multiplexes the reverse `publishDiagnostics` channel on the same socket). The hosting interface (likely `KonstructionService`, since LSP is per-konstruction) must extend **`RpcBidiService`** (compile-time tier). Optionally wrap with `LifecycleTrackingLanguageServer`.

**Engine — JetBrains official `kotlin-lsp` (validated by PoC):**
- Works for our exact shape (single synthesized `.kt` + explicit classpath, **no build project**) via a synthesized **`workspace.json`** at the konstruction's workspace root pointing at konstructor's **`lib.jar`** as one project library (it bundles kcsg + kcsg-dsl + stdlib).
- PoC evidence: accurate kcsg-aware diagnostics (real error caught, zero false positives on the DSL), completion offers DSL members, hover resolves kcsg signatures.
- Cost: heavyweight subprocess (standalone ~393MB w/ bundled JBR 25, 2GB heap, cold index ~120s, warm: init ~4s / didOpen→diagnostics ~8s / completion 0.2–2.2s). Run ONE warm instance; install separately on adolin and point konstructor at it; flag-gated so absence = LSP off.

[...]

PR #42, opened 2026-06-03
LSP Phase 2: warm kotlin-lsp subprocess + per-konstruction workspace.json + stdio bridge

### LINKED ISSUE #37: LSP Phase 2: warm kotlin-lsp subprocess + per-konstruction workspace.json
Part of #35. Blocked on Phase 1. Manage ONE warm JetBrains `kotlin-lsp` (`bin/intellij-server --stdio`, `--system-path` for index cache); per konstruction synthesize the `workspace.json` (→ `lib.jar` as one project library, schema in #35) + write the wrapped `.kt`; bridge the backend `KsrpcLanguageServer` impl to the subprocess over stdio LSP. Ops: install kotlin-lsp on adolin (~393MB, bundled JBR, 2GB heap), point konstructor at it via the flag; absence = LSP off. Heavyweight — keep warm.

monkopedia-coder:
Part of #35; Fixes #37. Builds on Phase 1 (#41). **Additive + flag-gated** (`SettingsViewModel.lspEnabled`, default OFF): with the flag off (and on CI, which has no engine binary), the editor behaves byte-for-byte as before.

Replaces Phase 1's stub `KsrpcLanguageServer` with a REAL bridge to a JetBrains **kotlin-lsp** (`intellij-server --stdio`) subprocess, producing genuine kcsg-aware diagnostics for the open konstruction. Positions stay in WRAPPED-`.kt` space for now — the ±3-line csgs↔kotlin translation is the next phase (#38).

## Bridge architecture (two LSP legs)
```
frontend ⇄ (nested ksrpc) ⇄ BridgeLanguageServer ⇄ (stdio LSP / asLspConnection) ⇄ kotlin-lsp
                             \__ pull diagnostics from engine, push up to editor __/
```
- **Frontend↔backend leg (nested ksrpc):** the Phase 1 `KonstructionService.lsp(client): KsrpcLanguageServer` is kept; the editor's `KsrpcLanguageClient` is stashed.
- **Backend↔subprocess leg (dedicated stdio LSP):** `KotlinLspProcess` spawns ONE warm `intellij-server --stdio --system-path <dir>`, wraps its stdin/stdout with lsp-ksrpc's **`asLspConnection`**, and `SingleChannelConnection<String>.connectAsLspClient(forwarder)` yields a subprocess-facing `KsrpcLanguageServer` stub. `BridgeLanguageServer` (a `DefaultLanguageServer` subclass) **delegates** each request to that stub. The `forwarder` is a `KsrpcLanguageClient` whose `textDocumentPublishDiagnostics` forwards to the stashed frontend client (with the wrapped-`.kt` URI rewritten back to the editor's `content.csgs` URI).
- Single managed subprocess (classpath is fixed); multi-konstruction multiplexing + keep-warm tuning is Phase 5 (#40). Phase 2 scopes to the active konstruction.

### Pull→push diagnostics (key finding)
kotlin-lsp does **not** proactively push `publishDiagnostics`; it answers PULL `textDocument/diagnostic` (confirmed against the binary: push=None across a full warm cycle). kodemirror's editor client (the Phase 1 design) listens for PUSHED diagnostics. So on didOpen the bridge **polls** the engine via pull (retrying through the ~120s cold index, then a slow steady-state refresh) and forwards the resulting items up as `publishDiagnostics`. Any push the engine *does* send is also forwarded — they coexist.

## Per-konstruction setup
`KonstructionLspWorkspace` synthesizes, under the konstruction's script dir, a workspace dir containing:
- the **wrapped `.kt`** via `KonstructionControllerImpl.copyContentToScript` (konstructor's EXACT compiler wrapping — header swap + footer — so LSP positions match the compiler), and
- a **`workspace.json`** mirroring the proven PoC recipe: a `JAVA_MODULE` with `moduleSource` + a `library` dep, and one project-level library pointing at the resolved `lib.jar` as a single `CLASSES` root with `inclusionOptions: root_itself`, lowercase `compile` scope, inline polymorphic `type` discriminators. Built with JSON object builders (not `@Serializable` classes) so the backend needs no serialization compiler plugin.

## Lifecycle gating (per the lsp-kotlin maintainer)
`LifecycleState` is driven by hand: `initialized()`/`shutdown()`/`exit()` → `lifecycle.advanceTo(...)`, and `lifecycle.awaitInitialized()` gates the first server→client emit (no diagnostics before the client sent `initialized`). No blocking client-bound request is made from inside `initialize()` (deadlock risk). On `exit` the poll is cancelled and the subprocess-facing stub is `close()`d (cheap leak guard; full teardown/leak test is Phase 5 / #40).

## Config / graceful-off
`KONSTRUCTOR_KOTLIN_LSP` (or `-Dkonstructor.kotlinLsp`) points at the binary; `KONSTRUCTOR_KOTLIN_LSP_SYSTEM_PATH` persists/warms the index. If the binary is unset or missing, `Config.isKotlinLspAvailable` is false, `KotlinLspProcess.connect` returns null, and `BridgeLanguageServer` degrades to an inert server (empty capabilities, notifications dropped) — LSP stays off, never a crash. **CI has no binary, so this degrades to off.**

Also fixes the #41 carry-over nit: `EditorPane`'s `produceState` now uses `catch (_: Exception)` (not `Throwable`).

## Verification

**Flag-OFF / CI-equivalent — GREEN (JDK 21, clean):**
- `clean shadowJar` — whole project compiles + bundles.
- `./gradlew test` — backend + protocol unit/integration tests pass (incl. the new CI-safe `KonstructionLspWorkspaceTest` asserting the workspace.json matches the strict PoC schema + wrapped-`.kt` reuse).
- `:backend/:frontend/:e2e/:protocol spotlessCheck` — clean.
- Full `:e2e:test -Pe2e` (Xvfb + Mesa-EGL) — **all green**. `LspPipeTest` was updated: the Phase-1 canned diagnostic is gone, so with no engine on CI the flag-ON path must establish cleanly and surface **zero** diagnostics (asserted), and the app stays healthy; the flag-OFF suite is unchanged.

**Real-engine — LOCAL ONLY (CI does NOT run the 393MB binary):**
Pointed config at `/tmp/kotlin-lsp-poc/.../bin/intellij-server` (warm `--system-path`) and ran `BridgeLanguageServerIntegrationTest` (double-gated on `-Dintegration=true` + `KONSTRUCTOR_KOTLIN_LSP`, so it SKIPS on CI). A deliberate-error konstruction (`val broken = thisSymbolDoesNotExist123` above a valid `cube {}` DSL line) drove the real bridge end to end:
```
REAL-ENGINE diagnostics (1): Unresolved reference 'thisSymbolDoesNotExist123'.
```
Exactly **1** real kcsg-aware diagnostic, routed to the editor's `content.csgs` URI, **zero false positives** on the valid `cube {}`/`primitive {}` DSL. This proves: synthesize workspace.json + wrapped `.kt` → spawn/connect subprocess → initialize/initialized/didOpen → pull diagnostics → push to the frontend client. **This engine path is not CI-tested** (no binary, no bundled lib resource on CI) — stated explicitly.


monkopedia-reviewer (approved):
Reviewed to full tier-1 standard (correctness + quality + CI-green); approving.

**Flag-OFF safety (load-bearing): airtight.** `EditorPane.produceState` returns early on `if (!lspEnabled)` before any `service.lsp()` call, so `BridgeLanguageServer` is never constructed when the flag is off — the editor is byte-for-byte unchanged. **Graceful-off (flag ON, no binary = CI):** `Config.isKotlinLspAvailable` gates `KotlinLspProcess.connect()` → null → `delegate=null` → `initialize()` returns empty `ServerCapabilities()` and every notification handler does `delegate ?: return`. Inert, never a crash.

**Lifecycle: correct.** Hand-driven `LifecycleState` starts at INITIALIZING; `advanceTo(INITIALIZED/SHUTTING_DOWN/EXITED)` are all legal transitions and `advanceTo` is the idempotent no-throw variant. `awaitInitialized()` gates the first push; no blocking client-bound call inside `initialize()`. `exit()` cancels the poll + scope Job and `exit()/close()`s the subprocess stub. The warm subprocess being kept across konstructions is the documented Phase-5/#40 deferral, not a leak.

**Tests meaningful.** `KonstructionLspWorkspaceTest` (CI-safe) asserts the strict workspace.json schema (JAVA_MODULE, moduleSource+library deps, lowercase `compile`, `root_itself` CLASSES root, lib.jar path) + wrapped-`.kt` reuse of `copyContentToScript`. `BridgeLanguageServerIntegrationTest` genuinely drives the real engine, correctly double-gated and CI-skipped. `LspPipeTest` updated to assert the inert flag-ON path surfaces zero diagnostics and the app stays healthy.

**Quality.** Idiomatic Kotlin; dead Phase-1 `StubLanguageServer` removed with no orphan refs; Apache headers present; scope limited to Phase 2 (no #38 position translation); the #41 `Throwable`→`Exception` nit is fixed. Two non-blocking nits: one inline-FQN `com.monkopedia.lsp.ServerCapabilities()` against the file's import convention, and `messages` computed a line before the empty-check in the integration test — neither worth blocking.

CI: `build-and-test` green, MERGEABLE/CLEAN (verified via gh pr checks). Approving and merging.

PR #52, opened 2026-06-04
LSP: guard nullable-result calls so a null engine response can't kill the session

monkopedia-coder:
## Bug
On first real LSP use, after the first cursor move, **completion and diagnostics silently stop** ("once I edit, no more updates"). Backend log showed a tight loop of `JsonDecodingException ... had 'n'` (JSON `null`) at `com.monkopedia.lsp.Hover$$serializer.deserialize`.

## Root cause (traced with the lsp-kotlin maintainer)
kotlin-lsp returns `null` for spec-nullable results (hover-over-nothing, no-completions, …), but lsp-types 1.1.0 generates ~45 of these results **non-nullable** (a codegen bug; maintainer is fixing it in 1.1.1). Decoding `null` into a non-nullable type throws a `SerializationException` **in our calling coroutine**. ksrpc's read-loop is per-call and survives — but the exception was escaping into the coroutine **scope we share** with the diagnostics publisher and other LSP calls, cancelling it. The editor fires hover on every cursor move, so one hover-over-nothing tore down the whole session.

## Fix
Wrap the four interactive nullable-result forwards — `textDocumentHover`, `textDocumentCompletion`, `completionItemResolve`, `textDocumentSignatureHelp` — so a `SerializationException` degrades to the existing empty result (`super.x(...)`, or the item unchanged) instead of escaping to cancel the shared scope. Catches **only** `SerializationException` (the null-decode), so real errors still surface; the happy path is unchanged.

## Verification
- Real-engine integration tests `realEngineResolvesKcsgHover` + `realEngineProducesKcsgDiagnostic` pass with the guards.
- `spotlessCheck` green all modules; backend compiles. Flag-off behavior unchanged (LSP default off).

## Follow-ups (not in this PR)
- Remove these catches once we bump to **lsp-types 1.1.1** (maintainer's systemic nullable-result fix — directly consumable on 2.4.0).
- **Separate, still-open**: completion isn't being *offered* by the engine (distinct from this connection bug) + the workspace `java.*`/JDK gap — both under investigation with the maintainer.

Part of epic #35.

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/lsp/BridgeLanguageServer.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/lsp/BridgeLanguageServer.kt
@@ -55,6 +55,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.serialization.SerializationException
 import kotlinx.serialization.json.JsonNull
 
 /**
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
Approve — tier 1 (bug_fix, flag-gated behind `lspEnabled`, default OFF; no public-API/UI/UX/perf surface, flag-off behavior unchanged).

Reviewed the whole diff to merge standard:
- **Scope:** exactly one file (`lsp/BridgeLanguageServer.kt`, +46/-16); Apache header intact; no smuggled changes.
- **Catch is correct:** all four guards catch ONLY `kotlinx.serialization.SerializationException` (import at line 58), per-call — real transport/other errors still propagate; only the null-decode degrades.
- **Happy path byte-identical:** the `try` wraps only the delegate call; the `translate*` step stays outside, so a successful decode behaves exactly as before.
- **Fallbacks reuse the file's own idiom:** completion/hover/signatureHelp fall back to `super.x(params)` (the same expression the pre-existing no-delegate branch uses); completionItemResolve returns the input `params` (item unresolved).
- **Stopgap is documented:** comments explain it's temporary pending lsp-types 1.1.1.
- Real-engine repro (`realEngineResolvesKcsgHover` + `realEngineProducesKcsgDiagnostic`) reported passing locally with the guards; CI `build-and-test` green.

Non-blocking follow-up (does not gate this merge): in lsp-ksrpc 1.1.0, `DefaultLanguageServer.{textDocumentHover,textDocumentCompletion,textDocumentSignatureHelp}` (i.e. `super.x()`) actually THROW `NotImplementedError` rather than returning an empty value (verified in the published jar). The SerializationException poison originates inside the *delegate* call (subprocess-channel decode), which the `try` correctly contains — so the scope-cancellation fix holds and the live repro passes; the subsequent `super.x()` throw surfaces only as a per-call RPC error to the frontend on that one hover-over-nothing (ksrpc's per-call isolation, which the editor tolerates). Still, emitting a spurious error per hover-over-nothing is slightly worse than a clean empty result. When you remove these catches on the 1.1.1 bump, consider whether a literal empty result is preferable to `super.x()` here. Not blocking — reused idiom + green repro.

PR #61, opened 2026-07-08
Factor repeated broadcast + launch-with-timeout boilerplate in KonstructionServiceImpl

### LINKED ISSUE #56: Factor repeated listener-broadcast + launch-with-timeout boilerplate in KonstructionServiceImpl
**Category:** duplication
**Where:** `backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionServiceImpl.kt`

**Current:** Two patterns recur throughout the class:

1. **Listener broadcast** — `listenerLock.withLock { listeners.values.toList() }.forEach { it.onX(...) }` appears 6 times (`setName`, `set`, `compile`, `konstruct` x2 broadcast blocks, plus the `onTaskComplete` tails). Every call site re-snapshots the listener map under the lock and iterates.

2. **Fire-and-forget request methods** — `requestCompile`, `requestKonstruct`, and `requestKonstructs` (lines ~221, ~290, ~304) are the same shape:
```kotlin
scope.launch {
    callContext("<name>.launch", baseCallSign = konstructionController.callSign) {
        withTimeout(120.seconds) { <delegate>() }
    }
}
```
and `withTimeout(120.seconds)` appears 4 times in the file (the three request* methods + `ListenerHandler.doCallback`) as an unnamed magic value.

**Proposed:**
- Add a private `suspend inline fun broadcast(block: ListenerHandler.() -> Unit)` that does the `listenerLock.withLock { listeners.values.toList() }.forEach { it.block() }` snapshot-and-iterate, and replace the 6 call sites.
- Add a private helper for the request* trio, e.g. `private fun launchRequest(name: String, block: suspend () -> Unit)` wrapping the `scope.launch + callContext + withTimeout` shell, so the three methods become one line each.
- Name the timeout, e.g. `private val callTimeout = 120.seconds`, and use it at all four sites.

**Why:** The broadcast snapshot is subtle (must `.toList()` under lock before iterating to avoid concurrent-modification / holding the lock across suspending callbacks) — centralizing it means that invariant lives in one place instead of being re-derived 6 times. The request* trio and the 120s magic number are pure copy-paste; one helper + one named constant makes the timeout a single tunable.

**Effort:** small refactor

monkopedia-coder:
Backend-only refactor addressing the code-health duplication flagged in #56. No behavior change.

## What changed
`KonstructionServiceImpl` re-derived two patterns at many call sites; both are now centralized:

- **Listener broadcast** — `listenerLock.withLock { listeners.values.toList() }.forEach { ... }` appeared 6 times (`setName`, `set`, `compile` x2, `konstruct` x2). Replaced with a single `private suspend inline fun broadcast(block: ListenerHandler.() -> Unit)`. The snapshot-under-lock-then-iterate invariant (copy before iterating so suspending callbacks never run under the lock / hit a CME) now lives in one place.
- **Fire-and-forget `request*`** — `requestCompile`, `requestKonstruct`, `requestKonstructs` shared the same `scope.launch + callContext("<name>.launch") + withTimeout` shell. Extracted into `private fun launchRequest(name, block)`; each method is now one line.
- **Magic timeout** — the four `120.seconds` sites (three `request*` + `ListenerHandler.doCallback`) now reference a named `callTimeout` constant.

## Behavior preserved
- Same broadcast semantics (snapshot under lock, iterate outside).
- Same `callContext` names (`requestCompile.launch`, etc.) and same base call sign.
- Same 120s timeout at all four sites.

## Verification
- `:backend:jvmTest` — BUILD SUCCESSFUL (the one pre-existing warning is in an unrelated test file).
- `:backend:spotlessKotlinCheck` — clean.

Fixes #56

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionServiceImpl.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionServiceImpl.kt
@@ -59,6 +59,9 @@ import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeout
 
+/** Upper bound on a single compile/konstruct/listener-callback before it's cancelled. */
+private val callTimeout = 120.seconds
+
 class KonstructionServiceImpl(
     private val config: Config,
     private val workspaceId: String,
@@ -75,6 +78,31 @@ class KonstructionServiceImpl(
 
     override val serviceName: String
         get() = "KonstructionService"
+
+    /**
+     * Snapshots the current listeners under [listenerLock] and invokes [block] on each.
+     *
+     * The `.toList()` under the lock is load-bearing: it copies the handler set before
+     * iterating so the (suspending) callbacks never run while holding the lock and can't
+     * hit a concurrent modification of [listeners].
+     */
+    private suspend inline fun broadcast(block: ListenerHandler.() -> Unit) {
+        listenerLock.withLock { listeners.values.toList() }.forEach { it.block() }
+    }
+
+    /**
+     * Fire-and-forget shell shared by the `request*` methods: launches [block] on [scope]
+     * inside a named [callContext], bounded by [callTimeout].
+     */
+    private fun launchRequest(name: String, block: suspend () -> Unit) {
+        scope.launch {
+            callContext("$name.launch", baseCallSign = konstructionController.callSign) {
+                withTimeout(callTimeout) {
+                    block()
+                }
+            }
+        }
+    }
 
     override suspend fun getShipper(u: Unit): Shipper {
         val info = konstructionController.info.konstruction
@@ -116,8 +144,8 @@ class KonstructionServiceImpl(
                 konstruction = info.konstruction.copy(name = name)
             )
             konstructionController.info = newInfo
-            listenerLock.withLock { listeners.values.toList() }.forEach {
-                it.onInfoChanged(newInfo, emptyList())
+            broadcast {
+                onInfoChanged(newInfo, emptyList())
             }
         }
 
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
Reviewed to merge standard — clean, correct, behavior-preserving refactor.

**Correctness:**
- `broadcast` preserves the snapshot-under-lock/iterate-outside invariant exactly: `listenerLock.withLock { listeners.values.toList() }.forEach { it.block() }` is a byte-for-byte equivalent of the six inlined patterns. The `suspend inline` is load-bearing (allows suspend calls inside the non-suspend lambda type by virtue of inlining into a suspend call site) — CI green confirms the compiler accepted it.
- `launchRequest` preserves all three call context names (`requestCompile.launch`, `requestKonstruct.launch`, `requestKonstructs.launch`) and the same `konstructionController.callSign` base — tracing is unaffected.
- All four `120.seconds` timeout sites (three `request*` + `ListenerHandler.doCallback`) now share `callTimeout`. No timeout semantics changed.
- No CancellationException swallowing; no GlobalScope; structured scope (`Dispatchers.IO + job`) unchanged.

**Quality:**
- `dry-refactor-should-shrink-callsites`: call sites genuinely shrank (each `request*` method: 10 lines → 1 line; each broadcast pattern: 3-5 lines → 2-3 lines). Near-neutral net line count is explained entirely by added KDoc — not added indirection.
- `callTimeout` correctly typed as `val` rather than `const` (Duration isn't a compile-time constant) and placed file-private at top level — idiomatic.
- KDoc on both helpers explains the load-bearing invariants (lock-release-before-iterate and the `$name.launch` naming convention). Consistent with surrounding comment density.
- No nullable-then-force-unwrap, no dead code, no scope creep (single file).

CI: build-and-test passed.

PR #66, opened 2026-07-16
Add CachingHeaders: no-cache for un-hashed frontend bundle, long cache for wasm

monkopedia-coder:
## What
Adds a Ktor `CachingHeaders` policy for the served frontend bundle in `App.kt::configureHttp`:
- `index.html` + `frontend.js` (NOT content-hashed) → `Cache-Control: no-cache` (must-revalidate)
- `*.wasm` (content-hashed, immutable URL) → `Cache-Control: max-age=31536000`

Ktor serves `.js` as `text/javascript` (not `application/javascript`), so that content type is matched explicitly.

## Why
The frontend's `index.html`/`frontend.js` are not content-hashed and were served with **no cache headers**, so after a redeploy Chrome could serve a stale `frontend.js` pointing at an old (mismatched) wasm hash. That surfaced as bizarre input/coordinate behavior (the "every click → 0,0" phantom) during the #171 editor click-targeting investigation on adolin. Verified live on the probe deploy: `frontend.js`/`index.html` → `no-cache`, wasm → `max-age=31536000`.

## Scope
Backend-only, additive. Extracted as a standalone change from the larger `lsp-types-1.2.0` WIP branch so it can land independently. No behavior change beyond response cache headers.

## Test
`:backend:compileKotlinJvm` + `spotlessCheck` green locally (JDK 21).

### DIFF
--- a/backend/build.gradle.kts
+++ b/backend/build.gradle.kts
@@ -52,6 +52,7 @@ kotlin {
         implementation(libs.clikt)
         implementation(libs.ktor.server.core)
         implementation(libs.ktor.server.compression)
+        implementation(libs.ktor.server.caching.headers)
         implementation(libs.ktor.server.cors)
         implementation(libs.ktor.server.websockets)
         implementation(libs.ktor.server.status.pages)
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/App.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/App.kt
@@ -27,11 +27,14 @@ import com.monkopedia.ksrpc.channels.SerializedService
 import com.monkopedia.ksrpc.ksrpcEnvironment
 import com.monkopedia.ksrpc.serialized
 import com.monkopedia.ksrpc.server.ServiceApp
+import io.ktor.http.CacheControl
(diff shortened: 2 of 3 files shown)


monkopedia-reviewer (approved):
Reviewed at tier-1 standard — both axes clean.

**Correctness:** CachingHeaders install is correct and complete. Content-type matching covers all three relevant types: `text/html` → no-cache, `application/javascript` + `text/javascript` (Ktor's actual MIME for .js) → no-cache, `application/wasm` → max-age=31536000. The `withoutParameters()` guard prevents charset suffixes from breaking the match. The null-ct guard and else-null default are right — no over-broad match that would wrongly long-cache API responses. Dependency wiring (`ktor-server-caching-headers` catalog entry + backend dep) is correct and uses the shared ktor version ref. PR is self-contained: only 3 files, no LspPrewarmer or other WIP-branch references, extracted cleanly. Import ordering in the PR branch is alphabetically correct (spotless confirmed this in CI too).

**Quality:** Consistent with the surrounding `install(Compression)` style. Comment block above the install explains the why clearly (heuristic caching, stale-frontend root cause, verified on adolin). `31_536_000` is readable. No dead code, redundant logic, or scope creep.

Issue #64, opened 2026-07-16
Migrate Shadow plugin 8.3.6 → 9.x (com.gradleup.shadow major bump)


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)
**Category:** dep-bump

**Where:** `gradle/libs.versions.toml` — `shadow = "8.3.6"`, plugin `shadow = { id = "com.gradleup.shadow", version.ref = "shadow" }`. Latest stable on the Gradle Plugin Portal: **9.5.1** (major 8 → 9).

**Current:** The backend's whole build hinges on Shadow — the `shadowJar` task is the orchestrator that bundles `lib-all.raj` + the wasmJs frontend into the fat JAR (see CLAUDE.md "Build Flow"). The repo is a full major version behind.

**Why this is split out (not in the routine dep-refresh issue):** Shadow 9 is a major release with breaking changes relative to 8.x — notably a raised minimum Gradle/JDK baseline, task-configuration/relocation API adjustments, and changed defaults for service-file / manifest merging. Because `shadowJar` here is load-bearing (it's the release artifact and packages resources from `:lib` and `:frontend`), the upgrade needs its own validation rather than riding along with patch/minor bumps.

**Proposed:** Bump `shadow` to 9.5.1, adapt any changed task/relocation API in the backend build script, and validate a **full** `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew clean shadowJar` — confirming the produced fat JAR still contains `lib-all.raj` and the frontend wasmJs bundle, and that the server boots from it. Cross-check against the Shadow 9 migration notes for the specific API deltas.

**Why:** Keeps the build toolchain current and off an EOL major line; doing it deliberately (with a full-build check) avoids a surprise break the next time the routine bumps force a Gradle/JDK move.

**Effort:** small refactor (version bump + likely a couple of build-script API tweaks + full-build validation)

_Filed by the nightly code-health pass. To act on it: relabel `agent-workable` (it'll flow into triage→work). To decline: close it — it won't be suggested again._

monkopedia-coder:
## Triage 2026-08-21: keeping `needs-decision` — stating the question, and recording the spike verdict, which never reached this issue

This issue has no comments, so the result of the spike run commissioned for it is discoverable only on PR #90. Bringing it here, because the decision cannot be made from what is currently written on the ticket.

[...]

PR #67, opened 2026-07-16
Release prep 0.3.0: hauler 0.4.2 / kcsg 0.4.5 / kodemirror 0.3.5 + toolchain refresh

monkopedia-coder:
## Release prep 0.3.0 — dependency + toolchain catch-up (staged for authorized cut)

PREP only. Do **not** auto-merge — this is staged for Jason's authorized release cut (release → tier-3). No user-facing feature changes.

### Landed dependency set (old → new)

| Dependency | Old | New |
|---|---|---|
| hauler | 0.4.0 | 0.4.2 |
| kcsg / kcsg-dsl | 0.4.3 | 0.4.5 |
| kodemirror | 0.3.2 | 0.3.5 |
| Kotlin | 2.4.0 | 2.4.10 |
| kotlinx-coroutines | 1.10.2 | 1.11.0 |
| kotlinx-serialization | 1.10.0 | 1.11.0 |
| kotlinx-datetime | 0.7.1 | 0.8.0 |
| Ktor | 3.4.2 | 3.5.1 |
| Koin | 4.2.0 | 4.2.2 |
| Compose Multiplatform | 1.10.3 | 1.11.1 |
| AndroidX Lifecycle | 2.10.0 | 2.11.0 |
| Clikt | 5.0.3 | 5.1.0 |
| Logback | 1.5.18 | 1.5.38 |
| Spotless | 8.4.0 | 8.8.0 |

### Held (intentionally not bumped)

- **lsp** — WIP; kept on the current version to avoid entangling in-flight LSP work with the release.
- **shadow** — blocked by #64.
- **slf4j** — catalog stays at **2.0.17**; the *real resolved* version is **2.0.18**, pulled transitively via hauler → ksrpc 1.1.4. (This transitive bump is what surfaced the fix below.)

### Fix: HaulerServiceProvider — SLF4J 2.0.x early-bind

`hauler 0.4.2` transitively pulls `ksrpc 1.1.4` → `slf4j 2.0.18`. SLF4J 2.0.x binds the MDC adapter **eagerly**: `LoggerFactory.earlyBindMDCAdapter()` calls `getMDCAdapter()` **before** `initialize()` runs. Our provider declared `mdcAdapter` as `lateinit` and only assigned it inside `initialize()`, so the early call threw `UninitializedPropertyAccessException` — which killed the compiled user-script subprocess and broke `ExecuteTaskTest`.

Fix: initialize `mdcAdapter` eagerly (drop `lateinit`) with the same adapter type used post-`initialize()`, so `getMDCAdapter()` always returns a valid adapter. Behavior after `initialize()` is unchanged.

```diff
-    private lateinit var mdcAdapter: MDCAdapter
+    private val mdcAdapter: MDCAdapter = NOPMDCAdapter()
 ...
     override fun initialize() {
         loggerFactory = HaulerLoggerFactory()
         markerFactory = BasicMarkerFactory()
-        mdcAdapter = NOPMDCAdapter()
     }
```

### Build changes

- **Node.js pin 22.11.0** for the Kotlin/JS + Compose-wasm toolchains (via the `EnvSpec` API), since the default Compose-wasm Node download can land on a 404/incompatible build.
- **lib archiveVersion side-effect:** setting the project `version` to 0.3.0 would rename the lib shadow JAR to `lib-<version>-all.jar`, but the backend locates it by the fixed name `lib-all.jar` / `lib-all.raj`. The lib shadowJar now pins `archiveVersion.set("")` so the artifact name stays stable.

### Does NOT close #171

The released **kodemirror 0.3.5 contains zero #171 changes**. The reported ½-character offset was a measurement artifact; the real layout-readiness race fix lands in a later kodemirror patch. This PR must not be treated as closing #171.

### Validation

`clean shadowJar :backend:jvmTest spotlessCheck` — shadowJar builds, `ExecuteTaskTest` GREEN (the test the slf4j bug broke), `spotlessCheck` GREEN.

Note: `PullDiagnosticsPublisherTest` is a **pre-existing full-suite flake** (a teardown/logging-coroutine race in untouched LSP test code — a leaked hauler log-write coroutine writes into a temp dir already deleted by `@AfterTest`). It fails nondeterministically (different tests each run) and passes on reruns; reproduced identically on clean `origin/main`. Unrelated to this PR.

### Proposed version

**0.3.0** (Jason may override at the cut).

---

Staged for Jason's authorized cut — **do not auto-merge**.


monkopedia-reviewer (changes requested):
## CI failing — Node.js pin not taking effect for Wasm webpack

**Blocking: `build-and-test` check is RED.**

The `:frontend:wasmJsBrowserProductionWebpack` task failed in CI because Gradle downloaded **node-v25.0.0** instead of the pinned 22.11.0:

```
Execution failed for task ':frontend:wasmJsBrowserProductionWebpack'.
> A problem occurred starting process 'command '/home/runner/.gradle/nodejs/node-v25.0.0-linux-x64/bin/node''
```

The `WasmNodeJsEnvSpec().version.set("22.11.0")` block in `build.gradle.kts` did not suppress the v25.0.0 download that the Compose-wasm toolchain apparently requests from a different resolution path. The `NodeJsPlugin` pin applied fine for the JS toolchain, but the Wasm webpack step still hit the wrong node.

### What needs to happen

1. Diagnose why the Wasm webpack task ignored the `WasmNodeJsEnvSpec` pin and fix it so CI actually uses node 22.11.0. Possible causes: the `WasmNodeJsPlugin`/`WasmNodeJsEnvSpec` API isn't the right hook for the Compose-wasm webpack node, or the spec needs to be configured differently (e.g. on the project rather than via `plugins.withType`), or Compose 1.11.1 changed the plugin class name.
2. Confirm the fix with a green `build-and-test` run on the PR branch.

Everything else in the diff (version bumps, `HaulerServiceProvider` fix, `archiveVersion.set("")`, CHANGELOG) looks correct — the Node pin is the only thing blocking.

monkopedia-reviewer (commented):
Reviewed the whole diff to merge standard — it's sound. Vetted: 14 dep bumps (hauler 0.4.2, kcsg/kcsg-dsl 0.4.5, kodemirror 0.3.5, Kotlin 2.4.10, coroutines 1.11.0, serialization 1.11.0, datetime 0.8.0, Ktor 3.5.1, Koin 4.2.2, Compose MP 1.11.1, lifecycle 2.11.0, Clikt 5.1.0, Logback 1.5.38, Spotless 8.8.0); lsp + shadow + slf4j-catalog correctly held; HaulerServiceProvider SLF4J 2.0.x early-bind fix (eager NOPMDCAdapter, drop lateinit) correct against the provider contract; version=0.3.0 + archiveVersion.set("") on BOTH lib and backend shadowJars keeps lib-all.raj / backend-all.jar stable; Node 22.11.0 pin via allprojects{} EnvSpec using non-deprecated .version Property API; CHANGELOG present, accurate, and explicitly states it does NOT close #171.

All three prior CI blockers confirmed resolved:
1. **Node pin** — now covers every project's EnvSpec via `allprojects { }`, fixing the frontend webpack node-v25.0.0 download.
2. **backend-all.jar rename** — `archiveVersion.set("")` on the backend shadowJar keeps e2e's `-Dkonstructor.jar` path valid (server startup no longer times out).
3. **Listener flake** — the de-flaked `oneSlowListenerDoesNotBlockOthers` test (from #70 via rebase) passes cleanly.

CI fully green now — `build-and-test` PASSED (12m39s): Spotless ✓, shadowJar build ✓, backend+protocol unit tests ✓, e2e 42/42 ✓.

The only thing remaining is your authorized release cut — proposed version **0.3.0** (override at will). Does **not** close #171. @Monkopedia

monkopedia-reviewer (approved):
Repo owner (Jason) authorized the cut directly. Diff previously reviewed to merge standard; CI green (backend + e2e 42/42). Merging the 0.3.0 release prep.

Issue #68, opened 2026-07-16
Flaky PullDiagnosticsPublisherTest: leaked hauler log-write coroutine writes to temp dir after @AfterTest deletes it


monkopedia-coder:
## Summary
`PullDiagnosticsPublisherTest` (backend `:backend:jvmTest`) fails **nondeterministically**. Reproduced on clean `origin/main` (not caused by any open PR): Run 1 failed (2 different Pull* tests), Run 2 passed clean. Passes on rerun and in isolation.

## Root cause (from stack trace)
A leaked **hauler log-write coroutine** calls `appendText` into `/tmp/konstructor-test-<rand>/log.txt.tmp` **after** `@AfterTest` has already deleted that temp dir → `FileNotFoundException`, which surfaces as `UncaughtExceptionsBeforeTest` at the *next* test boundary (so the failure is attributed to whichever Pull* test runs next — hence the moving target). The racing code is around `logging/Writers.kt`; the coroutine outlives the test's teardown.

## Why it matters now
CI currently does **not** run `:backend:jvmTest` (the lifecycle `test` task skips it — being fixed separately). Once CI actually runs the backend suite, this flake will intermittently red CI. So this should be fixed alongside / before enabling backend tests in CI, or CI will be unreliable.

## Fix direction
Ensure the hauler logging writer's coroutine/scope is fully awaited or cancelled before `@AfterTest` deletes the temp dir (deterministic teardown), or tolerate a post-close write. Likely adjacent to the publisher close-time work in 71d9dc5 ("3 publisher close-time fixes") and the LSP teardown/leak hardening.

_Filed during the Round-4 release-prep (found by the release-prep validation run)._

monkopedia-reviewer:
Fixed by #69 (merged 49f5c0de). The cancelAndJoin in BridgeLanguageServer.teardown() ensures hauler IO writers drain before the temp dir is deleted; the runCatching guards in Writers.kt and KonstructionController.info setter provide defense-in-depth against the same race via GlobalScope-parented coroutines. :backend:jvmTest confirmed green (98 passed, 15 skipped) on CI.

PR #69, opened 2026-07-16
CI: actually run :backend:jvmTest (lifecycle `test` skips it)

monkopedia-coder:
## What's broken

In a Kotlin Multiplatform project, the lifecycle `test` task does **not** wire `:backend:jvmTest`. A `--dry-run` confirms that `./gradlew test` only pulls in `:lib:test` (no-source) and `:e2e:test` — never `:backend:jvmTest`. This means the entire backend unit suite (KonstructorImplTest, ExecuteTaskTest, KonstructionControllerImplTest, all the LSP tests, etc.) has **never run in CI**. The CLAUDE.md doc also incorrectly stated `./gradlew test` was sufficient to run backend tests.

## Fix (ci.yaml)

Add `:backend:jvmTest :protocol:jvmTest` explicitly to the test step:

```
./gradlew test :backend:jvmTest :protocol:jvmTest
```

This is the minimal, explicit change. The existing `test` target is kept for `:lib` and `:e2e` (skipped without `-Pe2e`); the two JVM targets are named explicitly.

## Backend suite was RED on main — root cause + fixes included

Wiring the tests directly was blocked because `:backend:jvmTest` was **red on clean main** due to two inter-test contamination bugs. Both have been fixed in this PR so CI comes up green:

**1. `BridgeLanguageServer.teardown()` — cancel-but-no-join**

`teardown()` called `scope.cancel()` but returned immediately. Background hauler log-writer coroutines (on `Dispatchers.IO`) were still in flight when the test's `@AfterTest` deleted the temp dir, causing `FileNotFoundException` in a `GlobalScope`-parented coroutine that hit the global uncaught-exception handler. The next `runTest` call then saw `UncaughtExceptionsBeforeTest`.

Fix: change `scope.coroutineContext[Job]?.cancel()` to `?.cancelAndJoin()` so `teardown()` (which is `suspend`) waits for all scope children to finish before returning.

**2. `Writers.writeText/writeBinary` — unhandled `IOException` in `launchIn`**

The log-file flow collectors (`appendText` / `appendBinary`) were called directly inside `onEach` with no error handling. An `IOException` from a deleted log dir propagated to the `launchIn` coroutine's scope and then to the global uncaught-exception handler — same contamination vector as above, but from `KonstructorImpl`'s `GlobalScope`-based log writers.

Fix: wrap the `appendText`/`appendBinary` calls in `runCatching` in both `writeText` and `writeBinary`.

**3. `KonstructionController.info` setter — bare `GlobalScope.launch`**

The `info` setter fire-and-forgets a `GlobalScope.launch` to persist `info.json`. If the temp dir is deleted before this coroutine runs, the `FileOutputStream` throws `IOException` into `GlobalScope`, which has no exception handler → global uncaught-exception handler again.

Fix: wrap the write in `runCatching` with an `onFailure` log.

## Confirmed green

`:backend:jvmTest` confirmed green on clean `origin/main` with these fixes: **98 tests passed, 15 skipped** (integration tests behind `-Dintegration=true`).  
`:protocol:jvmTest`: **17 tests passed**.  
Exact CI command `./gradlew test :backend:jvmTest :protocol:jvmTest` also confirmed green locally.

## Test plan

- [ ] CI passes (both the new `:backend:jvmTest` + `:protocol:jvmTest` steps and the existing spotless + e2e steps)
- [ ] No integration tests are unexpectedly included (they're skipped without `-Dintegration=true`)
- [ ] CLAUDE.md now correctly documents the command

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,7 +48,7 @@ jobs:
         run: ./gradlew shadowJar
 
       - name: Run backend + protocol unit tests
-        run: ./gradlew test
+        run: ./gradlew test :backend:jvmTest :protocol:jvmTest
 
       - name: Run e2e tests under Xvfb
         run: xvfb-run --auto-servernum --server-args='-screen 0 1280x720x24' ./gradlew :e2e:test -Pe2e
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -31,7 +31,9 @@ JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew shadowJar
 ./gradlew :lib:shadowJar
 
 # Run backend/protocol unit+integration tests
-./gradlew test
+# Note: the lifecycle `test` task does NOT wire :backend:jvmTest in KMP projects;
+# both targets must be named explicitly.
+./gradlew test :backend:jvmTest :protocol:jvmTest
 
 # Run e2e tests (requires full build)
 ./gradlew :e2e:test -Pe2e
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionController.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionController.kt
@@ -19,6 +19,7 @@ package com.monkopedia.konstructor
 
 import com.monkopedia.hauler.CallSign
 import com.monkopedia.hauler.debug
+import com.monkopedia.hauler.error
 import com.monkopedia.hauler.hauler
 import com.monkopedia.kcsg.KcsgScript
 import com.monkopedia.konstructor.common.KonstructionInfo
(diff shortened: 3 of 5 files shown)


monkopedia-reviewer (approved):
Reviewed to tier-1 merge standard. All five parts are correct and coherent.

**CI:** `build-and-test` passed (11m42s) — `:backend:jvmTest` (98 passed, 15 skipped) and `:protocol:jvmTest` (17 passed) both green. The whole point of this PR is now verified.

**`cancelAndJoin` (primary fix — correct, no deadlock):** `BridgeLanguageServer.scope` is a standalone `CoroutineScope(Dispatchers.IO + SupervisorJob())`. `teardown()` is called from `close()` and `exit()`, both dispatched on the ksrpc/Ktor handler scope — entirely separate from the scope being joined. No deadlock risk. This is the right structured-concurrency fix: callers that free temp-dir resources now wait for hauler IO writers to actually drain before returning.

**`runCatching` guards in `Writers.writeText/writeBinary` (defense-in-depth):** Scope is narrow — wraps the single IO operation that can throw `IOException` on teardown. Log appends are genuinely low-stakes; a failed append doesn't corrupt state. Comments clearly explain the teardown-race rationale. Acceptable given the pre-existing `GlobalScope`-based log writer architecture.

**`KonstructionController.info` setter:** `runCatching` wraps the entire lock body; `onFailure { hauler.error(...) }` logs failures visibly rather than silently swallowing them. The `infoImpl = value` settle still runs after a failure, which is correct. No real IO errors hidden without notice.

**CI yaml + CLAUDE.md:** Minimal and correct. The KMP `test` lifecycle task not wiring `jvmTest` is a real gotcha and the doc note is useful. No scope creep, no publish/release automation touched.

Closes #68.

PR #70, opened 2026-07-17
Fix flaky KonstructionServiceListenerTest.oneSlowListenerDoesNotBlockOthers (CI timing race)

monkopedia-coder:
## What
`oneSlowListenerDoesNotBlockOthers` asserted `slow.contentChanges.isNotEmpty()` **immediately** after the fast listener's `awaitUntil` returned. Changed it to poll with `awaitUntil` instead.

## Why
The slow `FakeKonstructionListener` records its content change *before* parking on the gate, but its coroutine may not have been scheduled yet at the moment the fast listener finished — a scheduling race. It passes locally but flaked under CI load. It only became visible now that #69 wired `:backend:jvmTest` into CI (it had been silently un-run). Polling eliminates the race while still proving fan-out is not serialized (a serialized impl would hang the *fast* listener's `awaitUntil` above, failing first).

## Test
`:backend:jvmTest --tests "*KonstructionServiceListenerTest*"` — 5/5 consecutive `--rerun-tasks` green; spotless clean. Test-only change.

### DIFF
--- a/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/KonstructionServiceListenerTest.kt
+++ b/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/KonstructionServiceListenerTest.kt
@@ -204,10 +204,14 @@ class KonstructionServiceListenerTest {
         awaitUntil(message = "fast listener blocked by slow listener") {
             fast.contentChanges.isNotEmpty()
         }
-        assertTrue(
-            slow.contentChanges.isNotEmpty(),
-            "slow listener should have started its callback"
-        )
+        // Poll rather than assert immediately: the slow listener records its
+        // content change before parking on the gate, but its coroutine may not
+        // have been scheduled yet when the fast listener finished (a timing race
+        // that flaked under CI load). awaitUntil still proves fan-out isn't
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
Test-only fix for a real CI timing race.

**Correctness:** The race diagnosis is sound. `FakeKonstructionListener.onContentChange` records `contentChanges.add(Unit)` *before* `maybeBlock`, so the slow listener does record before parking — but its coroutine may not have been scheduled by the time the fast listener's `awaitUntil` exits. The immediate `assertTrue` raced against the scheduler; polling removes it.

**Fan-out proof preserved:** The test still proves non-serialized fan-out. A serialized dispatcher would never let the fast listener run while the slow one holds the gate, so `awaitUntil(fast.contentChanges.isNotEmpty())` would timeout first. The concurrency-proving value is intact.

**Timeout bounded:** `awaitUntil` uses `withTimeout(5_000ms)` — cannot hang forever.

**Quality:** The polling pattern is the established idiom throughout this file (used in 6 other call sites). The added comment is accurate and matches the file's existing comment density. No production code touched. CI green.

Tier 1 auto-approve.

PR #72, opened 2026-07-21
LSP: lsp-types 1.2.0 honest-nullable results + advertise client capabilities

monkopedia-coder:
## What

Brings the LSP **nullable-result + capabilities-advertising** work onto current `main` so editor completions work with modern spec-strict `kotlin-lsp` engines. Two coupled changes plus a version bump.

### 1. lsp-types 1.1.0 -> 1.2.0 (honest-nullable results)

lsp-types 1.1.0 generated ~45 spec-nullable LSP results (`completion`, `hover`, `signatureHelp`, ...) as **non-nullable**. A `null` engine response therefore threw a decode (`SerializationException`) in the calling coroutine, which cancelled the scope **shared** with the diagnostics publisher — silently killing all LSP updates. The `#52` `runCatching`/`SerializationException` guards were the stopgap.

1.2.0 types these results honestly, so a null result decodes to `null`. This PR:
- removes the `#52` `SerializationException` catches,
- rewrites the bridge overrides (`textDocumentCompletion` / `textDocumentHover` / `textDocumentSignatureHelp` / `completionItemResolve`) to **return `null`** on a null engine result,
- adds a `guardEngine` helper: a *crashed* engine (throwing delegate call) still degrades to inert ("no popup / no hover") instead of letting the exception ride up the shared frontend ksrpc connection and tear the whole editor session down. Only genuine cooperative cancellation of our own coroutine is re-thrown (`ensureActive()`).

### 2. Advertise client capabilities (`withBridgeFeatures`)

Modern kotlin-lsp builds (>= LS-262.6274.0) are **spec-strict**: they gate features on the *client* declaring support. kodemirror's LSP client initializes with **empty** `ClientCapabilities`, so:
- with no `textDocument.completion` -> the engine returns **no completions at all**, and
- with no `textDocument.diagnostic` -> the engine omits the pull `diagnosticProvider` entirely (so the pull->push publisher never even runs).

The bridge *does* service both (it forwards completion and runs the pull->push diagnostics loop), so `ClientCapabilities.withBridgeFeatures()` advertises `completion` + `diagnostic` **on the client's behalf** in `initialize` (preserving any sub-capabilities the editor did send).

## Why

Without (2), completions are simply empty on current engines. Without (1), the first `null` result from the engine throws and cancels the shared scope, killing diagnostics + completion for the rest of the session.

## ⚠️ Known risk — please evaluate before trusting on prod

This is the repo owner's original **"do not deploy yet"** concern and needs reviewer judgement:

**The bridge now returns `null` over the wire (instead of an RPC error) for a null hover / completion / signatureHelp result.** Previously kodemirror's lsp-client (0.3.5) received an RPC *error* in these cases; now it receives a literal `null`. **It is NOT yet verified that kodemirror 0.3.5 handles receiving `null` gracefully.** This needs **browser / e2e verification** (and possibly a kodemirror bump) before it can be trusted on prod. Do not deploy on the strength of the green backend build alone — the backend tests here do not exercise the frontend lsp-client's null handling.

## Scope / what's intentionally excluded

- Does **not** include `LspPrewarmer` / the App.kt prewarm launch (separate concern).
- Does **not** include any #171 debug scaffolding (`cursorColumn`/`konstructionIds` bridge fields, `[KM-HITDEBUG]`, EditorPane/EditorDiagnostics #171 changes) or the `kodemirror 0.3.5-fix171c-SNAPSHOT` pin — kodemirror stays at the released `0.3.5`, and **no `-SNAPSHOT` is introduced**.
- The only version-catalog change is `lsp` `1.1.0` -> `1.2.0`; every other dependency stays at main's current values.

## Reconciliation with #69

The WIP branched before #69. `teardown()` here keeps main's `cancelAndJoin()` (the #69 flake fix) — not the WIP's `.cancel()` — so callers freeing resources (e.g. the temp dir in tests) don't race in-flight IO coroutines. `#69` was the only main change to this file since the WIP base; it is preserved.

## Testing

- `clean shadowJar :backend:jvmTest :protocol:jvmTest spotlessCheck` is **GREEN** (JDK 21).
- The engine-dependent integration tests are gated on `KONSTRUCTOR_KOTLIN_LSP` (via `assumeTrue`) and **skip in CI** (no engine) — expected. The test file gains a whitespace-hover-suppression assertion (adapts to the new hover null-suppression) and otherwise compiles clean against the nullable API.

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/lsp/BridgeLanguageServer.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/lsp/BridgeLanguageServer.kt
@@ -18,6 +18,8 @@ package com.monkopedia.konstructor.lsp
 import com.monkopedia.hauler.error
 import com.monkopedia.hauler.hauler
 import com.monkopedia.konstructor.Config
+import com.monkopedia.lsp.ClientCapabilities
+import com.monkopedia.lsp.CompletionClientCapabilities
 import com.monkopedia.lsp.CompletionItem
 import com.monkopedia.lsp.CompletionItemTextEdit
 import com.monkopedia.lsp.CompletionParams
@@ -25,6 +27,7 @@ import com.monkopedia.lsp.ConfigurationParams
 import com.monkopedia.lsp.DefaultLanguageClient
 import com.monkopedia.lsp.DefaultLanguageServer
 import com.monkopedia.lsp.Diagnostic
+import com.monkopedia.lsp.DiagnosticClientCapabilities
 import com.monkopedia.lsp.DidChangeTextDocumentParams
 import com.monkopedia.lsp.DidCloseTextDocumentParams
 import com.monkopedia.lsp.DidOpenTextDocumentParams
@@ -43,6 +46,7 @@ import com.monkopedia.lsp.RegistrationParams
 import com.monkopedia.lsp.ServerCapabilities
 import com.monkopedia.lsp.SignatureHelp
 import com.monkopedia.lsp.SignatureHelpParams
+import com.monkopedia.lsp.TextDocumentClientCapabilities
 import com.monkopedia.lsp.TextDocumentCompletionResult
 import com.monkopedia.lsp.TextDocumentContentChangeEventVariant
 import com.monkopedia.lsp.TextDocumentIdentifier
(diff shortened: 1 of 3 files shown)


monkopedia-reviewer (approved):
Reviewed to merge standard (tier-2, owner-signoff obtained from Jason). Scope: 3 files (BridgeLanguageServer.kt, integration test, libs.versions.toml). #52 runCatching stopgap removed; spec-nullable results now honestly nullable and return null on null engine result; withBridgeFeatures gap-fills client capabilities so spec-strict kotlin-lsp engines advertise completion+diagnostic; #69 cancelAndJoin teardown preserved. Frontend null-handling verified by kodemirror against 0.3.5 lsp-client source (nullable T? on hover/completion/signatureHelp → no popup, no throw). CI green (build-and-test 11m13s). Merging.

PR #74, opened 2026-07-25
Fix ScopedShipper.requestPickup() infinite self-recursion (StackOverflowError)

### LINKED ISSUE #73: ScopedShipper.requestPickup() infinitely recurses (calls itself instead of shipper.requestPickup())
## Bug

`ScopedShipper.requestPickup(u: Unit)` calls **itself** instead of delegating to the wrapped `shipper`, producing unbounded self-recursion.

`backend/src/jvmMain/kotlin/com/monkopedia/konstructor/logging/ScopedShipper.kt:58-59`
```kotlin
override suspend fun requestPickup(u: Unit): DropBox {
    val pickup = requestPickup()   // <-- this.requestPickup() → same override → infinite recursion
```

The no-arg `requestPickup()` resolves to the generated no-arg convenience on the same receiver (`this`), which re-enters `ScopedShipper.requestPickup(Unit)` → StackOverflowError. It should call `shipper.requestPickup()`.

## Evidence it's a copy-paste slip
The two sibling overrides in the same class delegate correctly:
- `deliveries()` — `ScopedShipper.kt:30` → `shipper.deliveries()`
- `requestDockPickup()` — `ScopedShipper.kt:34` → `shipper.requestDockPickup()`

Only `requestPickup` (L59) omits the `shipper.` receiver.

## Reachability
- `ScopedShipper` is **live**: constructed via `WarehouseWrapper.getScoped()` (`WarehouseWrapper.kt:31-32`) and handed to the script subprocess as its `Shipper` through `service.setShipper(...)` at `hostservices/ScriptManager.kt:98` and `KonstructionServiceImpl.kt:109`.
- Whether the crash actually fires depends on the hauler shipping client calling `requestPickup()` (vs `requestDockPickup()`/`deliveries()`) on that scoped shipper. If it does, the backend hits an immediate StackOverflowError when a script attempts a log pickup. It is **not** a dead method — it's a reachable latent runtime fault gated on the client's shipping path.
- Note: the superficially-similar `warehouse.requestPickup().attach(scope)` at `KonstructorImpl.kt:61` is a **different, correct** path — `WarehouseWrapper.requestPickup` (`WarehouseWrapper.kt:38`) properly delegates to `warehouse.requestPickup()`.

## Fix
```kotlin
val pickup = shipper.requestPickup()
```
One-line change; consider a small test that drives a scoped-shipper pickup to lock the regression.

_Found during the Graphify A/B navigation trial (surfaced by both the control and treatment arms, verified against `main@8abac22`)._

monkopedia-coder:
## The bug

`ScopedShipper.requestPickup(u: Unit)` called `requestPickup()` with no receiver:

```kotlin
override suspend fun requestPickup(u: Unit): DropBox {
    val pickup = requestPickup()   // resolves to THIS override → infinite recursion
```

With no receiver, the unqualified `requestPickup()` resolves back to the same override rather than to the wrapped `shipper`. Each call re-enters itself, producing a `StackOverflowError`. Its sibling overrides delegate correctly — `deliveries()` → `shipper.deliveries()`, `requestDockPickup()` → `shipper.requestDockPickup()` — but this one did not. `ScopedShipper` is live (constructed via `WarehouseWrapper.getScoped()` and handed to the script subprocess via `setShipper`), so this was a reachable latent crash.

## The fix

One-line delegation, matching the siblings:

```kotlin
val pickup = shipper.requestPickup()
```

## Regression test

Added `ScopedShipperTest.requestPickupDelegatesToWrappedShipper`: wraps a fake `Shipper` whose `requestPickup()` returns a sentinel `DropBox` and counts invocations. It asserts the `ScopedShipper` delegates exactly once (before the fix this recursed and blew the stack instead of returning) and that the returned `DropBox` forwards `log` to the delegate with the scoped thread name.

## Verification

- `./gradlew :backend:jvmTest --tests "*.ScopedShipperTest"` — passing (1 test)
- `./gradlew spotlessCheck` — BUILD SUCCESSFUL

Fixes #73

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/logging/ScopedShipper.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/logging/ScopedShipper.kt
@@ -56,7 +56,7 @@ class ScopedShipper(private val tagPrefix: String, name: String, private val shi
     }
 
     override suspend fun requestPickup(u: Unit): DropBox {
-        val pickup = requestPickup()
+        val pickup = shipper.requestPickup()
         return object : DropBox {
             override suspend fun log(logEvent: Box) {
                 pickup.log(
--- a/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/logging/ScopedShipperTest.kt
+++ b/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/logging/ScopedShipperTest.kt
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2022 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.monkopedia.konstructor.logging
+
+import com.monkopedia.hauler.Box
+import com.monkopedia.hauler.DeliveryService
+import com.monkopedia.hauler.DropBox
+import com.monkopedia.hauler.Level
+import com.monkopedia.hauler.LoadingDock
+import com.monkopedia.hauler.Shipper
+import kotlin.test.Test
+import kotlin.test.assertTrue
+import kotlinx.coroutines.test.runTest
+
+class ScopedShipperTest {
+
+    /**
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
One-line delegation bug fix + regression test — both axes clean.

**Correctness:** `requestPickup()` → `shipper.requestPickup()` is the exact fix: unqualified `requestPickup()` in the class body resolves back to the same override (no extension to shadow it), causing infinite recursion. The fix matches the delegation pattern of the two sibling methods (`deliveries()` → `shipper.deliveries()`, `requestDockPickup()` → `shipper.requestDockPickup()`).

**Test:** `ScopedShipperTest.requestPickupDelegatesToWrappedShipper` wraps a call-counting fake `Shipper`, asserts delegation happens exactly once (would StackOverflow on the old code), and validates the returned `DropBox` correctly prefixes the thread name. Exercises the delegation and the failure path — not tautological.

**CI:** build-and-test pass. No style ledger violations.

PR #76, opened 2026-07-25
LSP bridge: self-heal after a kotlin-lsp engine crash (#54)

### LINKED ISSUE #54: LSP bridge doesn't auto-reconnect after a kotlin-lsp engine crash (stays inert until file reopen)
## Summary

After the kotlin-lsp engine subprocess crashes mid-session, the `BridgeLanguageServer` for an open konstruction stays **inert** until the user reopens the file — it does not auto-reconnect to the (respawned) engine. LSP features (completion, hover, diagnostics) silently stop working for that editor session even though the warm subprocess has been respawned.

## Background

`BridgeLanguageServer.delegate` (the subprocess-facing stub) is set once in `initialize()` from `KotlinLspProcess.connect(...)`. When the engine subprocess dies, `KotlinLspProcess` detects it and respawns on the *next* `connect()` — but an already-open bridge keeps pointing at the dead stub. Every forward call to it now fails.

This is no longer *fatal*: the engine-crash isolation guard (`BridgeLanguageServer.guardEngine`, foreign-cancellation aware) makes those failures degrade to inert instead of cascading into a frontend-connection teardown (see ksrpc#228). So the editor session survives — but LSP stays dark for that document until it's closed and reopened (which builds a fresh bridge against the live engine).

## Desired behavior

After an engine crash/respawn, the bridge should transparently re-establish its delegate against the respawned subprocess and re-`didOpen` the current document, so LSP resumes without the user having to reopen the file. Candidate approaches:

- On a guarded forward call detecting a dead delegate, attempt a single `KotlinLspProcess.connect(...)` refresh + re-synthesize/`didOpen` the open doc, then retry once.
- Or a lightweight liveness check that rebinds `delegate` when `KotlinLspProcess` reports a respawn.

## Notes

- Must stay resilient regardless of engine stability (lsp-kotlin's guidance: treat engine crash/respawn as expected).
- Separately worth checking whether engine-build staleness correlates with crash frequency (the engine binary expires ~40 days; a fresher build may reduce crashes — but resilience is the primary fix).
- Follow-up to the connection-cascade isolation work; not blocking (current behavior is safe, just not self-healing).

monkopedia-coder:
## Problem

After the kotlin-lsp engine subprocess crashes mid-session, `BridgeLanguageServer.delegate` is left pointing at the dead subprocess stub. `KotlinLspProcess` respawns the subprocess on the *next* `connect()`, but the already-open bridge never reconnects — every forward call fails, `guardEngine` degrades it to inert (no cascade — the session survives, per #72/ksrpc#228), and LSP (completion/hover/diagnostics) stays **dark for that document until the user closes+reopens the file**.

## Fix: single-flight lazy re-init on the next request

A forward call that fails now transparently self-heals instead of silently staying inert:

- **`withLiveDelegate`** wraps every forward call. On failure it attempts a **single-flight `reconnect`**, then retries the call **once** against the fresh delegate. Still dead ⇒ degrade to inert exactly as before.
- **`reconnect`** re-`connect`s a fresh subprocess-facing stub (the process manager respawns the dead subprocess), replays the engine `initialize`/`initialized` handshake (rebuilding the pull publisher), and re-`didOpen`s the **current** document (kept live by didOpen/didChange), so the engine resumes analyzing what the user has open.
- Engine acquisition is funneled through one **`connectEngine()`** seam shared by `initialize` and `reconnect`.

## Concurrency guard

- **`reconnectLock`** (a `Mutex`) single-flights the reconnect: N concurrent failing requests spawn **one** respawn, not N. The reconnect is keyed on the dead stub (`delegate !== deadServer`), so a coroutine that observes someone else already reconnected is a no-op.
- A **`torndown`** flag, set under the same lock in `teardown`, prevents an in-flight reconnect from **resurrecting the delegate after the scope is cancelled** (which would leak the fresh subprocess connection).

## Preserved from #72

- The `guardEngine` **no-cascade** contract: a dead engine never throws out of an LSP method to tear down the shared frontend ksrpc connection; foreign `CancellationException` is not propagated (`ensureActive()` gate).
- `teardown`'s `scope` **`cancelAndJoin`** semantics (the delegate release now happens under `reconnectLock`, the join stays after).

## Deliberately omitted (minimal by design)

No backoff / retry-limit / "reconnecting" editor state. Each failing request makes **at most one** reconnect+retry; recovery is naturally paced by request traffic (a later request that still fails simply tries again). Can be added later if crash frequency warrants — but per the issue, resilience is the primary fix and this is the simplest shape that works.

## Tests

Adds `BridgeLanguageServerReconnectTest` — **CI-runnable, engine-free** regression coverage that drives the crash→respawn transition with a fake in-process delegate (via the `connectEngine()` seam):
- a crashed delegate self-heals on the next request and retries once (LSP resumes, no reopen),
- reconnect replays the **current edited** content's didOpen (not the stale open text),
- after teardown the bridge does **not** resurrect the delegate,
- a still-failing respawn stays inert **without throwing**.

The real-engine half stays gated in `BridgeLanguageServerIntegrationTest` (skips without `KONSTRUCTOR_KOTLIN_LSP`).

## Verification

- `:backend:jvmTest` — green (new test: 4/4, 0 skipped)
- `spotlessCheck` — green

Heads-up: this is LSP/UX-adjacent (delicate lifecycle/concurrency area), so it may be tier-2 for review.

Fixes #54

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/lsp/BridgeLanguageServer.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/lsp/BridgeLanguageServer.kt
@@ -17,6 +17,7 @@ package com.monkopedia.konstructor.lsp
 
 import com.monkopedia.hauler.error
 import com.monkopedia.hauler.hauler
+import com.monkopedia.hauler.info
 import com.monkopedia.konstructor.Config
 import com.monkopedia.lsp.ClientCapabilities
 import com.monkopedia.lsp.CompletionClientCapabilities
@@ -62,6 +63,8 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.cancelAndJoin
 import kotlinx.coroutines.ensureActive
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import kotlinx.serialization.json.JsonNull
 
 /**
@@ -116,8 +119,22 @@ import kotlinx.serialization.json.JsonNull
  * [delegate] is null and the bridge degrades to an inert server: [initialize] returns
  * empty capabilities, notifications are dropped. The frontend is flag-gated anyway, so
  * with the flag off this class is never constructed.
+ *
+ * **Self-heal after an engine crash (#54).** The kotlin-lsp subprocess can crash
+ * mid-session; [KotlinLspProcess] then respawns it on the next [connect][KotlinLspProcess.connect].
+ * The bridge's [delegate], however, is set once in [initialize] and would keep pointing at the
+ * dead stub — so every forward call fails, [guardEngine] degrades it to inert, and LSP stays
+ * dark for that document until it is closed+reopened. To self-heal, a forward call that fails
+ * ([guardEngine]) triggers a single-flight [reconnect]: it re-[connect][KotlinLspProcess.connect]s
+ * a fresh delegate against the (respawned) subprocess, re-runs the engine handshake, and
+ * re-`didOpen`s the current document, then the failed call is retried once against the fresh
+ * delegate. The reconnect is guarded by [reconnectLock] (single-flight — concurrent failing
+ * requests don't spawn N handshakes) and is itself failure-isolated (a still-dead engine just
+ * leaves the bridge inert, exactly as before — it never throws out to kill the shared frontend
+ * connection). Deliberately minimal: no backoff/retry-limit machinery — each failing request
+ * makes at most one reconnect+retry attempt, so recovery is naturally paced by request traffic.
  */
-class BridgeLanguageServer(
+open class BridgeLanguageServer(
     private val config: Config,
     private val workspaceId: String,
     private val konstructionId: String,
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (commented):
I've reviewed this PR to merge standard — the reconnect logic, concurrency model, and tests are all sound. The only residual items below need @Monkopedia's judgment (design tradeoffs, not bugs).

## What I fully reviewed and confirmed sound

**No hot reconnect loop.** `withLiveDelegate` makes at most one reconnect + one retry per failing call. Recovery is paced by request traffic. No spin.

**Single-flight guard is correct.** `reconnectLock.withLock { if (delegate !== deadServer) return@withLock delegate }` genuinely deduplicates N concurrent failures: the first acquires the lock, swaps `delegate`, the rest observe the swap and skip. One engine spawned, not N.

**No cascade (#72 preserved).** The `coroutineContext.ensureActive()` gate in `withLiveDelegate` re-propagates cooperative cancellation of *this* coroutine only; a foreign CancellationException from the dead engine leg is caught and swallowed. The shared frontend ksrpc connection can't be killed by a dead LSP engine.

**Teardown resurrection prevention is correct.** `torndown` is set under `reconnectLock` before `delegate` is nulled; a reconnect queued behind teardown sees `torndown = true` and bails. Post-teardown `delegate` is null, so `delegate !== deadServer` is also true — dual guard.

**`torndown` visibility is correct despite lacking `@Volatile`.** It's only read/written while holding `reconnectLock` (a `Mutex`), which provides the necessary happens-before ordering. `@Volatile` is not needed here.

**No lock-held-across-suspend deadlock.** `kotlinx.coroutines.sync.Mutex` is coroutine-friendly — holding it across a `suspend` call doesn't block a thread. Callers queue on the Mutex and suspend, not deadlock.

**Test genuinely exercises crash→reconnect→retry.** The `FakeEngine.crashed` flag + `FakeEngineBridge.connectEngine()` seam exercise the full path: first engine throws, `withLiveDelegate` catches, calls `reconnect`, reconnect runs `handshakeEngine` + `replayDidOpen` against the second engine, retry lands on the second engine. All four test scenarios pin the PR's correctness claims.

---

## Three points for @Monkopedia's judgment

**1. Orphaned `PullDiagnosticsPublisher` after reconnect (design tradeoff, not a correctness bug).**
When `reconnect` runs `handshakeEngine`, it reassigns `diagnostics` to a new publisher wired to the fresh engine. The *old* publisher is not explicitly cancelled. Its bounded cold-index backoff (`maxAttempts=40`, `interval=4s`) will keep firing failing pulls against the dead server for up to 160 seconds, each caught and logged as an error. The new publisher is correct; the old one is benign but noisy.

The `PullDiagnosticsPublisher` has no exposed cancel/close method, so the PR can't stop it without either: (a) adding a `cancel()` method to `PullDiagnosticsPublisher`, or (b) explicitly nulling `diagnostics` before calling `handshakeEngine` (which orphans the old publisher and lets it spin out its backoff naturally, same as now). Both are small additions. Or: accept the bounded noise (40 error logs over 160s after each crash) as the deliberate minimal design. Your call.

**2. The `torndown` concurrent test is not covered.**
The test exercises the sequential case (teardown completes, *then* a call arrives → `delegate` is already null → early exit). The concurrent case (reconnect in-flight when teardown runs, race between the two mutex holders) is not directly tested. The logic is simple enough to verify by inspection (torndown=true under lock → reconnect bails), but a targeted concurrent test would add confidence. Minor gap — not blocking, but worth noting.

**3. Per-request reconnect attempt with no retry limit — intentional, but confirm.**
Every failing request attempts exactly one reconnect+retry, indefinitely. If the engine never recovers (e.g. the process manager can't respawn), each cursor move or keystroke triggers a silent reconnect attempt (connect subprocess + handshake). The PR explicitly calls this "minimal by design" and says "can be added later if crash frequency warrants." Just confirming you're OK with the tradeoff that a permanently-dead engine produces one reconnect attempt per user interaction forever (not once, not with a cooldown), until the session is closed.

---

CI: `build-and-test` — **PASS**. The four new unit tests ran clean, zero skipped.

monkopedia-reviewer (approved):
Approved by the repo owner (tier-2 signoff). Reviewed at tier-1 standard with no blocking findings: the reconnect is single-flight behind a Mutex, teardown wins the race via the `torndown` flag, there's no hot retry loop, and no #72-style cascade.

Owner accepted the three flagged tradeoffs as-is:
1. The orphaned `PullDiagnosticsPublisher` keeps polling the dead engine after a reconnect (bounded ~40 × 4s).
2. Concurrent teardown-during-reconnect is covered only by sequential tests.
3. No retry limit or cooldown — minimal by design.

Merging.

PR #79, opened 2026-07-26
Drop :protocol's phantom jvmMain/jvmTest blocks (keeps logback out of the fat JAR)

### LINKED ISSUE #78: logback-classic ships in backend-all.jar and collides with HaulerServiceProvider (dual SLF4J providers in prod)
## Symptom (confirmed in production)

adolin's live server log prints, on every start:

```
SLF4J(W): Class path contains multiple SLF4J providers.
SLF4J(W): Found provider [com.monkopedia.konstructor.lib.logging.HaulerServiceProvider@...]
SLF4J(W): Found provider [ch.qos.logback.classic.spi.LogbackServiceProvider@...]
```

SLF4J then picks one **non-deterministically**. Which one wins decides whether backend logs route through hauler (the intended path) or through an unconfigured logback — and there is no `logback.xml` anywhere in the repo, so the logback branch is silently misconfigured.

## Root cause

`protocol/build.gradle.kts` declares full `jvmMain` and `jvmTest` dependency blocks for source sets that **contain zero files** (`protocol/src` has only `commonMain/` and `commonTest/`). One of those phantom blocks pulls in `logback-classic`:

- `protocol/build.gradle.kts:49-57` — `jvmMain` block: `kotlin("stdlib")`, `kotlin("reflect")`, `slf4j-api`, serialization core+json, **`logback-classic`**
- `protocol/build.gradle.kts:63-65` — `jvmTest` block for a source set that doesn't exist

Chain to the shipped artifact:

1. `protocol/build.gradle.kts:56` puts logback-classic on protocol's JVM `runtimeElements` (`implementation` is excluded from `apiElements` but **not** from `runtimeElements`).
2. `backend/build.gradle.kts:48` `implementation(project(":protocol"))` → logback lands on backend's `jvmRuntimeClasspath`.
3. `backend/build.gradle.kts:151` `configurations = listOf(...getByName("jvmRuntimeClasspath"))` → logback is packed into `backend-all.jar`.
4. `backend/build.gradle.kts:49` also pulls `:lib`, which registers the competing binding via `lib/src/main/resources/META-INF/services/org.slf4j.spi.SLF4JServiceProvider` → `HaulerServiceProvider`.
5. `backend/build.gradle.kts:163` `mergeServiceFiles()` merges both entries into one file in the fat JAR.

## Verified against the built artifact

```
$ unzip -l backend/build/libs/backend-all.jar | grep -c 'ch/qos/logback'
730
$ unzip -p backend/build/libs/backend-all.jar META-INF/services/org.slf4j.spi.SLF4JServiceProvider
com.monkopedia.konstructor.lib.logging.HaulerServiceProvider
ch.qos.logback.classic.spi.LogbackServiceProvider
```

Related history: `CHANGELOG.md:36-43` describes fixing a `HaulerServiceProvider` early-bind crash — the same area. This is a latent second half of that problem.

## Proposed fix

- Delete `protocol/build.gradle.kts:49-57` and `:63-65` entirely. If the JVM variant genuinely needs `slf4j-api`, keep only that one line and drop `logback-classic`.
- Also drop the unused `kotlinx-serialization-json` from commonMain (`:45-46`) — commonMain only uses `@Serializable`, never `Json`.
- Verify after: rebuild the fat JAR, confirm `grep -c 'ch/qos/logback'` is 0 and the merged service file lists only `HaulerServiceProvider`; confirm the `SLF4J(W)` warnings are gone from the server log on startup.

Found during a read-only repo analysis at `8abac22`, then confirmed empirically against the built jar and the running prod server.

monkopedia-coder:
Fixes #78.

## What was wrong

`protocol/src` contains only `commonMain/` and `commonTest/` — there is no `jvmMain` or `jvmTest` source directory. Both source sets nonetheless declared full dependency blocks, and the `jvmMain` one pulled in `logback-classic`.

`implementation` deps are excluded from `apiElements` but **not** from `runtimeElements`, so logback rode protocol's JVM runtime variant into backend's `jvmRuntimeClasspath`, and from there into `backend-all.jar` (shadowJar packs exactly that configuration). `:lib` registers hauler's `SLF4JServiceProvider` via `META-INF/services`, `mergeServiceFiles()` merged both entries, and the shipped server printed on every start:

```
SLF4J(W): Class path contains multiple SLF4J providers.
SLF4J(W): Found provider [com.monkopedia.konstructor.lib.logging.HaulerServiceProvider@...]
SLF4J(W): Found provider [ch.qos.logback.classic.spi.LogbackServiceProvider@...]
```

SLF4J then picked a binding non-deterministically — and there is no `logback.xml` anywhere in the repo, so the logback branch was silently unconfigured.

## The change

Delete the two dead blocks in `protocol/build.gradle.kts` — 12 lines, no other files touched.

`backend/src/.../App.kt:46` imports `org.slf4j.LoggerFactory` and backend does not declare slf4j-api itself; it gets it transitively through hauler (`api` in commonMain), which is why nothing needs to re-declare anything. Confirmed by the build below.

I deliberately left the `kotlinx-serialization-json` line in `commonMain` alone even though #78 mentions it as a secondary cleanup — it's unrelated to the defect and `commonTest` does use Json, so it belongs in its own change if wanted.

## Verification (on the built artifact, not just the build log)

```
$ unzip -l backend-all.jar | grep -c 'ch/qos/logback'
0                                     # was 730

$ unzip -p backend-all.jar META-INF/services/org.slf4j.spi.SLF4JServiceProvider
com.monkopedia.konstructor.lib.logging.HaulerServiceProvider    # was both providers

$ unzip -l backend-all.jar | grep -c 'org/slf4j/LoggerFactory.class'
1                                     # slf4j-api still there, as needed
```

Plus a runtime smoke test of the built jar (`--http 8099`, temp `KONSTRUCTOR_HOME`): starts, serves HTTP 200, and the log contains **zero** `SLF4J(W)` lines — logs come out in hauler's format.

Gradle: `:protocol:jvmTest :backend:jvmTest shadowJar` and `spotlessCheck` all pass. `lib-all.raj` and the `web/` bundle are still present in the jar.

(`:protocol:wasmJsBrowserTest` fails on this box with "test task did not discover any tests" — a pre-existing browser/Karma config issue unrelated to this change; the JVM test tasks are what CLAUDE.md and CI run.)

### DIFF
--- a/protocol/build.gradle.kts
+++ b/protocol/build.gradle.kts
@@ -46,22 +46,10 @@ kotlin {
         implementation(libs.kotlinx.serialization.json)
         implementation(libs.kotlinx.coroutines.core)
     }
-    sourceSets["jvmMain"].dependencies {
-        implementation(kotlin("stdlib"))
-        implementation(kotlin("reflect"))
-        implementation(libs.slf4j.api)
-
-        implementation(libs.kotlinx.serialization.core)
-        implementation(libs.kotlinx.serialization.json)
-        implementation(libs.logback.classic)
-    }
     sourceSets["commonTest"].dependencies {
         implementation(kotlin("test"))
         implementation(libs.kotlinx.coroutines.test)
         implementation(libs.kotlinx.serialization.json)
-    }
-    sourceSets["jvmTest"].dependencies {
-        implementation(kotlin("test-junit"))
     }
 }
 


monkopedia-reviewer (approved):
Verified empirically against the PR-branch worktree (`.claude/worktrees/issue78`, JDK 21), not just the config diff:

**1. Shipped artifact actually changed.** `backend-all.jar` from this branch vs. the pre-fix jar in `backend/build/libs/`:
- `ch/qos/logback` class count: 730 → 0
- `META-INF/services/org.slf4j.spi.SLF4JServiceProvider`: was both `HaulerServiceProvider` + `LogbackServiceProvider`, now only `HaulerServiceProvider`
- `org/slf4j/LoggerFactory.class` still present (1) — slf4j-api didn't disappear

**2. Logging still works, not just single-provider.** Ran the built jar (`java -jar backend-all.jar --http 8098 --websockets`, scratch `KONSTRUCTOR_HOME`): started cleanly, `curl` returned HTTP 200, log output came through in hauler's format (`2026-07-25T23:20:05.329 INFO (main) ...`), zero `SLF4J(W)` warning lines. Killed the process after.

**3. Phantom blocks weren't load-bearing.** `protocol/src` only has `commonMain`/`commonTest` on disk — confirmed no `jvmMain`/`jvmTest` directories exist and nothing else in the repo references them. `:protocol:jvmTest :backend:jvmTest :e2e:compileTestKotlin` all pass clean on this branch. Traced where backend now gets slf4j-api from (App.kt imports `LoggerFactory` but backend declares neither slf4j nor logback itself): confirmed via `:backend:dependencies --configuration jvmRuntimeClasspath` — `org.slf4j:slf4j-api:2.0.18` arrives transitively through `lib`'s `api(libs.hauler)` and independently through ktor's own transitive deps. Solid either way.

**4. Scope discipline.** PR deliberately left `kotlinx-serialization-json` in `commonMain` alone (issue #78's secondary suggestion). Confirmed: that import is unused in `commonMain` but IS used in `commonTest` (`KonstructorTypesTest.kt`), so removing it would've needed more care than a drive-by. Correct call to defer it to a separate change — keeps this PR a clean, reviewable 12-line diff.

Category: `bug_fix` (fixes a real non-deterministic SLF4J-provider defect via dependency config) → tier 1 under konstructor's policy (also tier 1 if classified `infra`). No UI/UX or performance surface, nothing security/sandboxing-adjacent — no reviewer-judgment bump warranted. No style-guide rule violations. CI green (`build-and-test` SUCCESS), `mergeStateStatus` CLEAN.

Approving.

PR #83, opened 2026-07-30
LSP: tie engine-scoped work to a connection CoroutineScope (#80)

### LINKED ISSUE #80: LSP: tie engine-scoped work to a connection CoroutineScope instead of letting it poll on after a crash
Follow-up from PR #76 (merged, `65f5418`), which added self-heal reconnect for the kotlin-lsp engine.

## Current behavior

When the engine dies and `reconnect()` swaps in a fresh delegate, the previous `PullDiagnosticsPublisher` is not cancelled. It keeps polling and logging against the dead engine until its own retry budget runs out (~40 polls × 4s ≈ 160s). It's bounded and harmless, but it's noise, and every future piece of engine-scoped work would need its own equivalent countdown to stay bounded.

## Proposed shape (owner's suggestion)

Attach a `CoroutineScope` to the engine connection itself, launch engine-scoped work (the diagnostics publisher, and anything similar added later) in that scope, and cancel the scope when the connection fails or is replaced. Structured concurrency then handles teardown for everything attached to that connection, instead of each consumer carrying a bounded-retry loop.

That also subsumes tradeoff 3 from #76's review (no retry limit / cooldown): a connection whose scope is cancelled can't keep respawning work, so the "engine crashes repeatedly" case stops being an open-ended loop.

## Notes

- `BridgeLanguageServer` already has the seams this would build on: `connectEngine()`, the `reconnectLock` Mutex, and the `torndown` flag.
- Related review tradeoff also left open in #76: concurrent teardown-during-reconnect is currently covered only by sequential tests. A connection-scoped design is easier to test under real concurrency, so it's worth revisiting that test at the same time.
- Not urgent — the current behavior is bounded and accepted.

monkopedia-coder:
Fixes #80.

## The problem

PR #76 gave the bridge self-heal reconnect, but the *work* attached to a connection was owned by the bridge, not the connection. When the engine died and `reconnect()` swapped in a fresh delegate, the previous `PullDiagnosticsPublisher` kept polling the corpse until its own 40-attempt countdown ran out (~160s of failing pulls and error logs). Bounded, but only because that one consumer carried a countdown — anything else scoped to the engine would have needed its own.

## The change

`EngineConnection` (a private holder in `BridgeLanguageServer`) owns the subprocess stub, a `CoroutineScope` that every piece of work launched on that engine's behalf runs in, and the pull publisher fed by it. `reconnect()` and `teardown()` cancel that scope, so structured concurrency stops everything attached to a dead connection in one move.

Net effect on the bridge's state: `delegate` + `diagnostics` collapse into one `connection` field, and the bridge-wide `scope` (whose only child was ever the publisher) is deleted.

**Deleted mechanism.** With cancellation doing the bounding, the publisher's retry countdown is redundant: the `ColdIndexBackoff(maxAttempts = 40, interval = 4s)` data class collapses to a single `coldIndexRetry: Duration` constructor knob, and `pullWithColdIndexBackoff` becomes `pullUntilPublished` — it runs until it publishes, the doc closes, or its scope is cancelled. This is the tradeoff-3 item left open in #76's review: a cancelled connection can't keep respawning work.

## Cancellation correctness

- `reconnect` cancels the dead scope **without joining it** — it must not block (while holding `reconnectLock`) on work stuck against a dead engine. `teardown` still `cancelAndJoin`s, because its callers free resources right after (the temp dir in tests).
- **Cancelling the old scope can't kill the reconnect that replaces it**: `withLiveDelegate` is the only path into `reconnect`, and it is only ever called from editor-facing LSP methods — never from inside a connection scope. That invariant is now stated in its KDoc.
- `pullOnce` no longer swallows its own `CancellationException` inside `runCatching` (which would have logged our cancellation as an engine error). It uses the same `coroutineContext.ensureActive()` idiom as `guardEngine`, so *our* cancellation propagates while a *foreign* one (the dead engine leg surfacing through a call) is still treated as an engine failure, not a cascade.
- `teardown`'s `torndown` flag still wins: a reconnect queued behind teardown bails, and post-teardown `connection` is null so it can never `===` a non-null dead connection.
- Nothing here closes the subprocess-facing stub — it rides `KotlinLspProcess`'s shared keep-warm connection, same as before — so a cancelled scope can't leave a half-closed engine.

## Tests

- `BridgeLanguageServerReconnectTest.reconnect cancels the work scoped to the dead connection` — the fake engine now advertises pull mode and parks its `textDocument/diagnostic` on a gate that never completes, recording if that pull is cancelled. The test waits for the pull to park, crashes the engine, drives a completion (which self-heals), then asserts the parked pull was cancelled and that pulling moved to the fresh connection. **Verified to fail against the pre-change bridge** (stash the production file → `TimeoutCancellationException`, other 4 reconnect tests still green).
- `PullDiagnosticsPublisherTest.the cold-index retry runs until its scope is cancelled` — pins the contract that replaced the countdown: a permanently-failing first pull keeps retrying while the connection lives, and stops dead when the injected scope is cancelled.

## Verification

```
JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew :backend:jvmTest :protocol:jvmTest  → BUILD SUCCESSFUL (137 tests, 0 failures, 15 skipped/gated)
JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew spotlessCheck                       → BUILD SUCCESSFUL
```

The real-engine integration test (`-Dintegration=true` + `KONSTRUCTOR_KOTLIN_LSP`) was **not** run: no engine binary is present on this box (`/home/jmonk/kotlin-lsp` does not exist).

## Known tradeoff

The cold-index retry is now unbounded in wall-clock terms for one narrow case: an engine that stays *connected* but never answers a pull with a `full` report will be re-polled every 4s for as long as the doc is open (previously it gave up after ~160s). A crash is covered — the first editor-facing call detects it and cancels the scope — but a fully idle user in front of a wedged-but-alive engine would keep the retry going. Routing the publisher's pulls through the self-heal path would close that too, but not without the reconnect cancelling its own caller, so it is deliberately left out here.

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/lsp/BridgeLanguageServer.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/lsp/BridgeLanguageServer.kt
@@ -59,10 +59,11 @@ import com.monkopedia.lsp.ksrpc.LifecycleState
 import kotlin.coroutines.coroutineContext
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
 import kotlinx.coroutines.cancelAndJoin
 import kotlinx.coroutines.ensureActive
+import kotlinx.coroutines.job
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlinx.serialization.json.JsonNull
@@ -102,8 +103,8 @@ import kotlinx.serialization.json.JsonNull
  *  - Pulls are triggered on didOpen + debounced didChange, and on the engine's
  *    `workspace/diagnostic/refresh` — overridden on the backend [Forwarder] (the
  *    `DefaultLanguageClient` base THROWS for it; the frontend client's is a harmless
- *    no-op, a different object) to re-pull every open doc. A bounded cold-index backoff
- *    covers only the FIRST pull before the index warms.
+ *    no-op, a different object) to re-pull every open doc. A cold-index retry covers only
+ *    the FIRST pull before the index warms.
  *  - `previousResultId` is threaded per doc; an `unchanged` report does not republish
  *    (only a `full` report emits `publishDiagnostics`).
  *
@@ -116,14 +117,14 @@ import kotlinx.serialization.json.JsonNull
  * client-bound request from inside [initialize] (deadlock risk).
  *
  * If the subprocess is unavailable (binary unset/missing or spawn failure),
- * [delegate] is null and the bridge degrades to an inert server: [initialize] returns
+ * [connection] is null and the bridge degrades to an inert server: [initialize] returns
  * empty capabilities, notifications are dropped. The frontend is flag-gated anyway, so
  * with the flag off this class is never constructed.
  *
  * **Self-heal after an engine crash (#54).** The kotlin-lsp subprocess can crash
  * mid-session; [KotlinLspProcess] then respawns it on the next [connect][KotlinLspProcess.connect].
- * The bridge's [delegate], however, is set once in [initialize] and would keep pointing at the
- * dead stub — so every forward call fails, [guardEngine] degrades it to inert, and LSP stays
+ * The bridge's engine [connection], however, is set once in [initialize] and would keep pointing
+ * at the dead stub — so every forward call fails, [guardEngine] degrades it to inert, LSP stays
  * dark for that document until it is closed+reopened. To self-heal, a forward call that fails
  * ([guardEngine]) triggers a single-flight [reconnect]: it re-[connect][KotlinLspProcess.connect]s
  * a fresh delegate against the (respawned) subprocess, re-runs the engine handshake, and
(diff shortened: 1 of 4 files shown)


monkopedia-reviewer (commented):
**Review verdict: approve-quality — but this is tier-2, so it is NOT approved. It needs the repo owner's signoff.**

Recording the verdict here so it has a durable trace; this is a comment, not an approval, and the merge gate is unchanged.

Reviewed SHA: `110dd9a4272a3438492439e173a5e42eed129be7` (merge-base `af98d0e`).

**What was verified empirically, not just read:**
- **The new test really guards the bug.** Reverse-applied only `BridgeLanguageServer.kt` while keeping the new tests: `reconnect cancels the work scoped to the dead connection` fails with `TimeoutCancellationException`; the other four stay green.
- **The parked pull is genuinely cancelled**, not merely pointed at a dead stub.
- **No scope outlives its connection** — no `GlobalScope`, no bare `CoroutineScope()`, no injected app-level scope reachable from engine-scoped work.
- **Diagnostics re-establish on the fresh connection** after a crash, so recovery is not silently diagnostics-dead.
- **The `withLiveDelegate` → `reconnect` invariant** documented in the new KDoc was checked against the actual call sites, not taken on the comment's word.
- `ColdIndexBackoff` is gone repo-wide; no stragglers.
- Local suite reproduces the PR's tallies: 137 tests, 0 failures, 15 skipped. `spotlessCheck` green.

**One deliberate tradeoff for the owner to rule on.** The deleted `ColdIndexBackoff` capped cold-index retry at ~40 attempts x 4s (~160s), after which pull diagnostics were permanently abandoned for that document. Retry is now bounded by scope cancellation instead of wall-clock. A crashed engine is covered — the next editor-facing call detects the crash and cancels the scope. The uncovered case is an engine that stays CONNECTED but never returns a `full` report: that is re-polled every 4s for as long as the document stays open. Trading a bounded-but-gives-up retry for an unbounded-but-scope-cancelled one is the right default here, but it is a behavior change worth stating out loud rather than burying in a deletion.

monkopedia-reviewer (approved):
Owner signoff received: Jason accepted the unbounded-but-scope-cancelled cold-index retry over a retry ceiling, on the specific ground that the retry cannot outlive its connection — the failure is bounded by client behaviour rather than by a counter. He declined a follow-up issue for the ceiling, so none is filed.

Approving on that signoff. Review verdict was approve-quality at this SHA (110dd9a); nothing has changed since — mergeStateStatus CLEAN, build-and-test pass.

Recorded downside, accepted knowingly: a long-lived connection against a permanently broken index will retry inside its scope indefinitely without surfacing an error. If that turns out to be reachable in practice rather than theoretically, that is new information worth raising fresh, not a reopening of this decision.

PR #90, opened 2026-08-07
Migrate the Shadow plugin 8.3.6 → 9.6.1 (#64)

### LINKED ISSUE #64: Migrate Shadow plugin 8.3.6 → 9.x (com.gradleup.shadow major bump)

monkopedia-coder:
Migrates the Shadow plugin **8.3.6 → 9.6.1** (current stable; the issue body said 9.5.1, the 2026-08-07 spike ran 9.6.1).

**This is a revival of the original spike branch, rebased onto `main` and rewritten** — not a fresh PR — because Jason's ruling was *"when there's time rebase and merge on green CI"* and the spike branch is what it names. The only rebase conflict was `shadow`/`slf4j` sitting on adjacent lines in `gradle/libs.versions.toml`; `main`'s `slf4j = "2.0.18"` was kept.

⚠️ **The spike's price of "~2 lines" was wrong.** The version bump plus `register` → `named` builds green *and quietly damages the artifact*. Both problems were found by diffing the fat JAR's contents against an 8.3.6 baseline; neither is visible from `BUILD SUCCESSFUL`, and the e2e suite passes with both present.

## What Shadow 9 actually required

| # | Change | Why |
|---|---|---|
| 1 | `shadow = "9.6.1"` | the bump |
| 2 | `tasks.register("shadowJar")` → `tasks.named(...)` | 9.x auto-registers a `shadowJar` task for this KMP-jvm module; `register` fails with *task already exists* |
| 3 | **drop `from(jvmJar.map { it.outputs })`** | 9.0.0 aligned `ShadowJar.from` with Gradle's `AbstractCopyTask.from`, which copies a jar as an **opaque file** instead of unzipping it |
| 4 | **`filesMatching("META-INF/services/**") { duplicatesStrategy = INCLUDE }`** | 9.x actually honours `duplicatesStrategy`, default `EXCLUDE`, which drops duplicates **before** transformers see them — `mergeServiceFiles()` silently stops merging |

### #3 — the nested-jar regression

With the `from(...)` line kept, `backend-all.jar` gains a top-level `backend-jvm-0.3.0.jar` entry that no classloader can read: **44.9 MB → 63.9 MB.** The auto-registered task already pulls in the jvm main output, so the line is redundant as well as harmful. `:backend:shadowJar --dry-run` confirms `copyJsBundleToKtor`, `:lib:shadowJar`, `copyLibToKtor`, `jvmProcessResources` and `jvmMainClasses` are all still in the task graph without it.

### #4 — `mergeServiceFiles()` stops merging

Shadow 9.5.0 added a build-time warning for exactly this, and it fired for **all nine** `META-INF/services` files in this build. Measured damage before the fix:

```
META-INF/services/com.github.ajalt.mordant.terminal.TerminalInterfaceProvider
  8.3.6:  3 providers (…jna.TerminalInterfaceProviderJna,
                       …ffm.TerminalInterfaceProviderFfm,
                       …nativeimage.TerminalInterfaceProviderNativeImage)   270 B
  9.6.1:  1 provider  (…jna.TerminalInterfaceProviderJna)                    84 B
```

**Note for the merge condition recorded on the issue:** the SLF4J file named there is a *false negative* for this check. Only `lib` ships `META-INF/services/org.slf4j.spi.SLF4JServiceProvider`, so it is never a merge and survives a broken merger byte-identical — it looked fine in the damaged jar. The mordant file is the real positive control, and it is the one that caught this.

## Artifact verification — `backend-all.jar`, 8.3.6 baseline vs. this branch

Both built from the same tree (`origin/main` @ `7a45c61`), outputs deleted first, all tasks executed.

- **entries 14790 → 14787**, zero duplicate entries in either
- **removed (4):** `META-INF/INDEX.LIST`, `META-INF/versions/9/module-info.class`, `META-INF/versions/11/module-info.class`, and the now-empty `META-INF/versions/11/` dir — Shadow 9 excludes `module-info.class` under Multi-Release folders by default. A stale `INDEX.LIST` in a fat jar is actively wrong, so both are improvements.
- **added (1):** `META-INF/services/` — a zero-byte directory entry.
- of the **14786 shared entries, exactly 3 differ byte-for-byte**:
  - `META-INF/MANIFEST.MF` — gains `Multi-Release: true` (new 9.x default: injected when any dependency is an MR jar)
  - `META-INF/services/reactor.blockhound.integration.BlockHoundIntegration` — 672 → 668 B; two blank comment lines dropped by the transformer, provider line identical
  - `lib-all.raj` — because `lib-all.jar` itself changed the same way (7920 → 7919 entries: drops its own stray `META-INF/versions/9/module-info.class`, gains `Multi-Release: true`). Its service files are byte-identical.
- `lib-all.raj` verified as a valid zip **in place inside the fat jar**: 7919 entries, `HaulerServiceProvider` present, 149 kcsg classes, `testzip()` clean
- both wasm bundles (`web/*.wasm`), `web/frontend.js`, `web/index.html` and the Compose font resource present and **byte-identical**

## The JAR was exercised, not just built

`java -jar backend-all.jar --http <port> --cors --websockets`:

- boots and reaches `Responding at http://0.0.0.0:<port>` in ~1s
- `GET /` → 200, the **identical 3695-byte** `index.html`, served out of the jar's own resources
- `GET /frontend.js` → 200, **1267973 bytes**, matching the baseline exactly
- log lines come out in hauler's format, i.e. the SLF4J provider genuinely resolves from the merged service file

The same check was run against the 8.3.6 baseline jar as a control; both behave identically.

## Behaviour changes worth knowing

- **`Multi-Release: true` is now set** on both `backend-all.jar` and `lib-all.raj`. Under 8.3.6 the manifest lacked it while `META-INF/versions/9/…/BuiltInsResourceLoader.class` was present, so kotlin-reflect's versioned class was dead weight; it will now actually be preferred on Java 9+. That is upstream kotlin-reflect's intended behaviour, but it is a real runtime change rather than a no-op.
- **`kotlin_module` `DuplicatesStrategy` warnings remain in the build log** (~80 of them, from `KotlinModuleMetadataTransformer`). They are advisory: every `META-INF/*.kotlin_module` entry in the built jar is byte-identical to the 8.3.6 baseline, so nothing regressed — 8.3.6 did not merge them either. Silencing them would mean switching those to `INCLUDE`, which *would* change the output, so it is deliberately left out of scope.
- **`backend-jvm-0.3.0.jar` is no longer written to `backend/build/libs/`** as a side effect of `shadowJar`, since the task no longer depends on `jvmJar`. Nothing consumes it: `e2e` and `publish.yaml` both reference `backend-all.jar`.
- **`CHANGELOG.md` is untouched.** Its `0.3.0` section lists shadow under *"Held (intentionally not bumped) — blocked by issue #64"*; that is a record of a shipped release, so it is left as written rather than rewritten.

## Not verified here

- No local `:e2e:test` run (needs the Mesa-EGL override on this box); CI runs it against this jar.
- No local `:backend:jvmTest -Dintegration=true`; CI runs it.
- Not deployed to adolin.

Fixes #64


monkopedia-coder:
## Result — #64 priced: TAKE (small migration)

Third run (`31224916305`) **completed in 11m52s**, all steps EXECUTED and passed (not a no-op — the 11m52s vs the earlier 54s fast-fails is the tell):

```
success  Spotless check
success  Build shadowJar            <- fat jar assembles under Shadow 9.6.1
success  Run backend + protocol unit tests
success  Run e2e tests under Xvfb   <- launches backend-all.jar & drives Playwright against it
```

**The e2e step is the strong evidence:** it boots the Shadow-9-built `backend-all.jar` and runs the browser suite against it, so a green e2e means the migrated jar **assembles AND boots AND serves** — not just "compiles."

**Total migration cost:** the diff on this branch — a one-line version bump (`8.3.6 → 9.6.1`) plus one `register` → `named` in `backend/build.gradle.kts` (Shadow 9.x auto-registers the `shadowJar` task; `lib` already used `named`). The `ShadowJar` class did NOT move packages (stayed `com.github.jengelman…`, verified against `shadow-gradle-plugin-9.6.1.jar`).

**Verdict: TAKE** — ~2 lines, CI green end-to-end. Not urgent (8.3.6 works; no user-facing gain), but cheap and low-risk whenever wanted.

### One residual to check BEFORE a real merge (a green here does NOT establish it)
`mergeServiceFiles()` must still collapse to the single `HaulerServiceProvider` `META-INF/services` entry that #79 hardened (logback-out-of-jar). e2e green shows the app runs, but SLF4J can resolve even with a regressed service file, so this needs inspecting the built jar's `META-INF/services/org.slf4j.spi.SLF4JServiceProvider` directly — a 30-second check when promoting this to a real PR.

⛔ Still a SPIKE — this branch prices the decision; it does not merge itself.

monkopedia-coder:
## CI green — fresh run, post-recovery

Run [`33008186050`](https://github.com/Monkopedia/Konstructor/actions/runs/33008186050) — `status=completed`, `conclusion=success`, `jobs.total_count=1`, **22m45s**. Every step executed and passed:

```
success  Spotless check
success  Build shadowJar          <- CLEAN build; validates dropping from(jvmJar…)
success  Run backend + protocol unit + integration tests   (-Dintegration=true)
success  Run e2e tests under Xvfb <- boots this backend-all.jar & drives Playwright
```

The `Build shadowJar` step matters beyond "it compiles": it is a **cold** build, so it proves `copyJsBundleToKtor` / `:lib:shadowJar` / `copyLibToKtor` / `jvmProcessResources` still order correctly ahead of `shadowJar` now that the task no longer depends on `jvmJar`. Locally that was only checked via `--dry-run` on a warm tree.

Started after Actions was confirmed `operational` again, so nothing here comes from the outage window.

**Not merging.** Green CI was the ruling's precondition, not authorisation — this is left open for independent review.

monkopedia-reviewer (approved):
## Independent review — verdict: **APPROVE**

I rebuilt this from scratch in my own worktrees and measured everything below myself. I did not take the PR body on trust; every number here is one I produced. **Every substantive claim in the PR body reproduces.** Three findings of my own are at the bottom, all non-blocking.

Method: three worktrees off `8f9dd82` and its parent `7a45c61`, outputs deleted first, all tasks `executed` (no `UP-TO-DATE`, no `FROM-CACHE`), fresh artifact mtimes checked against the clock.

- **BASE** — `7a45c61`, Shadow 8.3.6
- **BARE** — the bare bump: 8.3.6 tree + `shadow = "9.6.1"` + `register` → `named` **only**, i.e. this PR with both fixes reverted (diffed against BASE to confirm that is the entire delta)
- **PR** — `8f9dd82` as-is

---

## 1. Both regressions are real. I reproduced them independently.

### Nested-jar bloat — confirmed

| | `backend-all.jar` | top-level `backend-jvm-0.3.0.jar` entry |
|---|---|---|
| BASE (8.3.6) | 44,884,309 B | absent |
| **BARE (bare bump)** | **63,878,480 B** | **present, 18,804,751 B** |
| PR | 45,126,234 B | absent |

Exactly the reported 44.9 MB → 63.9 MB. Upstream corroborates the mechanism: Shadow 9.0.0, *"**BREAKING CHANGE:** Align the behavior of `ShadowTask.from` with Gradle's `AbstractCopyTask.from` … it will not unzip the files, only copy the files as-is."*

### `mergeServiceFiles()` silently stops merging — confirmed

`META-INF/services/com.github.ajalt.mordant.terminal.TerminalInterfaceProvider`, provider lines:

| BASE | BARE | PR |
|---|---|---|
| **3** (jna, ffm, nativeimage) 270 B | **1** (jna only) 84 B | **3**, byte-identical to BASE |

Shadow's own warning fired for **all nine** service files in the BARE build and for **zero** in the PR build. Upstream 9.0.0 again: the default `duplicatesStrategy` went `EXCLUDE` → `INCLUDE` and then *"is changed back to `EXCLUDE` in 9.0.1"*, with the explicit note *"Some `ResourceTransformer`s (e.g. `ServiceFileTransformer`) do not work with `EXCLUDE`"* and the recommendation *"Use `filesMatching` to override the default strategy for specific files."* The fix in this PR is exactly the remedy upstream prescribes.

I also checked the pattern lines up: `ServiceFileTransformer`'s own default include pattern in the 9.6.1 jar is `META-INF/services/**` — identical to the `filesMatching` pattern here, so the override is neither wider nor narrower than the transformer it is feeding.

---

## 2. The SLF4J false-negative claim — **independently confirmed, twice, by two different methods**

This was the claim I was least willing to take on trust, so I constructed the broken state and checked.

**Method A — construct the failure and run the check against it.** The BARE jar has a demonstrably dead merger (mordant 3 → 1). Against that jar:

```
META-INF/services/org.slf4j.spi.SLF4JServiceProvider
  BASE sha256: 12ad658bbe4ff3a81a70a89d574c6299ffc5f9da8f9e273562432f8fcc251dd0
  BARE sha256: 12ad658bbe4ff3a81a70a89d574c6299ffc5f9da8f9e273562432f8fcc251dd0
  content (both): com.monkopedia.konstructor.lib.logging.HaulerServiceProvider
```

**Byte-identical.** The acceptance criterion recorded on #64 — *"`mergeServiceFiles()` must still collapse to the single `HaulerServiceProvider` entry"* — **passes cleanly on a jar whose service-file merger is broken.** It is an instrument that cannot fail for the defect it was written to catch.

**Method B — count the contributors.** I dumped `backend:jvmRuntimeClasspath` (87 jars) and scanned every one for `META-INF/services/*`:

| service file | contributing jars |
|---|---|
| `…mordant…TerminalInterfaceProvider` | **3** — `mordant-jvm-jna-jvm`, `mordant-jvm-ffm-jvm`, `mordant-jvm-graal-ffi-jvm` |
| `org.slf4j.spi.SLF4JServiceProvider` | **1** — `lib-0.3.0.jar` |
| the other 7 | **1** each |

So the SLF4J file is structurally never a merge, and **the mordant file is the only genuine multi-contributor merge in the entire build** — the only available positive control. The author's reading is correct on both counts.

**My verdict on the merge condition:** it was performed, it passed, and the PR is right that passing it proves nothing. It is satisfied *in substance* by the mordant measurement, which is a real positive control that demonstrably goes red when the merger breaks. Nothing here needs to go back to the desk — the condition was met and then some. But the record on #64 should not be left implying the SLF4J check did the work.

---

## 3. Artifact content diff — my counts match the PR body exactly

⚠️ One methodological correction first, which I hit and which changes how this must be measured: **the wasm/JS bundle is not reproducible across different build directories.** The small wasm embeds absolute source paths in its debug section (`File '<Konstructor:frontend>/…/wt/<dir>/frontend/src/wasmJsMain/…/ThreeJsRenderer.kt'`), which changes its content hash, which changes its filename, which changes `frontend.js`. My first cross-directory diff therefore showed a renamed wasm and a 22-byte `frontend.js` delta — 4 of those bytes were literally `base` → `pr90`. Three worktrees produced three different wasm hashes.

So I rebuilt the PR head **in the same directory** as the baseline build and re-diffed. Clean result:

```
BASE  14790 entries, 0 duplicates
PR    14787 entries, 0 duplicates

REMOVED (4):  META-INF/INDEX.LIST
              META-INF/versions/9/module-info.class
              META-INF/versions/11/module-info.class
              META-INF/versions/11/            (now-empty dir)
ADDED   (1):  META-INF/services/               (zero-byte dir)

of 14786 shared entries, exactly 3 differ:
  META-INF/MANIFEST.MF                                   71 -> 92 B   (+ Multi-Release: true)
  META-INF/services/…BlockHoundIntegration              672 -> 668 B
  lib-all.raj                                     12973954 -> 13100294 B
```

Every number matches the PR body. The BlockHound delta is exactly two blank `#` comment lines; provider line identical. 8 of the 9 service files are byte-for-byte identical to the 8.3.6 baseline. `web/index.html`, `web/frontend.js`, `web/frontend.js.map` and **both** `web/*.wasm` are byte-identical. `META-INF/INDEX.LIST` in the baseline is a stale `JarIndex-Version: 1.0` naming `netty-codec-4.2.16.Final.jar` — wrong in a fat jar; dropping it is an improvement.

**`lib-all.raj`, opened in place inside the fat jar:** valid zip, `testzip()` clean, 7,919 entries, `META-INF/services/org.slf4j.spi.SLF4JServiceProvider` → `HaulerServiceProvider`, `HaulerServiceProvider.class` present, kcsg classes present. Diffing `lib-all.jar` BASE vs PR: 7920 → 7919, one entry removed (`META-INF/versions/9/module-info.class`), one entry changed (manifest gains `Multi-Release: true`), **all 6 service files and all 50 `.kotlin_module` entries byte-identical**.

---

## 4. `Multi-Release: true` — the PR is right to call it a real change, and I confirmed it is safe

The PR flags this rather than waving it through, which is the correct call. I verified both halves.

**It genuinely changes class resolution.** Reading `…/builtins/BuiltInsResourceLoader.class` back through the fat jar's own classloader on JDK 21:

```
BASE (8.3.6): 2234 bytes, classfile major 50   <- root entry
PR   (9.6.1):  896 bytes, classfile major 53   <- META-INF/versions/9 entry
```

Not cosmetic. The versioned variant is a different implementation — its constant pool references `Class.getModule()` / `Module.getResourceAsStream`, which the major-50 root class does not.

**And it behaves identically here.** Instantiating it and calling `loadResource("kotlin/kotlin.kotlin_builtins")` off each jar returns the same stream and the same 30,920 bytes, and `Reflection.getOrCreateKotlinClass` works on both. On JVM 17 target / JDK 21 runtime this is kotlin-reflect's intended path — under 8.3.6 the versioned class was shipped but unreachable dead weight. Note this now applies to `lib-all.raj` too, which is the classpath for the user-script compile subprocess; CI's `-Dintegration=true` compile+execute tests and the e2e suite both ran against this jar.

---

## 5. The JAR was exercised — I booted it, with an 8.3.6 control

`java -jar backend-all.jar --http <port> --cors --websockets`, temp `KONSTRUCTOR_HOME`:

| | PR (9.6.1) | BASE (8.3.6) control |
|---|---|---|
| boot | `Responding at http://0.0.0.0:…`, ~0.2 s | same |
| `GET /` | 200, **3695 B**, sha `3336965966…` | 200, **3695 B**, sha `3336965966…` |
| `GET /frontend.js` | 200, **1268056 B**, sha `8d8f606cf1…` | 200, **1268056 B**, sha `8d8f606cf1…` |
| `SLF4J(W)` lines | **0** | **0** |

Served bytes `cmp`-identical between the two, and the served `index.html` hash matches the `web/index.html` entry inside the jar — so it is genuinely serving from its own resources. Log lines come out in hauler's format, i.e. the SLF4J provider really does resolve out of the merged service file.

---

## 6. CI — verified as the triple, on this head

```
run 33008186050:  status=completed  conclusion=success  jobs.total_count=1
head_sha = 8f9dd82e3528c705782f7cf1827f343d9dec93f7   (== current PR head)
run_attempt = 1,  event = pull_request
all 17 steps success, including "Build shadowJar", the -Dintegration=true tests, and e2e under Xvfb
```

---

## 7. The `kotlin_module` warnings really are advisory

Confirmed the substance: **81** `.kotlin_module` entries in `backend-all.jar` and **50** in `lib-all.jar` — **0 differing** from the 8.3.6 baseline in either. Nothing regressed; 8.3.6 did not merge them either. Leaving them out of scope is right.

One correction to the PR body: the count is **131**, not ~80 — and it splits exactly `81` from `:backend:shadowJar` + `50` from `:lib:shadowJar`, matching the entry counts one-for-one. All 131 come from `KotlinModuleMetadataTransformer`; **zero** are on service files after the fix.

---

## Findings of my own — all non-blocking

**(a) There is still no automated guard for this defect class, and that is the one thing that worries me.** Both regressions build green, pass unit + integration tests, and pass e2e. The only thing that caught them was a manual artifact diff against a baseline — which will not happen on the next Shadow bump, or on a dependency change that adds a second contributor to some other service file. Worth a follow-up issue: `failOnDuplicateEntries = true` on the task (Shadow 9 offers it, and it is the natural companion to a scoped `INCLUDE` — see below), plus a small test asserting the mordant service file still lists 3 providers. Not a merge blocker; this PR is strictly better than what is on `main` today.

**(b) One latent edge in the `filesMatching` fix, currently inert.** `ServiceFileTransformer` includes `META-INF/services/**` but *excludes* `META-INF/services/org.codehaus.groovy.runtime.ExtensionModule` (verified in the 9.6.1 jar). A file matched by the `filesMatching` `INCLUDE` but skipped by the transformer would be emitted as duplicate zip entries. No Groovy on this classpath and I measured **0 duplicate entries** in the PR jar, so nothing is wrong today — but `failOnDuplicateEntries = true` would convert that from "someone has to notice" into a build failure. Same follow-up as (a).

**(c) An undocumented behaviour change worth a line in the PR body.** Shadow 9 wires `assemble` to `shadowJar` (upstream #1524). Measured via `--dry-run`:

```
8.3.6  :backend:assemble  ->  :backend:jvmJar        (no backend shadowJar)
9.6.1  :backend:assemble  ->  :backend:shadowJar     (and :lib:shadowJar)
```

So a plain `./gradlew build` / `assemble` on backend now builds the whole fat jar including the frontend wasm distribution. Harmless — CI invokes `shadowJar` explicitly, and I re-verified nothing consumes `backend-jvm-*.jar` (`publish.yaml`, `e2e/build.gradle.kts` and `README.md` all reference `backend-all.jar` only; the stale `backend-jvm-0.3.0.jar` left in `build/libs` from the previous build was **not** regenerated, mtime unchanged) — but it belongs in the "behaviour changes worth knowing" list.

**(d) Two cosmetic corrections to the PR body**, neither affecting any conclusion: the warning count is 131 not ~80 (see §7), and the served `frontend.js` size is environment-dependent — I measure 1268056 B on both jars on my box where the PR reports 1267973 B. The *equality between the two jars* is the load-bearing fact and it holds.

---

**Approving.** The diff is 20 lines and the risk was never in the diff; the verification behind it is what this needed, and it holds up under independent reconstruction. The two regressions are real, are fixed, and would have shipped silently under the "~2 lines, CI-green" framing from the original spike. Catching and reporting the false negative in its own acceptance criterion — rather than quietly banking the pass — is the right instinct and the most valuable thing in this PR.

*Not merging, per instruction. Verdict only.*

Issue #82, opened 2026-07-29
CI green signals that measure nothing: 9 integration tests never run, 4 e2e tests assert nothing, missing baselines pass silently


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)
**Category:** test-gap (green signals that measure nothing)

**Where:** `.github/workflows/ci.yaml:51`; `e2e/src/test/kotlin/com/monkopedia/konstructor/e2e/{SaveFlowTest,ScreenshotTest,ManualVerificationTest,MigrationVisualRegressionTest}.kt`

Three independent mechanisms by which PR CI currently reports PASS without checking anything. Each was verified on `main` (`af98d0e`).

---

### 1. Nine integration tests are gated behind a flag CI never sets

`CompileAndExecuteIntegrationTest.kt:41-45`, `SubprocessFailureTest.kt:77-80` and `BridgeLanguageServerIntegrationTest.kt:70-73` all `@Before { assumeTrue(..., System.getProperty("integration") == "true") }`. `ci.yaml:51` runs:

```
./gradlew test :backend:jvmTest :protocol:jvmTest
```

— no `-Dintegration=true`. So **14 test methods have never executed on any PR**, including the three that cover compile+execute end to end and the six covering subprocess-failure handling. `:backend:jvmTest` is green regardless.

**Verified they pass and are cheap.** Locally on `main`:

```
./gradlew shadowJar
./gradlew :backend:jvmTest -Dintegration=true \
  --tests '*CompileAndExecuteIntegrationTest*' --tests '*SubprocessFailureTest*'
→ CompileAndExecuteIntegrationTest tests=3 skipped=0 failures=0
→ SubprocessFailureTest           tests=6 skipped=0 failures=0
BUILD SUCCESSFUL in 42s
```

`BridgeLanguageServerIntegrationTest` has a *second* gate on `KONSTRUCTOR_KOTLIN_LSP` pointing at a real `intellij-server`, so it keeps self-skipping in CI — correctly, since that engine binary isn't available there.

**Proposed:** add `-Dintegration=true` to the `ci.yaml:51` step. One line, +42s, +9 real tests, and the LSP test stays skipped on its own env gate. (The `shadowJar` step already runs first, and `jvmProcessResources` depends on `copyLibToKtor` anyway, so the prerequisites are in place.)

---

### 2. Four e2e tests in the PR gate assert nothing — one of them catches its own failure

[...]

PR #91, opened 2026-08-08
⛔ SPIKE — DO NOT MERGE: measure the -Dintegration CI gate (#82a)

monkopedia-coder:
**⛔ SPIKE — DO NOT MERGE.** Prices #82(a): CI never passes `-Dintegration`, so ~14 `assumeTrue(-Dintegration)` methods silently skip and CI is green over them. This adds `-Dintegration=true` to the unit-test step so they actually run — measuring how much of the gate CI can execute and whether the runnable half reveals anything.

## The number this produces
Executed/passed/skipped counts for the integration methods, from the test report — **not** the job badge.

## Prediction (to test, not assert)
- `CompileAndExecuteIntegrationTest`, `SubprocessFailureTest` → should EXECUTE (need kotlinc/subprocess, not an engine).
- `BridgeLanguageServerIntegrationTest` → still SKIPS: it `assumeTrue`s on BOTH `-Dintegration` AND `KONSTRUCTOR_KOTLIN_LSP`, and there's no engine on the runner. That asymmetry is part of the answer.
- `KonstructionControllerSoakTest` → still skips (`-Dsoak` not set — deliberately left off; soak has a 2h timeout).

## Decision it prices
- Runnable integration tests **pass** → the gate is cheap to turn on for that subset (CompileAndExecute + subprocess); TAKE that much.
- Any **red** → a real bug the gate was hiding — the highest-value outcome.
- LSP integration stays skipped → the honest cost of engine-less CI: those need a runner-side engine (a bigger, separate decision), so `-Dintegration` alone does not make CI's LSP coverage real.

## Case AGAINST
Turning the gate on commits to KEEPING it green — a flaky integration red then blocks merges, and some of these touch the compile/execute subprocess which can be timing-sensitive. Worth confirming they're deterministic (run twice) before wiring it in for real.

## Does NOT establish
- #82(b) assertion-free e2e tests and #82(c) missing-baseline silent-pass — separate sub-findings, not touched here.
- Whether the runnable integration tests are STABLE (this is one run; failure counts on subprocess-shaped tests can vary — a second run is needed if any red appears).

Read the executed counts from the uploaded report; verdict reported separately.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,7 +48,13 @@ jobs:
         run: ./gradlew shadowJar
 
       - name: Run backend + protocol unit tests
-        run: ./gradlew test :backend:jvmTest :protocol:jvmTest
+        # SPIKE (#82): add -Dintegration=true to turn ON the integration gate that
+        # CI never passes today, to measure how many of the assumeTrue-gated
+        # integration methods actually execute + pass on a runner. The LSP
+        # integration tests additionally require KONSTRUCTOR_KOTLIN_LSP (a real
+        # engine), so those will still skip here — that asymmetry is part of the
+        # answer. DO NOT MERGE.
+        run: ./gradlew test :backend:jvmTest :protocol:jvmTest -Dintegration=true
 
       - name: Run e2e tests under Xvfb
         run: xvfb-run --auto-servernum --server-args='-screen 0 1280x720x24' ./gradlew :e2e:test -Pe2e


monkopedia-coder:
## Result — #82(a) priced: TAKE (18 hidden passing tests surfaced for one line)

Run `31230955572`, 12m29s, `success`. But the badge isn't the finding — here are the EXECUTED counts from the run's JUnit XML (fresh download):

| integration class | tests | skipped | failures |
|---|---|---|---|
| WorkspaceLifecycleTest | 3 | 0 | 0 |
| SubprocessFailureTest | 6 | 0 | 0 |
| ListenerIntegrationTest | 3 | 0 | 0 |
| KonstructionLifecycleTest | 3 | 0 | 0 |
| CompileAndExecuteIntegrationTest | 3 | 0 | 0 |
| **BridgeLanguageServerIntegrationTest** | 5 | **5 (still skipped)** | 0 |
| KonstructionControllerSoakTest | 1 | **1 (still skipped)** | 0 |

**`-Dintegration=true` turns 18 integration tests from skipped → executed, and all 18 PASS.** No hidden failures — the coverage was correct-but-unobserved (same shape as lsp-kotlin's mingwX64: never broken, only never run). The 5 `BridgeLanguageServerIntegrationTest` methods stay skipped because they also `assumeTrue(KONSTRUCTOR_KOTLIN_LSP)` and there's no engine on the runner — predicted, and it's the honest boundary of engine-less CI. Soak stays skipped (`-Dsoak` off, deliberate — 2h timeout).

**Verdict: TAKE.** Add `-Dintegration=true` to ci.yaml's test step — 18 real integration tests execute green for a one-line change, and #82's "9 never run" actually undercounted (it's 18 that this flag alone unlocks, +5 LSP +1 soak needing more infra). Cost: ~37s over the baseline green run (#90 was 11m52s; this 12m29s).

### Caveats (green does NOT establish these)
- **Single run.** `SubprocessFailureTest` is subprocess/timing-shaped; per the run-twice rule, confirm the 18 are deterministic before wiring the flag in permanently.
- **LSP integration coverage is still zero in CI** — the 5 skipped need a runner-side kotlin-lsp engine, a separate/bigger decision than this flag.
- #82(b) assertion-free e2e and #82(c) missing-baseline-silent-pass are untouched here.

⛔ SPIKE — prices #82(a); does not merge itself. The real fix is one line in ci.yaml.

monkopedia-coder:
Superseded by #94, which implements #82(a) (adds -Dintegration=true to the CI unit-test step) with the same measurement this spike produced, plus the group-A demonstrate-to-fail proof (9 tests skip without the flag; RED with the flag on a broken assertion). Closing this DO-NOT-MERGE spike; the branch stays if anyone wants the raw spike history. Refs #82.

Issue #63, opened 2026-07-16
Refresh dependencies to latest stable (Kotlin patch, ktor, coroutines, serialization, compose-multiplatform, koin, lifecycle, misc)


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)
**Category:** dep-bump

**Where:** `gradle/libs.versions.toml` — routine latest-stable bumps (STABLE only; pre-releases excluded). Verified against Maven Central / Gradle Plugin Portal metadata on 2026-07-16.

| Coordinate (`[versions]` key) | Current | Latest stable | Jump |
|---|---|---|---|
| `kotlin` | 2.4.0 | 2.4.10 | patch |
| `kotlin-coroutines` | 1.10.2 | 1.11.0 | minor |
| `kotlin-serialization` | 1.10.0 | 1.11.0 | minor |
| `kotlin-datetime` | 0.7.1 | 0.8.0 | minor |
| `ktor` | 3.4.2 | 3.5.1 | minor |
| `koin` | 4.2.0 | 4.2.2 | patch |
| `compose-multiplatform` | 1.10.3 | 1.11.1 | minor |
| `lifecycle` | 2.10.0 | 2.11.0 | minor |
| `clikt` | 5.0.3 | 5.1.0 | minor |
| `slf4j` | 2.0.17 | 2.0.18 | patch |
| `logback` | 1.5.18 | 1.5.38 | patch |
| `spotless` | 8.4.0 | 8.8.0 | minor |

**Notes / sequencing:**
- The `compose-multiplatform` bump generally wants the matching Compose-compiler; here the compose-compiler plugin is pinned to `kotlin` (2.4.x), so verify compose-multiplatform 1.11.1 is compatible with Kotlin 2.4.10 before landing.
- **First-party (coordinate via bus, not withheld):** `ksrpc` 1.1.1 → 1.1.5 (patch) and `hauler` 0.4.0 → 0.4.2 (patch) both have newer stable releases published. `kcsg` (0.4.4), `ktlint` (1.8.0), and the `javafx` plugin (0.1.0) are already at latest stable — no change.
- **Explicitly excluded from this pass:** `kodemirror` and `lsp` (active local WIP / debug pin — intentionally not touched here).
- `shadow` (8.3.6 → 9.x) is a MAJOR bump with a real migration surface and is filed as its own separate `dep-bump` issue.

[...]

PR #94, opened 2026-08-10
ci: run integration tests so the compile+execute gate is real (part a of #82)

monkopedia-coder:
Refs #82 (part a of three — does not close the issue)

## What

`.github/workflows/ci.yaml` ran the unit-test step without `-Dintegration`, so the nine integration tests covering the core compile+execute and subprocess-failure paths self-skipped via `assumeTrue` on every PR — CI reported green over code that never ran. This adds `-Dintegration=true` to that step.

- `CompileAndExecuteIntegrationTest` (3) + `SubprocessFailureTest` (6) now execute in CI.
- `BridgeLanguageServerIntegrationTest` stays skipped: its second gate needs a real kotlin-lsp engine, absent on the runner.
- Soak stays off (`-Dsoak` unset; 2h timeout).

## Scope
One line in `ci.yaml` (plus an explanatory comment). This is **part (a)** of #82; part (b) the four assertion-free e2e tests and part (c) the silent missing-baseline pass are separate follow-ups. #82 stays open — parts (b) and (c) are not addressed here.

## Verification (local, build slot, on merged main including the #63 dep bumps)

**The nine tests pass deterministically** — ran the subset twice with the flag:
```
RUN 1: CompileAndExecuteIntegrationTest tests=3 failures=0 | SubprocessFailureTest tests=6 failures=0
RUN 2: CompileAndExecuteIntegrationTest tests=3 failures=0 | SubprocessFailureTest tests=6 failures=0
```

**The gate is demonstrated to bite** (a gate that has never rejected anything is indistinguishable from the no-op it replaces):
- *Without* the flag — the tests silently skip (the bug this fixes):
```
CompileAndExecuteIntegrationTest tests=3 skipped=3
SubprocessFailureTest           tests=6 skipped=6
=> BUILD SUCCESSFUL  (green while testing nothing)
```
- *With* the flag, after breaking one assertion on purpose (`assertEquals(SUCCESS, ...)` inverted to `FAILURE`) — the suite goes RED naming the test:
```
CompileAndExecuteIntegrationTest tests=3 failures=1
=> BUILD FAILED
```
The injected break was ephemeral (never committed) and reverted; the working tree ended clean with only `ci.yaml` changed, so there is no probe SHA to strip.

**Cost:** +~42s on the unit-test step (measured; the `shadowJar` step already runs first, so the resource prerequisites are in place).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -47,8 +47,12 @@ jobs:
       - name: Build shadowJar (produces the backend fat jar e2e tests run against)
         run: ./gradlew shadowJar
 
-      - name: Run backend + protocol unit tests
-        run: ./gradlew test :backend:jvmTest :protocol:jvmTest
+      - name: Run backend + protocol unit + integration tests
+        # -Dintegration=true runs the compile+execute and subprocess-failure
+        # integration tests (9 methods) that otherwise self-skip via assumeTrue.
+        # BridgeLanguageServerIntegrationTest stays skipped (its second gate needs
+        # a real kotlin-lsp engine, absent on the runner); soak stays off.
+        run: ./gradlew test :backend:jvmTest :protocol:jvmTest -Dintegration=true
 
       - name: Run e2e tests under Xvfb
         run: xvfb-run --auto-servernum --server-args='-screen 0 1280x720x24' ./gradlew :e2e:test -Pe2e


monkopedia-reviewer (approved):
Approve — tier 1 (infra / CI-config gate-fix, part a of #82).

Independently verified on a worktree of head 7f4007f:
- **Scope clean**: diff is only `.github/workflows/ci.yaml` — one run-line gains `-Dintegration=true` plus an explanatory comment. Nothing else touched.
- **Gate demonstrably executes** (fresh XML, delete-first, `--rerun-tasks`):
  - WITH `-Dintegration=true`: CompileAndExecuteIntegrationTest tests=3 skipped=0 failures=0; SubprocessFailureTest tests=6 skipped=0 failures=0 → 9 real tests run.
  - WITHOUT the flag: 3+6 all `skipped`, BUILD SUCCESSFUL — i.e. the green-over-nothing bug this fixes. BridgeLanguageServerIntegrationTest correctly stays skipped on its engine gate.
- **CI green**: build-and-test passed (12m33s) on the actual PR head — so `-Dintegration=true` is not flaky in the CI environment; the integration suite ran green remotely.
- **close-refs = []** (stable-polled k=8; local regex gate on body+commits arms nothing). #94 does NOT close #82 — parts (b)/(c) remain open, as the body states.

PR #99, opened 2026-08-21
refactor(backend): make cachingEnabled/compilerOpts real Config params

### LINKED ISSUE #98: Config.cachingEnabled / compilerOpts are hardcoded constants masquerading as configuration
**Category:** dead-code (vestigial config)

**Where:** `backend/src/jvmMain/kotlin/com/monkopedia/konstructor/Config.kt:45-48`
```kotlin
val cachingEnabled: Boolean
    get() = true
val compilerOpts: String
    get() = ""
```

**Current:** Both are read-only getters returning a fixed literal — there is no way to set either. `Config` is not `@Serializable`, so this is **not** a persisted-state / migration concern.
- `cachingEnabled` is read at exactly one site: `ScriptHostImpl.kt:46` (`override suspend fun supportsCaching(u: Unit) = config.cachingEnabled`) and only ever yields `true`.
- `compilerOpts` is used at exactly one site: `CompileTask.kt:49` (`val opts = "${config.compilerOpts} ${config.runtimeOpts}"`) and always contributes an empty string — so every compile command is built with a leading space for no effect.

Verified via `git grep`: those are the only references to each.

**Proposed:** Either (a) inline the constants at the two call sites and drop the properties, or (b) if these are intended future knobs, promote them to real constructor parameters with defaults (like `executeTimeout`/`kotlinLspBinary` already are) so they are actually configurable. As written they are pseudo-configuration that can never change. Note `cachingEnabled` has a test (`ConfigTest.kt:55`) asserting it is `true`; adjust or drop alongside.

**Why:** The getters give a false impression of configurability, and `compilerOpts` injects a dead leading space into every compile invocation. Small, but it's the kind of "looks like a setting, isn't" cruft that misleads the next reader.

**Effort:** mechanical (inline) / small refactor (promote to real params)
_Filed by the nightly code-health pass. To act on it: relabel `agent-workable` (it'll flow into triage->work). To decline: close it — it won't be suggested again._

monkopedia-coder:
Fixes #98

## What

`Config.cachingEnabled` and `Config.compilerOpts` were read-only getters returning a fixed literal, so `Config` advertised two knobs that nothing could set. `compilerOpts` was worse than inert: it always contributed an empty string to `"${compilerOpts} ${runtimeOpts}"`, so every `kotlinc` invocation was built with a dead leading space.

Both are now constructor parameters with their current values as defaults — the same shape `executeTimeout` and `kotlinLspBinary` already use — and the flag string is assembled by dropping blank entries, so an unset `compilerOpts` contributes no argument at all.

The issue offered two options: (a) inline the constants and delete the properties, or (b) promote them to real parameters. This takes (b). `cachingEnabled = false` is a genuinely useful switch when diagnosing a stale STL cache, and `compilerOpts` is the natural place to pass an extra `kotlinc` flag; deleting them would remove the only place either belongs. Behaviour at the defaults is unchanged.

## Demonstrated RED

Each new test fails without the production change:

- `CompileTask.compileOpts` is extracted so the flag string is directly assertable. Restoring the old `"${compilerOpts} ${runtimeOpts}"` interpolation reproduces the defect exactly:
  `compile opts omit a blank compilerOpts entirely -> expected:<[]-cp lib.jar> but was:<[ ]-cp lib.jar>`
- `ScriptHostImplTest.testSupportsCachingFollowsConfig` builds a host from `Config(cachingEnabled = false)` and asserts `supportsCaching` reports `false`, so the flag is proven wired through `HostService` rather than hardcoded. Against the old constant getter it does not even compile — there is no such parameter.
- `ConfigTest` covers both defaults and both overrides.

## Verification

`./gradlew spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true` — **143 tests, 0 failures, 6 skipped** across 28 freshly written result XMLs (old XML deleted first; `:backend:jvmTest` executed, not `UP-TO-DATE`). Green re-confirmed after reverting the RED control.

## Notes for the reviewer

- `main` is unprotected here, so `--auto` is not a CI gate. Please poll `gh pr checks 98` to a completed SUCCESS yourself before merging, then plain `--squash --delete-branch`.
- Params are appended after `kotlinLspBinary`, so no positional caller changes meaning (only `dataDirFile` is ever passed positionally).

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/Config.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/Config.kt
@@ -40,12 +40,21 @@ class Config(
     val kotlinLspBinary: File? = (
         System.getenv("KONSTRUCTOR_KOTLIN_LSP")
             ?: System.getProperty("konstructor.kotlinLsp")
-        )?.let(::File)
+        )?.let(::File),
+    /**
+     * Whether the script host advertises STL caching to running scripts
+     * ([com.monkopedia.konstructor.lib.HostService.supportsCaching]). Turning it off
+     * makes every execution recompute its geometry, which is useful when diagnosing a
+     * stale-cache result.
+     */
+    val cachingEnabled: Boolean = true,
+    /**
+     * Extra flags handed to `kotlinc` when compiling a konstruction, ahead of
+     * [runtimeOpts]. Empty by default; whitespace-only values contribute nothing to the
+     * command line.
+     */
+    val compilerOpts: String = ""
 ) {
-    val cachingEnabled: Boolean
-        get() = true
-    val compilerOpts: String
-        get() = ""
     val runtimeOpts: String
         get() = "-cp ${LibsJar.getLibsJar(this).absolutePath} -J-Xmx4g -J-Xms4g"
     val json: Json = Json {
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/tasks/CompileTask.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/tasks/CompileTask.kt
@@ -46,7 +46,7 @@ class CompileTask(private val config: Config, private val input: File, private v
     ).use { executeThread ->
         val callSign = coroutineContext[CallSign.Key]
         runInterruptible(executeThread + (callSign ?: EmptyCoroutineContext)) {
-            val opts = "${config.compilerOpts} ${config.runtimeOpts}"
+            val opts = compileOpts(config.compilerOpts, config.runtimeOpts)
             val command = "kotlinc $opts -d ${output.absolutePath} ${input.absolutePath}"
             val (stdOut, stdError, result) = executeAndWait(command)
             if (result == 0) {
@@ -68,6 +68,17 @@ class CompileTask(private val config: Config, private val input: File, private v
     companion object {
         @OptIn(DelicateCoroutinesApi::class)
         private val hauler = hauler().asAsync(GlobalScope)
+
+        /**
+         * The `kotlinc` flag string: [Config.compilerOpts] ahead of [Config.runtimeOpts],
+         * with blank entries dropped so an unset [Config.compilerOpts] contributes no
+         * argument (and no stray leading space) to the command line.
+         */
+        internal fun compileOpts(compilerOpts: String, runtimeOpts: String): String =
+            listOf(compilerOpts, runtimeOpts)
+                .filter { it.isNotBlank() }
+                .joinToString(" ")
+
         private val errorRegex = Regex("^(.*):([0-9]+):([0-9]+): (.*)$")
         private val headerLines = KcsgScript.HEADER.split("\n").size
         private val footerLines = KcsgScript.FOOTER.split("\n").size
--- a/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/ConfigTest.kt
+++ b/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/ConfigTest.kt
@@ -18,6 +18,7 @@ package com.monkopedia.konstructor
 import com.monkopedia.konstructor.common.Space
 import kotlin.test.Test
 import kotlin.test.assertEquals
+import kotlin.test.assertFalse
 import kotlin.test.assertTrue
 import kotlin.time.Duration.Companion.minutes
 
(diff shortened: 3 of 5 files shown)


monkopedia-reviewer (approved):
Independent review — I did not write this PR and re-derived every claim from the code rather than the description.

## 1. Does a test go RED when the knob is SET and ignored?

Yes, and I proved it by mutation rather than by reading. Two mutations applied to a clean checkout of 46230bf:

- `ScriptHostImpl.supportsCaching` hardcoded back to `true` (the knob ignored):
  `ScriptHostImplTest > testSupportsCachingFollowsConfig FAILED — java.lang.AssertionError: Expected value to be false.`
- `CompileTask.compileOpts` restored to `"$compilerOpts $runtimeOpts"` (the dead leading space):
  `CompileTaskTest > compile opts omit a blank compilerOpts entirely FAILED — org.junit.ComparisonFailure: expected:<[]-cp lib.jar> but was:<[ ]-cp lib.jar>`

`12 tests completed, 2 failed`. Both bite on *setting* the value, not on the default surviving. Reverted afterwards; tree back to 46230bf with an empty `git diff`.

## 2. Does anything parse / match / hash / fingerprint the kotlinc command string?

No. Repo-wide (`git grep` + `/usr/bin/grep -r`, so gitignored trees are not silently skipped) the assembled string reaches exactly two places:

- `ExecUtil.executeAndWait` → `bash -c "$command"`, and
- `hauler.debug("Executing: $command")` — the only place the string is ever rendered, and nothing asserts on it (no test, no e2e expectation, no log matcher).

No hashing anywhere near it: the only `hashCode` in the tree is in `LoggingService`, and the STL cache key is the `hash` string computed inside kcsg from script content and handed to `ScriptHostImpl.checkCached/storeCached` — it never sees the compile command. `e2e/` contains no reference to either property.

Worth stating plainly: because the command goes through `bash -c`, the shell collapses the whitespace, so the old leading space was never delivered to `kotlinc` as an argument. The fix is real (the dead space is gone) but the executed process is byte-for-byte equivalent. Risk here is zero, not merely low.

## 3. Is `Config` serialised or persisted?

In-process only — nothing persisted, nothing migrates.

`class Config(` is a plain class (not `data`, not `@Serializable`), and no code decodes or encodes it: it is constructed at exactly one production site, `App.kt:57` `Config()`, plus test sites. Its inputs are env vars read at construction (`KONSTRUCTOR_HOME`, `KONSTRUCTOR_KOTLIN_LSP`, `KONSTRUCTOR_KOTLIN_LSP_SYSTEM_PATH`) — there is no config file. The `config.json` referenced throughout the codebase is a `Json` *instance*, used to serialise `Space` / `KonstructionInfo` / `TaskResult` into `info.json` / `compile.json` / `result.json` under `dataDir`; none of those carry a `Config` field. Adding constructor params therefore changes no on-disk shape, and no existing file on the running service is affected. Not a `data class`, so `PathController`'s `managers.getOrPut(config)` identity-keyed cache is also unchanged.

So this is tier-1: an in-process refactor with unchanged default behaviour.

## Other checks

- **Positional safety:** params are appended after `kotlinLspBinary`, and every caller in the tree passes at most `dataDirFile` positionally (`Config()`, `Config(tempDir)`, `Config(env.tempDir, executeTimeout = ...)`). No call site changes meaning.
- **`ConfigTest.testCachingEnabled`:** still meaningful — it now pins the *default* of a real parameter rather than restating a literal, and the new `testCachingEnabledIsOverridable` covers the other direction.
- **`compileOpts` correctness:** neither blank → `"-Xfoo -cp lib.jar"`; compilerOpts blank/whitespace → `"-cp lib.jar"`; both blank → `""`, which would leave `"kotlinc  -d ..."` with a double space — unreachable, since `runtimeOpts` is a computed getter that always begins `-cp `, and harmless under `bash -c` regardless.
- **Visibility/location:** `internal` in `CompileTask`'s companion, next to its only caller, reachable from `jvmTest` as an associated compilation. Right call — no wider surface than needed.

## Verification

Local, on the fleet build slot, worktree at 46230bf, old XML deleted with the `find -exec rm` form first:
`./gradlew spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true` → BUILD SUCCESSFUL, 75 actionable tasks all executed (`:backend:jvmTest` executed, not UP-TO-DATE). Freshly written XML: **28 files, 143 tests, 0 failures, 0 errors, 6 skipped**, mtimes inside the run window. Independently reproduces the PR's numbers.

CI: `build-and-test` completed **SUCCESS** in 13m19s on headSha `46230bf` (run 32493496521) — spotless, shadowJar, backend+protocol integration tests and e2e under Xvfb all green. Polled to completion rather than relying on `--auto`.

Approving and merging with a plain squash.

PR #100, opened 2026-08-21
refactor(backend): extract the shared info.json CRUD scaffolding

### LINKED ISSUE #97: Factor the duplicated CRUD scaffolding shared by KonstructorImpl and WorkspaceImpl (list-scan, generateId, create)
**Category:** duplication

**Where:**
- `backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructorImpl.kt` — `list()` (~L68-77), `create()` (~L96-110), `generateId()` (~L111-118)
- `backend/src/jvmMain/kotlin/com/monkopedia/konstructor/WorkspaceImpl.kt` — `list()` (~L38-47), `create()` (~L64-80), `generateId()` (~L82-89)

**Current:** `KonstructorImpl` (over `Space`) and `WorkspaceImpl` (over `KonstructionInfo`/`Konstruction`) are parallel services that hand-repeat the same three pieces of the on-disk-`info.json` CRUD contract, differing only in the base dir and the deserialized type:

1. **Directory scan** — identical in both:
   ```kotlin
   dir.listFiles()?.mapNotNull {
       if (!it.isDirectory) return@mapNotNull null
       val infoFile = File(it, "info.json")
       if (!infoFile.exists()) return@mapNotNull null
       infoFile.inputStream().use { input -> config.json.decodeFromStream<T>(input) }
   } ?: emptyList()
   ```
2. **`generateId()`** — byte-identical body (the `KonstructorImpl` copy is just additionally wrapped in `callContext("generateId")`):
   ```kotlin
   val usedIds = list().map { it.id }
   var id = 0
   while (usedIds.contains(id.toString())) { id++ }
   id.toString()
   ```
3. **`create()`** — same shape: `id = newItem.id.ifEmpty { generateId() }`, then the `targetInfo.exists() || targetInfo.parentFile.exists()` guard, `mkdirs()`, `encodeToStream`.

**Proposed:** Extract the shared pieces once — e.g. `inline fun <reified T> File.listInfoJson(json: Json): List<T>` for the scan, and a `firstFreeId(usedIds: Collection<String>): String` helper — and have both `create()`/`list()` implementations call them (each still supplying its own dir/type and post-decode mapping, e.g. `WorkspaceImpl` maps `KonstructionInfo -> .konstruction`).

**Why:** The "skip non-dir / skip missing info.json / stream-decode" invariant and the id-allocation logic are load-bearing persistence rules duplicated verbatim; a fix has to be made in two places or the two services drift. In particular the `generateId()` scan is O(n^2) (a `contains` over the id list per increment) in **both** copies — folding it into one helper lets that be fixed once. Verified via `git grep`: `generateId` has exactly these two definitions, each called once from its own `create()`.

**Effort:** small refactor
_Filed by the nightly code-health pass. To act on it: relabel `agent-workable` (it'll flow into triage->work). To decline: close it — it won't be suggested again._

monkopedia-coder:
Fixes #97

## What

`KonstructorImpl` (over `Space`, under the data dir) and `WorkspaceImpl` (over `KonstructionInfo`, under a workspace dir) are parallel services over the same on-disk contract — one directory per item, each holding an `info.json` — and each restated the same three pieces of it. New `InfoStore.kt` holds them once:

| helper | replaces |
| --- | --- |
| `File.listInfo<T>(json)` | the identical "skip non-directories, skip directories with no `info.json`, stream-decode the rest" scan in both `list()` implementations |
| `firstFreeId(usedIds)` | both byte-identical `generateId()` bodies |
| `writeNewInfo(target, id, json, value)` | the "file **or** its directory already present ⇒ id taken" guard, `mkdirs()`, and encode in both `create()` implementations |

`WorkspaceImpl` still supplies its own post-decode mapping (`KonstructionInfo -> .konstruction`) and its workspace-id guard; only the shared scaffolding moved.

The complexity note from the issue is fixed as part of the fold: both copies tested membership with `usedIds.contains(...)` against a `List`, so allocating an id was O(n²). `firstFreeId` hashes the ids once, making it linear.

## Behaviour

Unchanged. The existing `KonstructorImplTest` / `WorkspaceImplTest` are the behaviour-preservation evidence and pass untouched.

## Tests

`InfoStoreTest` asserts the extracted rules directly, including two the scan depends on that nothing covered before: a loose file in the container, and a half-created directory with no `info.json`, are both **skipped** rather than failing the whole listing.

**Mutation control** — the tests bite. Starting `firstFreeId` at 1 and dropping the missing-`info.json` guard produced:
```
firstFreeIdFillsTheLowestGap                    expected:<[0]> but was:<[1]>
listInfoSkipsLooseFilesAndDirectoriesWithoutInfo java.io.FileNotFoundException: .../half-created/info.json
```

## Verification

`./gradlew spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true` — **143 tests, 0 failures, 6 skipped**, fresh result XML (old deleted first, `:backend:jvmTest` executed rather than `UP-TO-DATE`).

## One thing the reviewer should know: a latent flake I hit and did not ship

An earlier revision of `InfoStoreTest` carried a wall-clock test asserting the linear id allocation finished 50k ids inside 5s. With it present, `ScopedShipperTest.requestPickupDelegatesToWrappedShipper` failed **2 times in 9 full-suite runs** with `UncaughtExceptionsBeforeTest` — an exception escaping some *earlier* test's coroutine and being attributed to the next `runTest`. Measurements:

| tree | failures / runs |
| --- | --- |
| this branch **with** the timing test | 2 / 9 |
| this branch **without** the timing test | 0 / 6 |
| this branch, production change only, no new tests | 0 / 6 |
| unrelated control branch off the same base | 0 / 6 |

I dropped the timing test — it is environment-dependent and it is the only variable that co-occurs with the failures. **That is correlation, not proof**: 0/6 cannot rule out a ~20% rate, so I am not claiming the timing test causes it. What the data does show is that the flake is not in the refactor. The underlying fragility (a leaked async exception landing on an unrelated `runTest`) is pre-existing and I am filing it separately.

## Notes for the reviewer

`main` is unprotected here, so `--auto` is not a CI gate — please poll `gh pr checks` to a completed SUCCESS yourself before merging, then plain `--squash --delete-branch`.

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/InfoStore.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/InfoStore.kt
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2022 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+@file:OptIn(ExperimentalSerializationApi::class)
+
+package com.monkopedia.konstructor
+
+import java.io.File
+import kotlinx.serialization.ExperimentalSerializationApi
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.decodeFromStream
+import kotlinx.serialization.json.encodeToStream
+
+/**
+ * On-disk layout shared by every item store: one directory per item, each holding an
+ * [INFO_JSON] describing it. [KonstructorImpl] keeps [com.monkopedia.konstructor.common.Space]
+ * dirs under the data dir; [WorkspaceImpl] keeps
+ * [com.monkopedia.konstructor.common.KonstructionInfo] dirs under a workspace.
+ *
+ * These rules are load-bearing persistence behaviour, so they live in one place rather
+ * than being restated per service.
+ */
+internal const val INFO_JSON = "info.json"
+
+/**
+ * Every immediate subdirectory of [this] that holds an [INFO_JSON], decoded as [T].
+ *
+ * Non-directories and directories with no [INFO_JSON] are skipped (a half-created or
+ * hand-made directory must not fail the whole listing), and an unreadable parent yields
+ * an empty list.
+ */
+internal inline fun <reified T> File.listInfo(json: Json): List<T> =
+    listFiles()?.mapNotNull { child ->
+        if (!child.isDirectory) return@mapNotNull null
+        val infoFile = File(child, INFO_JSON)
+        if (!infoFile.exists()) return@mapNotNull null
+        infoFile.inputStream().use { input ->
+            json.decodeFromStream<T>(input)
+        }
+    } ?: emptyList()
+
+/**
+ * The lowest non-negative integer, as a string, that is not in [usedIds].
+ *
+ * Membership is tested against a hash set, so allocating an id is linear in the number
+ * of existing items rather than quadratic.
+ */
+internal fun firstFreeId(usedIds: Collection<String>): String {
+    val used = usedIds.toHashSet()
+    var id = 0
+    while (used.contains(id.toString())) {
+        id++
+    }
+    return id.toString()
+}
+
+/**
+ * Claim [targetInfo] for a new item and write [value] to it.
+ *
+ * Both the file and its directory must be absent — an existing directory means the id is
+ * taken even when its [INFO_JSON] was never written — otherwise this throws
+ * [IllegalArgumentException] naming [id].
+ */
+internal inline fun <reified T> writeNewInfo(targetInfo: File, id: String, json: Json, value: T) {
+    if (targetInfo.exists() || targetInfo.parentFile.exists()) {
+        throw IllegalArgumentException("$id has been used already")
+    }
+    targetInfo.parentFile.mkdirs()
+    targetInfo.outputStream().use { output ->
+        json.encodeToStream(value, output)
+    }
+}
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructorImpl.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructorImpl.kt
@@ -40,8 +40,6 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlinx.serialization.ExperimentalSerializationApi
-import kotlinx.serialization.json.decodeFromStream
-import kotlinx.serialization.json.encodeToStream
 
 @OptIn(DelicateCoroutinesApi::class)
 class KonstructorImpl(private val config: Config) :
@@ -66,14 +64,7 @@ class KonstructorImpl(private val config: Config) :
     }
 
     override suspend fun list(u: Unit): List<Space> = callContext("list") {
-        config.dataDir.listFiles()?.mapNotNull {
-            if (!it.isDirectory) return@mapNotNull null
-            val infoFile = File(it, "info.json")
-            if (!infoFile.exists()) return@mapNotNull null
-            infoFile.inputStream().use { input ->
-                config.json.decodeFromStream<Space>(input)
-            }
-        } ?: emptyList()
+        config.dataDir.listInfo<Space>(config.json)
     }
 
     override suspend fun konstruction(id: Konstruction): KonstructionService =
@@ -97,24 +88,12 @@ class KonstructorImpl(private val config: Config) :
         val newItemWithId = newItem.copy(
             id = newItem.id.ifEmpty { generateId() }
         )
-        val targetInfo = newItemWithId.infoFile
-        if (targetInfo.exists() || targetInfo.parentFile.exists()) {
-            throw IllegalArgumentException("${newItemWithId.id} has been used already")
-        }
-        targetInfo.parentFile.mkdirs()
-        targetInfo.outputStream().use { output ->
-            config.json.encodeToStream(newItemWithId, output)
-        }
+        writeNewInfo(newItemWithId.infoFile, newItemWithId.id, config.json, newItemWithId)
         newItemWithId
     }
 
     private suspend fun generateId(): String = callContext("generateId") {
-        val usedIds = list().map { it.id }
-        var id = 0
-        while (usedIds.contains(id.toString())) {
-            id++
-        }
-        id.toString()
+        firstFreeId(list().map { it.id })
     }
 
     override suspend fun delete(item: Space): Unit = callContext("delete") {
@@ -134,5 +113,5 @@ class KonstructorImpl(private val config: Config) :
     fun getInputStream(target: String): InputStream = File(config.dataDir, target).inputStream()
 
     private val Space.infoFile: File
-        get() = File(File(config.dataDir, id), "info.json")
+        get() = File(File(config.dataDir, id), INFO_JSON)
 }
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/WorkspaceImpl.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/WorkspaceImpl.kt
@@ -36,14 +36,7 @@ class WorkspaceImpl(private val config: Config, private val workspaceId: String)
         get() = "Workspace"
 
     override suspend fun list(u: Unit): List<Konstruction> = callContext("list") {
-        workspaceDir.listFiles()?.mapNotNull {
-            if (!it.isDirectory) return@mapNotNull null
-            val infoFile = File(it, "info.json")
-            if (!infoFile.exists()) return@mapNotNull null
-            infoFile.inputStream().use { input ->
-                config.json.decodeFromStream<KonstructionInfo>(input).konstruction
-            }
-        } ?: emptyList()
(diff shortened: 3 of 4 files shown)


monkopedia-reviewer (approved):
Independent review — I did not write this PR. Worktree off `cc3a5ed` (detached), every number below measured here, nothing taken from the PR body.

**Verdict: the refactor is faithful and I found no defect in it.** I did find a real data-loss race in `create()` and root-caused the disclosed flake; **both are pre-existing and neither is caused or widened by this PR** — filed as #102 and evidenced on #101.

---

## 1. Is `firstFreeId` atomic with the write? No — and the guard does not reliably catch it

There is a window between "scan used ids" and "claim one", and a second, narrower window *inside* `writeNewInfo` between the `exists()` guard and `mkdirs()`. Two racers that both clear the guard before either creates the directory both proceed, and the second `encodeToStream` overwrites the first record.

I wrote a concurrent-create probe (8 racers off a `CompletableDeferred` gate, 40 rounds, fresh temp data dir per round, each racer taking its own `konstructor.get(ws)` handle as separate ksrpc calls would):

| | rounds | rounds losing >=1 record | records lost | rounds where **every** caller got success and records were still lost |
| --- | --- | --- | --- | --- |
| `KonstructorImpl.create` | 40 | 24 | 41 | 2 |
| `WorkspaceImpl.create` | 40 | 24 | 40 | 0 |

A fully silent round, verbatim from the run:

```
WS round 2: ok=8 err=0 returnedIds=[2, 0, 1, 1, 2, 4, 3, 0] distinct=5 onDisk=5
            diskNames=[0=w1, 1=w2, 2=w0, 3=w6, 4=w5] errs=[]
```

Eight creates, zero exceptions, five records. `w3` holds id `1`, which on disk is `w2`. Traffic reaches the wrong object and nothing errors. So: **sometimes** loud (`IllegalArgumentException: N has been used already` — most rounds), **sometimes** silent. `writeNewInfo`'s guard is a narrowing, not a fix.

**New or pre-existing:** pre-existing, and *not widened*. Same probe with `KonstructorImpl.kt`/`WorkspaceImpl.kt` reverted to `HEAD~1` and everything else identical: 23/39 and 27/38 rounds losing records (37 and 52 records) — statistically indistinguishable from the post-PR numbers. Per the reviewing rule that a relocated pre-existing race is a follow-up and not a blocker, that is what I have done: **#102**. If anything the hash-set lookup *shortens* the scan-to-claim window.

Two extras the probe surfaced, both in #102: `list()` can read a **torn** `info.json` (`JsonDecodingException: Expected colon ':', but had 'EOF' instead at path: $.konstruction`) and one torn file fails the *whole* listing; and `PathController` does `mkdirs()` eagerly, so an id whose directory exists without `info.json` is skipped by `listInfo` (re-offered by `firstFreeId`) but rejected by `writeNewInfo` — permanently unallocatable.

## 2. Same code, same concurrency? Nearly — and the `Mutex` does not cover create

- `KonstructorImpl` is a **single instance shared across every connection** (`App.createChannel()` serializes one `service`). Its `Mutex` guards **only** `konstructionLookup` inside `konstruction()` — `create`, `list` and `delete` run unlocked. Confirmed by reading the class; `create` is not under it.
- `WorkspaceImpl` is constructed **fresh per `Konstructor.get(id)` call**, so it could not hold useful per-instance state even if it wanted to — a lock there would have to live on the shared `KonstructorImpl`.
- Callers: `WorkspaceMutations.createWorkspace` / `createKonstruction` / `uploadStl`, each fired from `NavigationDialogViewModel.mutate` -> `viewModelScope.launch`, no client-side serialization; plus the e2e `JsBridge` actions. `KonstructionController` does not create.

Different exposure, same defect — measured above at the same rate for both. Folding them does not change that.

## 3. `writeNewInfo` — file **or** directory, and it is checked *after* the id is chosen

`if (targetInfo.exists() || targetInfo.parentFile.exists())` — the `info.json` **or** its containing directory. The check runs on `newItemWithId.infoFile`, i.e. strictly after `generateId()` returned, exactly as before. So yes, a pre-existing directory is how a reused id gets discovered late — that is the `PathController.mkdirs()` interaction above. Ordering is byte-identical to the pre-PR code; the PR neither improves nor worsens it, and the new `writeNewInfoRejectsAnIdWhoseDirectoryExistsWithoutInfo` test pins it.

## 4. The `listInfo` skip is byte-for-byte the old behaviour in both services

Diffed both old bodies against the helper:

- **`KonstructorImpl.list`** — `!isDirectory -> null`, `!infoFile.exists() -> null`, `decodeFromStream<Space>`. Identical, no post-mapping.
- **`WorkspaceImpl.list`** — same two skips, `decodeFromStream<KonstructionInfo>(input).konstruction`. The PR moves `.konstruction` out of the `mapNotNull` into a trailing `.map`. `KonstructionInfo.konstruction` is declared `val konstruction: Konstruction` (non-nullable, `KonstructorTypes.kt:54`), so the old `mapNotNull` could never have dropped anything there — `mapNotNull` and `map` are equivalent on this lambda. No widening.

Nothing *new* depends on the skip; it is load-bearing pre-existing behaviour (see the `PathController` note) that this PR is the first to actually test. `listFiles()` returning null on an unreadable/absent parent -> `emptyList()` is also unchanged.

## 5. Should they be allowed to diverge? Not on these three, and no silent choice was made

The one place the two bodies genuinely differed was instrumentation: `KonstructorImpl.generateId` was wrapped in `callContext("generateId")` and `WorkspaceImpl.generateId` was not. That asymmetry is **preserved** (`KonstructorImpl` still wraps the `firstFreeId` call; `WorkspaceImpl` does not), so the fold did not quietly pick a winner. The other divergences — base dir, decoded type, `KonstructionInfo` wrapper, the workspace-id guard — all stay at the call sites. That is the right seam: what is shared is the on-disk contract, and it should not be allowed to drift; what differs is per-service and stays per-service.

## 6. Existing on-disk state: unaffected

Verified rather than assumed. Same paths (`File(File(dir, id), "info.json")`, with the literal replaced by `INFO_JSON = "info.json"` — same value), same `config.json` instance, same reified serializers (`Space` and `KonstructionInfo`, inferred identically at both call sites), same guards in the same order. Nothing migrates, renames, rewrites or deletes anything on disk, and no read path becomes stricter or looser. A data dir written by the current production build is read identically by this build and vice versa. Also confirmed `WorkspaceImpl.getName`/`setName` still import and use `decodeFromStream`/`encodeToStream` — the import removal was `KonstructorImpl`-only, and the tree compiles.

---

## Verification I ran myself

`spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true`, old XML deleted with the find form first, `:backend:jvmTest` **executed** (not `UP-TO-DATE`), counted from freshly written XML:

**143 tests, 0 failures, 0 errors, 6 skipped** across 29 result files (126 backend + 17 protocol) — matching the PR body exactly. `SubprocessFailureTest` (32.4s) and `CompileAndExecuteIntegrationTest` (6.7s) both ran, so the integration gate was live. `InfoStoreTest` = 6 tests.

**Mutation control — re-run, not trusted.** Applied both mutations to `InfoStore.kt` (`var id = 0` -> `1`; delete the `if (!infoFile.exists()) return@mapNotNull null` line), `2 of 6` failed:

```
InfoStoreTest > firstFreeIdFillsTheLowestGap FAILED
    org.junit.ComparisonFailure: expected:<[0]> but was:<[1]>
        at com.monkopedia.konstructor.InfoStoreTest.firstFreeIdFillsTheLowestGap(InfoStoreTest.kt:79)

InfoStoreTest > listInfoSkipsLooseFilesAndDirectoriesWithoutInfo FAILED
    java.io.FileNotFoundException: /tmp/info-store-test-3558185857062761689/half-created/info.json (No such file or directory)
        at com.monkopedia.konstructor.InfoStoreTest.listInfoSkipsLooseFilesAndDirectoriesWithoutInfo(InfoStoreTest.kt:150)
```

Confirmed as reported. The tests bite.

## The disclosed flake — it is real, it is not the timing test, and I have the culprit

CI attempt 1 on this head ([32493613297](https://github.com/Monkopedia/Konstructor/actions/runs/32493613297)) went **red** on exactly this flake — `126 tests completed, 1 failed`, `ScopedShipperTest.requestPickupDelegatesToWrappedShipper` / `UncaughtExceptionsBeforeTest` — on the tree **with the timing test already dropped**. So the "2/9 with it vs 0/6 without it" split should not be read as implicating the timing test; the PR body was right to hedge, and the hedge was the correct call.

The `junit-xml` artifact carries the leaked exception. It is **not** the `GlobalScope` log plumbing — it is `ExecUtil.ExecProcess`:

```
Suppressed: java.io.IOException: Stream closed
	at kotlin.io.ByteStreamsKt.copyTo(IOStreams.kt:125)
	at com.monkopedia.konstructor.tasks.ExecUtil$ExecProcess$1.invokeSuspend(ExecUtil.kt:82)
	Suppressed: DiagnosticCoroutineContextException: [... StandaloneCoroutine{Cancelled} ..., Dispatchers.IO]
```

One coroutine pumps `proc.errorStream.copyTo(System.err)` catching only `CancellationException`; a sibling closes `proc.errorStream` after `waitFor()`. The blocked read wakes with `IOException: Stream closed`, the catch misses it, and with only a bare `SupervisorJob` and no handler it escapes to the default uncaught handler and lands on the next `runTest`. `ExecUtil.kt` was last touched 2026-05-21 and is untouched by this PR; it only fires for the `-Dintegration=true` subprocess tests. Full write-up and suggested fix on #101.

**My own flake sampling on this branch:** 0 failures in **7** full local runs (`--rerun`, fresh XML each, 143 tests each) vs **1 failure in 1** CI run. 0/7 does not establish stability — it is consistent with a low-rate machine-timing race that a 2-core runner hits and this workstation mostly does not. Any single green CI run on this repo is one sample, not a stable gate.

---

## Non-blocking note

`writeNewInfo`'s KDoc says "**Claim** [targetInfo] for a new item" and "an existing directory means the id is taken". Read on its own that sounds stronger than it is — a reader could reasonably take it as making `create()` collision-safe, which #102 shows it is not. Worth one sentence saying neither helper is atomic with respect to concurrent callers, pointing at #102. Not worth blocking a behaviour-preserving refactor over.

Approving on the code. Merging only once `gh pr checks` is a completed SUCCESS.

Issue #101, opened 2026-08-21
ExecUtil's stderr pump leaks an IOException: the catch is narrower than the ways the read can fail


monkopedia-coder:
## What happens

`ScopedShipperTest.requestPickupDelegatesToWrappedShipper` intermittently fails the full `:backend:jvmTest` suite with:

```
kotlinx.coroutines.test.UncaughtExceptionsBeforeTest: There were uncaught exceptions before the
test started. Please avoid this, as such exceptions are also reported in a platform-dependent
manner so that they are not lost.
	at kotlinx.coroutines.test.TestScopeImpl.enter(TestScope.kt:231)
	at kotlinx.coroutines.test.TestBuildersKt__TestBuildersKt.runTest-8Mi8wO0(TestBuilders.kt:308)
	...
	at com.monkopedia.konstructor.logging.ScopedShipperTest.requestPickupDelegatesToWrappedShipper
```

**The test named in the failure is not the test at fault.** `UncaughtExceptionsBeforeTest` means an exception escaped to the global handler *before* this test started, and `kotlinx-coroutines-test` attributes it to whichever `runTest` runs next. `ScopedShipperTest` itself uses no background coroutines — it builds a fake `Shipper`, calls `requestPickup()` and asserts on a `DropBox`. It is simply the next `runTest` in line.

So the real defect is that **some earlier test leaks an uncaught exception from a coroutine that outlives it** — the obvious suspects being the code paths that `GlobalScope.launch` (e.g. `KonstructorImpl`'s warehouse/log plumbing in `init`, and the controller concurrency tests), none of which is scoped to the test that starts them.

## It reproduces, and it is order/timing sensitive

Measured on this machine while working on #97, all full `:backend:jvmTest -Dintegration=true` runs with results counted from freshly written XML:

| tree | failures / runs |
| --- | --- |
| #97 branch **with** an extra CPU-heavy timing test | **2 / 9** |
| #97 branch **without** it | 0 / 6 |
| #97 branch, production change only, no new tests | 0 / 6 |
| unrelated control branch off the same base | 0 / 6 |

Run in isolation (`--tests '*ScopedShipperTest*'`) it passes every time.

**Read this as evidence of order/timing sensitivity, not as a diagnosis.** 0/6 cannot rule out a ~20% rate (p ≈ 0.22 of seeing zero), so the table does not establish that any particular test causes it — only that adding load or shifting execution order is enough to surface it, and that the failure does not live in the test it is reported against.

[...]

Issue #84, opened 2026-08-07
LSP self-heal is blind to an expired engine, and respawns unbounded on a dead one


monkopedia-coder:
## Summary

The `reconnect()`/"self-heal" path added in #76 does not cover the failure mode that actually took prod down on 2026-07-28: an **expired** kotlin-lsp engine binary. And its retry has no ceiling, so a permanently-dead engine is re-spawned on every editor event.

## 1. "Degrade to inert" cannot detect an expired engine

`BridgeLanguageServer`'s KDoc states that if the subprocess is unavailable the bridge "degrades to an inert server." Unavailability is detected as exactly three things:

- env var unset,
- binary absent / not executable (`Config.kt` `isKotlinLspAvailable` = `exists() && canExecute()`),
- `ProcessBuilder.start()` **throwing** (`KotlinLspProcess.kt`).

An expired EAP build satisfies none of these. It is present, executable, and `start()` returns normally — the process launches, prints `This build of intellij-server has expired`, and exits. `connection` is set **non-null**, so the inert-degrade branch is never taken. This is precisely the 07-28 → 08-03 outage class, and engines expire on a ~40-day clock, so it is the *expected* failure mode, not an exotic one.

## 2. Retry is unbounded on a permanently-dead engine

The same KDoc justifies omitting backoff: *"recovery is naturally paced by request traffic."* But `textDocumentHover` is routed through `withLiveDelegate` (the reconnecting path), and its own comment notes the editor fires hover **on every cursor move**. On a dead-but-started engine, each hover → `reconnect` → `ensureConnection` sees `process.isAlive == false` → `shutdownLocked()` → **spawns a fresh `intellij-server` JVM**. There is no ceiling; "paced by request traffic" is the user's cursor-move rate.

## Why it matters

Institutional-memory trap: at the next expiry, someone greps and finds "self-heal shipped 2026-07-28, verified green" and rules out the bridge — the exact wrong place to stop. The #76 tests exercise a `FakeEngine` respawn that succeeds; the real `KotlinLspProcess.ensureConnection` failure is `assumeTrue`-gated off in CI.

## Suggested direction

Treat "process started but exited / never completed `initialize`" as unavailable (readiness handshake or a short liveness probe before marking `connection` live), and bound respawns per unit time. Related: #83 (connection-scope) already reworked the surrounding lifecycle; #35 (LSP epic). Found in the 2026-08-07 audit.

monkopedia-coder:
[...]

PR #103, opened 2026-08-21
feat(frontend): default lspEnabled to ON (#35)

monkopedia-coder:
Refs #35 — the LSP epic stays open; this lands one decision from it.

## The ruling

> *"sure, flip it on, we have basically no users, so no real harm"* — 2026-08-21

This settles go-live decision 2 from 2026-06-04, which had never actually been ruled on: the feature shipped OFF because that is how the flag was built, not because anyone chose it.

**The reasoning is the ruling, not a justification attached to it.** He judged the **blast radius**, not the feature. What was unmeasured when he answered, and still is: the PoC found accurate kcsg diagnostics with zero false positives, but nobody has a month of real editing behind it, and the cost figures are PoC numbers (~2GB heap, ~120s cold index, ~4s warm init, ~8s didOpen→diagnostics, 0.2–2.2s completion).

So the condition is his own: **if konstructor acquires real users, the basis for this decision is gone and it gets re-asked** — not silently carried forward. That is written into the comment above the flag, not just here, because the person who finds this in six months will otherwise see "owner approved default-on" without the premise it rested on. One flag, one feature; this is not a licence to flip other defaults.

## What changes, and what does not

`SettingsViewModel.kt`: `PersistedStateFlow.boolean("lspEnabled", false)` → `true`.

**That is the DEFAULT only, and the default applies only where nothing is stored.** `lspEnabled` is a persisted per-user setting:

| who | before | after |
| --- | --- | --- |
| explicitly turned LSP **off** | off | **off** — unchanged |
| explicitly turned LSP **on** | on | **on** — unchanged |
| never touched it / fresh browser | off | **on** |

If the intent were "everyone gets LSP now", a default change does not deliver that — something would have to clear stored values, which is more invasive and **not authorised**. The intent taken is "new users get it", for which this is exactly right and nothing more is needed. Worth stating plainly rather than leaving as "probably academic", because that is the reasoning that stops people checking.

`Config.isKotlinLspAvailable` remains the second gate: with no `kotlin-lsp` binary configured or present, the backend never spawns the subprocess and the editor behaves exactly as it does without LSP. CI has no binary, so on the runner the flag being `true` changes nothing observable — which is precisely why the e2e coverage asserts the flag rather than an editor behaviour.

## Why a test is edited alongside a behaviour change

`ThemeAndKeymapTest.testLspEnabledToggle` **asserted the old default, and the default is the thing being changed.** That is not a test bent to fit the code — a reviewer seeing a test move next to a behaviour change should ask, and the answer is here. It now also toggles off and back on, so the stored value it leaves behind matches the new default and it does not change what a later test sees.

**RED control, because "the test passes" is not evidence that it measures the default.** With the new test in place and the production default reverted to `false`, the full wasm bundle and shadow JAR rebuilt:

```
com.monkopedia.konstructor.e2e.ThemeAndKeymapTest  tests 1  failures 1
FAILED testLspEnabledToggle | java.lang.AssertionError: expected:<true> but was:<false>
```

The default was then restored to `true` and the restore verified in the source before committing.

## Verification

Local, fleet build slot, old result XML deleted with the find form first:

- `./gradlew spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true shadowJar` — **backend 132 tests / 0 failures / 6 skipped** (27 XML), **protocol 17 / 0 / 0** (2 XML).
- `./gradlew :e2e:test -Pe2e --tests "*ThemeAndKeymapTest*"` under Xvfb with the Mesa EGL override — **4 tests / 0 failures / 0 skipped**; `testLspEnabledToggle` ran for 13.7s. Run locally on purpose: the assertion being changed is an e2e one, and it should not be CI that discovers a problem with it.

## Notes for the reviewer

- `main` is unprotected here, so `--auto` is not a CI gate — please poll `gh pr checks` to a completed SUCCESS yourself, then plain `--squash --delete-branch`.
- Known-flaky gate on the way: `ExecUtil`'s stderr pump leaks an `IOException` that lands on an unrelated `runTest` under `-Dintegration=true` (#101). If CI goes red on `ScopedShipperTest`, that is it, and it is not this change. Please confirm the failure signature against #101 rather than assuming.
- Deployment is out of scope. :8001 is manual and the owner's domain; nothing here touches it.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -66,7 +66,7 @@ The backend's `shadowJar` task orchestrates the full build:
 - Koin for dependency injection with ViewModels (`koin-compose-viewmodel`).
 - State management via Kotlin StateFlows, collected in Composables via `collectAsState()`.
 - JsBridge (formerly TestBridge) pattern for Playwright e2e testing: exposes app state to `globalThis.__konstructor` since Compose renders to a WebGL canvas (no DOM elements for Playwright selectors).
-- **LSP**: a kcsg-aware language server (completion/hover/pull-diagnostics) is wired through `backend/src/jvmMain/.../lsp/` (a `BridgeLanguageServer` fronting a standalone kotlin-lsp/intellij-server engine spawned via `KONSTRUCTOR_KOTLIN_LSP`) and the frontend editor's LSP client. Opt-in via a Settings toggle.
+- **LSP**: a kcsg-aware language server (completion/hover/pull-diagnostics) is wired through `backend/src/jvmMain/.../lsp/` (a `BridgeLanguageServer` fronting a standalone kotlin-lsp/intellij-server engine spawned via `KONSTRUCTOR_KOTLIN_LSP`) and the frontend editor's LSP client. On by default since 2026-08-21; opt out via a Settings toggle.
 
 ## Frontend Structure (src/wasmJsMain/)
 
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionServiceImpl.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionServiceImpl.kt
@@ -119,7 +119,8 @@ class KonstructionServiceImpl(
             // param, so its publishDiagnostics channel rides the same WebSocket) is
             // stashed and the bridge forwards real kcsg-aware diagnostics to it.
             //
-            // Still wholly flag-gated on the frontend (lspEnabled, default OFF) and the
+            // Still wholly flag-gated on the frontend (lspEnabled, default ON since
+            // 2026-08-21 — the flag still exists, only its default moved) and the
             // bridge degrades to an inert server when no kotlin-lsp binary is configured
             // (CI), so the absence of the engine never crashes konstructor.
             BridgeLanguageServer(config, workspaceId, id, client)
--- a/e2e/src/test/kotlin/com/monkopedia/konstructor/e2e/LspPipeTest.kt
+++ b/e2e/src/test/kotlin/com/monkopedia/konstructor/e2e/LspPipeTest.kt
@@ -21,6 +21,7 @@ import com.monkopedia.ksrpc.ktor.websocket.asWebsocketConnection
 import com.monkopedia.ksrpc.toStub
 import io.ktor.client.HttpClient
 import io.ktor.client.plugins.websocket.WebSockets
+import kotlin.test.assertEquals
 import kotlinx.coroutines.runBlocking
 import org.junit.Test
 
@@ -38,9 +39,12 @@ import org.junit.Test
  *
  * This test therefore asserts the flag-ON path is wired and harmless WITHOUT a
  * binary: the editor mounts, the session establishes, the app stays on the main
- * screen, and no spurious diagnostics appear. The flag-OFF behavior (editor
- * byte-for-byte unchanged) is covered by the rest of the suite running with the
- * default-off flag.
+ * screen, and no spurious diagnostics appear.
+ *
+ * Flag-OFF used to be covered implicitly, by the whole suite running under the
+ * default-off flag. **That stopped being true when the default flipped to ON**
+ * (2026-08-21), so [testEditorWorksWithLspFlagOff] now covers it explicitly rather
+ * than the coverage disappearing silently along with the default.
  */
 class LspPipeTest : BaseE2eTest() {
 
@@ -116,6 +120,60 @@ class LspPipeTest : BaseE2eTest() {
         assert(count == 0) {
             "With no kotlin-lsp binary on CI the bridge is inert; expected 0 " +
                 "diagnostics, got count=$count"
+        }
+    }
+
+    @Test
+    fun testEditorWorksWithLspFlagOff() {
+        loadApp()
+        waitForBridge()
+
+        bridgeAction("createWorkspace", "LspOffWs")
+        waitForState(
+            "globalThis.__konstructor.state.workspaceIds && " +
+                "globalThis.__konstructor.state.workspaceIds.length >= 1"
+        )
+        val wsId = bridgeStateStringList("workspaceIds").first()
+
+        val konId = runBlocking {
+            val env = ksrpcEnvironment { }
+            val client = HttpClient { install(WebSockets) }
+            val conn = client.asWebsocketConnection("${server.baseUrl}/konstructor", env)
+            val service = conn.defaultChannel().toStub<Konstructor, String>()
+            val workspace = service.get(wsId)
+            val kon = workspace.create(
+                Konstruction(name = "LspOffTest", workspaceId = wsId, id = "")
+            )
+            service.konstruction(kon).set(validScript)
+            kon.id
+        }
+
+        page.reload()
+        waitForBridge()
+        waitForMainScreen()
+
+        // Explicitly OFF. Since 2026-08-21 the default is ON, so this is the only place
+        // the flag-off editor is exercised — it is not the ambient state of the suite
+        // any more.
+        bridgeAction("setLspEnabled", "false")
+        assertEquals(false, bridgeStateBoolean("lspEnabled"))
+
+        bridgeAction("setCodePaneMode", "EDITOR")
+        bridgeActionNoWait("selectKonstruction", konId)
+        page.waitForTimeout(5000.0)
+
+        // With the flag off the editor takes the pre-LSP path: no lsp() call, no
+        // languageServerSupport, and therefore nothing can publish diagnostics.
+        waitForMainScreen()
+        assertEquals(false, bridgeStateBoolean("lspEnabled"))
+        val count = (
+            page.evaluate(
+                "() => (globalThis.__konstructor.lspDiagnostics && " +
+                    "globalThis.__konstructor.lspDiagnostics.count) || 0"
+            ) as? Number
+            )?.toInt() ?: 0
+        assert(count == 0) {
+            "With lspEnabled=false nothing should publish diagnostics, got count=$count"
         }
     }
 }
--- a/e2e/src/test/kotlin/com/monkopedia/konstructor/e2e/ThemeAndKeymapTest.kt
+++ b/e2e/src/test/kotlin/com/monkopedia/konstructor/e2e/ThemeAndKeymapTest.kt
@@ -131,19 +131,24 @@ class ThemeAndKeymapTest : BaseE2eTest() {
     fun testLspEnabledToggle() {
         setupWorkspaceWithCode()
 
-        // Default is opt-out: LSP off, so the editor is unchanged for users who
-        // never enable it.
-        assertEquals(false, bridgeStateBoolean("lspEnabled"))
-
-        // The Settings switch (and its bridge action) flips the persisted flag;
-        // the snapshot reflects it so the toggle shows the right state.
-        bridgeAction("setLspEnabled", "true")
-        page.waitForTimeout(3000.0)
+        // Default is opt-OUT as of 2026-08-21: a browser with nothing stored gets
+        // LSP on. Config.isKotlinLspAvailable is still the second gate, so on a
+        // runner with no kotlin-lsp binary the flag being true changes nothing the
+        // user can see — which is exactly why this asserts the flag and not an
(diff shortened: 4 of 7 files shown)


monkopedia-reviewer:
## Independent review — verified, with one question that is the owner's to answer

I did not write this PR and took nothing in the body as evidence. Everything below I confirmed myself against `632800d` in a detached worktree. **The implementation is correct and I found no defect in it.** I am not merging yet for one reason, in §3.

---

### 1. Is it really default-only? ✅ Confirmed

`PersistedStateFlow.create` (`frontend/.../viewmodel/PersistedStateFlow.kt`):

```kotlin
val initial = try {
    storage.getItem(fullKey)?.let(decode) ?: default
} catch (_: Throwable) { default }
```

The default is reached **only** when `getItem` returns null, i.e. nothing is stored. `boolean()` decodes with `toBooleanStrictOrNull()`, so a stored `"false"` decodes to `false` and a stored `"true"` to `true` — **both directions of an existing stored value survive.** The diff touches two files and neither writes, migrates, or clears storage. Repo-wide (`/usr/bin/grep -r`, not the shell's ugrep) the only code that removes `konstructor.*` keys is the pre-existing `Ctrl+Shift+Alt+R` escape hatch in `frontend/src/wasmJsMain/resources/index.html:92` — user-initiated, untouched here. The table in the description is accurate.

One edge, non-blocking: a *corrupt* stored value (anything other than `true`/`false`) falls back to the default **in memory only**, storage untouched — so such a user would now see ON where they saw OFF. Only this code writes the key, so it is effectively unreachable.

### 2. Is the second gate intact? ✅ Confirmed — and "inert" really means inert

`KotlinLspProcess.connect` opens with `if (!config.isKotlinLspAvailable) return null`; `BridgeLanguageServer.initialize` then leaves `connection = null`, logs once, and returns empty `ServerCapabilities`. From there **every** editor-facing method short-circuits on a null connection — `textDocumentDidOpen:381`, `textDocumentDidChange:412`, `textDocumentCompletion:622`, `textDocumentHover:658`, `textDocumentSignatureHelp`. Critically `withLiveDelegate` starts `val dead = connection ?: return fallback`, so with no engine there is **no reconnect attempt, no retry loop, no per-call logging**. `PullDiagnosticsPublisher` is never constructed (no `diagnosticProvider` in the empty capabilities). No error reaches the user.

Not *literally* zero, and worth stating precisely rather than as "changes nothing": per editor session it still opens the nested ksrpc `lsp()` sub-service, runs `lspWorkspace.synthesize()` (writes the wrapped `.kt` + `workspace.json` under the data dir), installs `languageServerSupport`, and emits exactly **one** log line — `kotlin-lsp engine unavailable; bridge is inert`. didOpen/didChange frames travel the socket and are dropped server-side. Nothing user-visible, no subprocess. Empirically confirmed: the **full** e2e suite on a runner with no binary is green (numbers below).

### 3. What a real user with an engine gets on first load — 🟡 owner call

Plainly: **yes, a fresh page load now starts an LSP session that did not start before.** `EditorPane.kt:212` keys `produceState` on `lspEnabled`, so with nothing stored the session builds as soon as a konstruction's editor pane mounts — `service.lsp()` → `initialize` → `KotlinLspProcess.connect` → **spawn the warm `intellij-server` subprocess** (PoC figures: ~2GB heap, ~120s cold index, ~4s warm init, ~8s didOpen→diagnostics) → handshake → pull-diagnostics publisher. On CI that is all no-ops. On :8001, which has an engine configured, it is the real thing. That much is simply "the feature is now on", which is what was ruled.

The part I do not think I should decide is this: **open issue #84** ("LSP self-heal is blind to an expired engine, and respawns unbounded on a dead one", filed 2026-08-07, still open, unfixed at `632800d`) documents that an **expired** engine binary is *not* caught by the inert path — it exists, it is executable, `start()` returns normally, so `connection` is set non-null. Then `textDocumentHover` — which the editor fires on **every cursor move** — goes through `withLiveDelegate` → `reconnect` → `ensureConnection` sees a dead process → spawns a fresh `intellij-server` JVM, **with no ceiling**. #84 states this is the failure that took prod down 2026-07-28 → 08-03, and that engines expire on a ~40-day clock, so it is the expected failure mode rather than an exotic one.

Before this PR, reaching that required a user to tick the Settings box. After it, it is the default for every fresh browser. The ruling — *"we have basically no users, so no real harm"* — is a judgement about **user-visible** blast radius; #84's blast radius is the **server**. That may well still be fine (the toggle is right there, and #84 predates this change), but it is a premise the ruling did not cover, so I am asking rather than assuming.

**@Monkopedia — one question: land this now, or land #84 first?** Say "land it" and I will merge immediately; everything else is verified and CI is green. Nothing about the code needs to change either way.

### 4. The e2e test edit — ✅ honest, and I reproduced the RED control myself

The test asserted the old default and the default is what changed; that is legitimate, and the edit does not weaken it — it goes from 2 assertions to 3, still covering both toggle directions. I verified it actually measures the default rather than merely passing: production default reverted to `false`, wasm bundle + shadow JAR rebuilt, new test kept:

```
ThemeAndKeymapTest > testLspEnabledToggle FAILED
    java.lang.AssertionError at ThemeAndKeymapTest.kt:139
4 tests completed, 1 failed

<failure message="java.lang.AssertionError: expected:&lt;true&gt; but was:&lt;false&gt;" ...>
```

The other 3 tests in the class passed under the reverted default, so the assertion is isolated to the default. Restored to `true` afterwards and verified: `git status --porcelain` and `git diff --stat` both empty at `632800d`.

### 5. Test isolation — ✅ not an issue (the toggle-back is belt-and-braces)

e2e tests do **not** share browser storage, for three independent reasons: `BaseE2eTest.setUpBase` calls `browser.newPage()`, which creates a fresh `BrowserContext` with its own localStorage; `playwright`/`browser` are instance fields and JUnit 4 constructs a new test instance per method, so `playwright == null` is true every time and each test gets its own Browser; and `ServerFixture` binds `ServerSocket(0)`, so every test is a different `localhost:<port>` **origin** and localStorage is origin-keyed. The toggle-back is harmless and I would keep it, but nothing depended on it.

### 6. Other places asserting the old default — 4 stale statements left behind

Found with `git grep` / `/usr/bin/grep -r`. None change behaviour, but the description argues this change should be legible to whoever finds it in six months, and these now say the opposite:

1. `backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionServiceImpl.kt:122` — *"Still wholly flag-gated on the frontend (lspEnabled, default OFF)"*.
2. `e2e/src/test/kotlin/com/monkopedia/konstructor/e2e/LspPipeTest.kt:43` — *"The flag-OFF behavior (editor byte-for-byte unchanged) is covered by the rest of the suite running with the default-off flag."* This one is more than a stale comment: it describes coverage that **this PR silently removes**. With the default flipped, all 43 other e2e tests now run with the LSP session established, and **nothing exercises the flag-OFF editor path any more**. Worth a deliberate decision (one test that explicitly sets it off), not a comment left describing a world that no longer exists.
3. `frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/ui/settings/SettingsPane.kt:97` — *"Opt-in; requires the server to have an LSP engine configured."*
4. `CLAUDE.md:69` — *"Opt-in via a Settings toggle."*

Also `BridgeStateSnapshot.kt:50` still has `val lspEnabled: Boolean = false`. Harmless — `JsBridge.kt:427` always passes the value explicitly — but it mirrors the old default and may as well match.

---

### Verification I ran (fleet build slot, old XML deleted with the find form first, fresh worktree)

| gate | result |
| --- | --- |
| `spotlessCheck` | pass |
| `:backend:jvmTest -Dintegration=true` | **132 tests / 0 failures / 0 errors / 6 skipped** (27 fresh XML) |
| `:protocol:jvmTest -Dintegration=true` | **17 / 0 / 0 / 0** (2 fresh XML) |
| `shadowJar` | pass |
| `:e2e:test -Pe2e` — **full suite**, Xvfb + Mesa EGL override | **44 tests / 0 failures / 0 errors / 9 skipped** (19 fresh XML) |
| RED control (default reverted to `false`) | `testLspEnabledToggle` fails `expected:<true> but was:<false>` ✅ |
| CI `build-and-test` @ `632800d` | **pass, 11m57s** — polled to completion, no re-run needed |

I ran the **whole** e2e suite rather than just `ThemeAndKeymapTest`, precisely because flipping the default changes what all 43 other tests run under. It is green. The #101 `ExecUtil` flake did not fire in either my run or CI, so no re-run was needed and no red needed attributing.

*Reviewed independently; claim held pending the §3 answer.*

monkopedia-reviewer:
## Re-review of `044ef03` — approved, merging

I re-verified the new head myself rather than inheriting the numbers.

**The delta is what it claims to be.** `git diff 632800d 044ef03` is one new e2e test plus four comment/default corrections — no production behaviour beyond the one-line default that was already reviewed.

- `LspPipeTest.testEditorWorksWithLspFlagOff` — this is the finding I cared about, and it is properly fixed. Flag-OFF is now asserted explicitly (the flag stores `false`, the app stays on the main screen, nothing publishes diagnostics), and the class KDoc records *why* the test exists — that the coverage used to be ambient and stopped being so when the default flipped. That is the part that stops a future reader deleting it as redundant. Honest caveat, which the KDoc is already straight about: with no engine on CI, flag-ON and flag-OFF both yield zero diagnostics, so the load-bearing assertions here are the flag value and that the pre-LSP editor path (`if (!lspEnabled) return@produceState` — no `lsp()` call at all) still mounts cleanly. That is the right thing to assert on an engine-less runner.
- The four stale claims are corrected accurately in `KonstructionServiceImpl.kt`, `SettingsPane.kt`, and `CLAUDE.md`.
- `BridgeStateSnapshot.lspEnabled` `false` → `true`: correct, and worth one note for the record. There is a **second** construction site — `JsBridge.kt:458` in `refreshKonstructions` — that omits `lspEnabled` (along with `editorTheme`, `keymap`, and hardcoding `codePaneMode = "EDITOR"`) and so falls through to the data-class default unconditionally, not just before the settings flow emits. That path reported `false` regardless of the real setting before this change and reports `true` regardless after it: a pre-existing test-bridge wart, exactly as wrong in both directions, neither introduced nor worsened here. I checked it cannot destabilise the new test — `testEditorWorksWithLspFlagOff` creates its konstruction over the ksrpc API and reloads, so no `refreshKonstructions` runs after the toggle. Not worth a change in this PR; noting it so it is findable if the bridge ever reports a stale setting.

**My verification on `044ef03`** (fleet build slot, fresh detached worktree, old XML deleted with the find form first):

| gate | result |
| --- | --- |
| `spotlessCheck` | pass |
| `:backend:jvmTest -Dintegration=true` | **132 / 0 failures / 0 errors / 6 skipped** (27 fresh XML) |
| `:protocol:jvmTest -Dintegration=true` | **17 / 0 / 0 / 0** (2 fresh XML) |
| `shadowJar` | pass |
| `:e2e:test -Pe2e` — **full suite**, Xvfb + Mesa EGL | **45 / 0 failures / 0 errors / 9 skipped** (19 fresh XML) |
| CI `build-and-test` | **pass, 11m51s**, run `32504174815` on sha `044ef03` |

I re-ran the **full** e2e suite rather than just `LspPipeTest`, for the reason it mattered the first time: the default flip changes the ambient state every test runs under, and this head also changes a snapshot default that every test reads through the bridge. 44 → 45 tests, exactly the one added, everything else unchanged and green. `testEditorWorksWithLspFlagOff` ran in 7.809s. The #101 `ExecUtil` flake fired in neither my run nor CI, so nothing needed re-running or attributing.

**On the escalation.** I agree with the ruling and would have reached it myself given the deploy gate. My §3 concern was never that the code was wrong — it was that default-ON silently widens exposure to #84 from "someone who ticked a box" to "anyone's first hover". Merging does not create that exposure, because merging is not deploying and :8001 is manual and the owner's alone. What removes the risk is the gate, and I confirmed it is actually recorded rather than taking it on trust: #84 now carries the explicit "precondition for the next :8001 deploy" note and has been labelled `agent-workable`. A written gate on a queued bug is a real control; an indefinite hold on a correct, tested change is not. No objection.

#35 stays open — the body says `Refs`, not `Fixes`.

Monkopedia:
#84 first

PR #105, opened 2026-08-21
fix(tasks): stop ExecProcess leaking IOException from the stderr pump

### LINKED ISSUE #101: ExecUtil's stderr pump leaks an IOException: the catch is narrower than the ways the read can fail

monkopedia-coder:
Fixes #101

## The defect

`ExecProcess` closes the process streams once `waitFor()` returns, which wakes the stderr pump with `IOException: Stream closed`. The pump caught only `CancellationException`, so under a bare `SupervisorJob` with no handler the exception escaped to the JVM's default uncaught handler. In CI that surfaced as an **unrelated** test failing with `UncaughtExceptionsBeforeTest` and a stack trace pointing at `ExecUtil.kt`:

```
Suppressed: java.io.IOException: Stream closed
	at kotlin.io.ByteStreamsKt.copyTo(IOStreams.kt:125)
	at com.monkopedia.konstructor.tasks.ExecUtil$ExecProcess$1.invokeSuspend(ExecUtil.kt:82)
```

A catch clause narrower than the ways the operation can fail, with the punishment landing somewhere else.

## The fix — two layers, because the catch and the escape are separate defects

1. **The stderr pump and the connection setup also catch `IOException`.** A closed stream is the ordinary way a subprocess ends; catching only cancellation left a *normal outcome* escaping as an error.
2. **The scope carries a `CoroutineExceptionHandler`**, so nothing from this process's plumbing can reach the global handler again. Anything genuinely unexpected is logged where it names `ExecProcess`, instead of ambushing whichever test runs next.

Layer 1 without layer 2 would fix this instance and leave the next one to land on a stranger.

## Why it appeared now

It only fires under `-Dintegration=true`, which CI did not pass until `8c162ec` (#94, part of #82). **The leak is not new — its exposure is.** That is a gate doing its job, not a regression introduced by #94.

## Demonstrated RED

Reverting `ExecUtil.kt` to main with the test unchanged:

```
runningASubprocessToCompletionLeavesNoUncaughtException FAILED
java.lang.AssertionError: Running a subprocess to completion leaked 1 uncaught
exception(s): IOException: Stream closed
```

Restoring it goes green. `:backend:jvmTest` confirmed executed (not `UP-TO-DATE`) with result XML deleted first, and the reverted run additionally verified at the **class** level — the fix introduces a log string that exists nowhere in the original source, and it was confirmed absent from the compiled class during the red run, ruling out a stale binary.

## The test ships with its own guard, and that is not ceremony

**Three earlier versions of this test reported no failure and looked like evidence the fix was unnecessary.** They were not. A *deliberate* leak did not fail them either: the harness assumed `kotlinx-coroutines-test` charges a leaked exception to the next `runTest`, which it does in a full CI run but **not** when a single class is run under a `--tests` filter — which is how the proof was being executed. Every one of those results was void.

`uncaughtExceptionRecorderIsNotBlind` now asserts the recorder still catches a known leak, so the real test can only pass for the right reason. **A regression test that has silently gone blind is worse than no test, because it reports safety.**

Two dead ends are documented in the test body so they are not re-introduced as simplifications:
- A subprocess that writes and exits does **not** reproduce it — stderr hits EOF as the process dies, `copyTo` returns cleanly, and the later close has nothing to interrupt. The pump must be *blocked in a read*.
- An unthrottled writer (`yes >&2`) reproduces it but feeds Gradle's captured test output until the heap dies — it failed the full suite with `OutOfMemoryError`, which the test then dutifully reported as a "leak". **It passed under a `--tests` filter and would have hit CI.** The writer is now dripped: many small reads over time, same window, bounded memory.

## Verification

`./gradlew spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true` on the full suite — **backend 134 tests / 0 failures / 6 skipped** (28 XML), **protocol 17 / 0 / 0** (2 XML), spotless clean. Full suite deliberately, not the filtered class: the filtered run is exactly what hid the OOM.

## Notes for the reviewer

- Worth challenging: the new `IOException` catches are deliberately silent, on the grounds that a closed stream at process exit is expected and logging it would be noise on every single subprocess. If you think a genuine mid-stream I/O error should be distinguishable from that, say so — it is the one judgement call in this diff.

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/tasks/ExecUtil.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/tasks/ExecUtil.kt
@@ -130,6 +130,14 @@ object ExecUtil {
                     // built on are closed when the process exits. Leave `connection`
                     // uncompleted — callers already handle a process that died before it
                     // could be attached to.
+                    //
+                    // Logged, unlike the stderr pump, because the two failures are not
+                    // equally expected here. `connection` is awaited in one place
+                    // (ScriptManager, inside withTimeout), so a GENUINE I/O failure while
+                    // attaching would otherwise present only as a timeout with the cause
+                    // discarded and no trace anywhere. Debug level keeps the ordinary
+                    // shutdown case from being noise.
+                    hauler.debug("Process streams closed before the channel attached: ${t.message}")
                 }
             }
             parentScope.launch {
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/tasks/ExecUtil.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/tasks/ExecUtil.kt
@@ -18,6 +18,7 @@ package com.monkopedia.konstructor.tasks
 import com.monkopedia.hauler.CallSign
 import com.monkopedia.hauler.asAsync
 import com.monkopedia.hauler.debug
+import com.monkopedia.hauler.error
 import com.monkopedia.hauler.hauler
 import com.monkopedia.hauler.info
 import com.monkopedia.ksrpc.ErrorListener
@@ -25,10 +26,12 @@ import com.monkopedia.ksrpc.channels.Connection
 import com.monkopedia.ksrpc.ksrpcEnvironment
 import com.monkopedia.ksrpc.sockets.asConnection
 import java.io.BufferedReader
+import java.io.IOException
 import java.io.InputStreamReader
 import kotlin.coroutines.EmptyCoroutineContext
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.DelicateCoroutinesApi
 import kotlinx.coroutines.Dispatchers
@@ -67,8 +70,25 @@ object ExecUtil {
 
     class ExecProcess(private val proc: Process) {
         private val parentJob = SupervisorJob()
+
+        /**
+         * Nothing from this process's plumbing may reach the global handler.
+         *
+         * [parentJob] is a [SupervisorJob] with no handler of its own, so a child that
+         * throws does not cancel its siblings — it goes straight to the JVM's default
+         * uncaught-exception handler instead, where nothing connects it back to this
+         * class. In CI that surfaced as an unrelated test failing with
+         * `UncaughtExceptionsBeforeTest` and a stack trace pointing at this file
+         * (konstructor#101). Anything unexpected is logged here instead, naming its
+         * actual source.
+         */
+        private val exceptionHandler = CoroutineExceptionHandler { _, thrown ->
+            if (thrown !is CancellationException) {
+                hauler.error("Uncaught failure in subprocess plumbing", thrown)
+            }
+        }
         val parentScope = CoroutineScope(
-            parentJob + Dispatchers.IO + (
+            parentJob + Dispatchers.IO + exceptionHandler + (
                 CallSign.threadLoggingName?.let(::CallSign)
                     ?: EmptyCoroutineContext
                 )
@@ -82,6 +102,12 @@ object ExecUtil {
                     proc.errorStream.copyTo(System.err)
                 } catch (t: CancellationException) {
                     // That's fine.
+                } catch (t: IOException) {
+                    // Also fine, and the normal way this pump ends: the waitFor coroutine
+                    // below closes the process streams once it exits, and a read that was
+                    // blocked at that moment wakes with "Stream closed". The process ending
+                    // is not an error, so catching only CancellationException here left an
+                    // ordinary outcome escaping as an uncaught exception (konstructor#101).
                 }
             }
             parentScope.launch {
@@ -99,6 +125,11 @@ object ExecUtil {
                     )
                 } catch (t: CancellationException) {
                     // That's fine.
+                } catch (t: IOException) {
+                    // Same shutdown race as the stderr pump above: the streams this is
+                    // built on are closed when the process exits. Leave `connection`
+                    // uncompleted — callers already handle a process that died before it
+                    // could be attached to.
                 }
             }
             parentScope.launch {
--- a/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/tasks/ExecProcessLeakTest.kt
+++ b/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/tasks/ExecProcessLeakTest.kt
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2022 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.monkopedia.konstructor.tasks
+
+import java.io.IOException
+import java.util.concurrent.CopyOnWriteArrayList
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeout
+
+/**
+ * Regression test for konstructor#101: running a subprocess to completion must leave no
+ * uncaught exception behind.
+ *
+ * [ExecUtil.ExecProcess] closes the process streams once `waitFor()` returns, which wakes
+ * the stderr pump with `IOException: Stream closed`. The pump caught only
+ * `CancellationException`, so under a bare `SupervisorJob` with no handler the exception
+ * escaped to the process-wide uncaught handler — and in a test JVM that surfaced as an
+ * unrelated test failing later with a stack trace pointing at `ExecUtil.kt`.
+ *
+ * Needs only `bash`, so unlike the rest of this package it runs without `-Dintegration`.
+ */
+class ExecProcessLeakTest {
+
+    /**
+     * Guards the guard, and it is not ceremony — it was earned.
+     *
+     * The first version of [runningASubprocessToCompletionLeavesNoUncaughtException] was
+     * built on the assumption that `kotlinx-coroutines-test` charges a leaked exception to
+     * the next `runTest`. It does so in a full CI run, but NOT when a single class is run
+     * under a `--tests` filter — and that is how the proof was being executed. Three
+     * separate reproduction attempts reported no failure, which looked like evidence the
+     * fix was unnecessary. It was not: a **deliberate** leak did not fail that `runTest`
+     * either, so the harness had never been able to see anything and all three results
+     * were void. Recording the JVM's default uncaught handler works in both cases.
+     *
+     * A regression test that has silently gone blind is worse than no test, because it
+     * reports safety. This asserts the recorder still catches a known leak, so the test
+     * below can only pass for the right reason.
+     */
+    @Test
+    fun uncaughtExceptionRecorderIsNotBlind() = runBlocking {
+        val seen = recordUncaught {
+            CoroutineScope(SupervisorJob() + Dispatchers.IO).launch {
+                throw IOException("deliberate positive-control leak")
+            }
+            delay(SETTLE_MS)
+        }
+        assertTrue(
+            seen.isNotEmpty(),
+            "The uncaught-exception recorder saw nothing when handed a deliberate " +
+                "SupervisorJob leak. It is blind, so the regression test below cannot " +
+                "detect the ExecProcess leak either and its result means nothing."
+        )
+    }
+
+    @Test
+    fun runningASubprocessToCompletionLeavesNoUncaughtException() = runBlocking {
+        val seen = recordUncaught {
+            // The pump has to be BLOCKED IN A READ when the streams close — that is the
+            // only state producing "Stream closed" rather than a clean EOF. A command that
+            // writes and exits will not do it: stderr hits EOF as the process dies and
+            // `copyTo` returns normally. Backgrounding a writer that inherits stderr holds
+            // the pipe open past the direct child's exit, and making it CONTINUOUS keeps
+            // the pump looping so the close lands between two reads. That is also the
+            // production shape — see `ExecUtil.kill()`, where `bash -c "kotlin ..."` leaves
+            // a grandchild JVM holding the pipe.
+            //
+            // The writer is DRIPPED, not flooded. An unthrottled writer (`yes >&2`) also
+            // reproduces the leak, but `copyTo(System.err)` feeds Gradle's captured test
+            // output, so it exhausted the heap and failed the whole suite with an
+            // OutOfMemoryError — which this test then dutifully reported as a "leak". Many
+            // small reads spread over time give the same window with bounded memory.
+            val process = ExecUtil.executeWithChannel(
+                "(for i in \$(seq 1 400); do echo probe-\$i >&2; sleep 0.01; done) & " +
+                    "sleep 0.4; exit 0"
+            )
+            val exit = withTimeout(30_000) { process.exitCode.await() }
+            assertEquals(0, exit, "probe subprocess should exit cleanly")
+            delay(SETTLE_MS)
+        }
+        assertTrue(
+            seen.isEmpty(),
+            "Running a subprocess to completion leaked ${seen.size} uncaught " +
+                "exception(s): " + seen.joinToString { "${it::class.simpleName}: ${it.message}" }
+        )
+    }
+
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (changes requested):
## Independent review — the fix is right, the test's positive control is not

I reproduced everything from scratch in a detached worktree at `eaf1afc`. **The production fix in `ExecUtil.kt` is correct and I verified it end to end.** One blocking finding, in the new test rather than the fix, plus one smaller point on the judgement call you flagged.

### 1. RED reproduces, and the binary was genuinely rebuilt

`git checkout origin/main -- ExecUtil.kt`, test unchanged, `:backend:jvmTest --tests '*ExecProcessLeakTest*'`:

```
java.lang.AssertionError: Running a subprocess to completion leaked 1 uncaught
exception(s): IOException: Stream closed
  at ExecProcessLeakTest.kt:78
```

Stale-binary excluded at the class level, not the source level. Test XML deleted (find-form) and `ExecUtil*.class` deleted before every run:

| case | `:backend:compileKotlinJvm` | `strings ExecUtil*.class \| grep -c "Uncaught failure in subprocess plumbing"` | result |
|---|---|---|---|
| ExecUtil.kt reverted | executed (53 of 53 tasks executed) | **0** | 1 failure — the message above |
| ExecUtil.kt restored | executed | **1** | 2 tests, 0 failures |

The `red` build recompiled everything (`53 actionable tasks: 53 executed`) and produced a class file with **no** `…$special$$inlined$CoroutineExceptionHandler$1` and marker count 0; the `green` build produced the handler class and marker count 1. The two runs were demonstrably different binaries.

### 2. The self-guard bites

Broke it deliberately — `Thread.setDefaultUncaughtExceptionHandler { _, _ -> }` (records nothing), fix left in place, marker count still 1:

```
ExecProcessLeakTest[jvm] > uncaughtExceptionRecorderIsNotBlind[jvm] FAILED
    java.lang.AssertionError at ExecProcessLeakTest.kt:62
```

The guard is real. Good.

### 3. 🔴 BLOCKING — `uncaughtExceptionRecorderIsNotBlind` re-creates the exact defect this PR fixes

The positive control deliberately leaks an `IOException` into a bare `SupervisorJob` **in the shared test JVM**. `kotlinx-coroutines-test` 1.11.0 registers `ExceptionCollectorAsService` as a process-wide `CoroutineExceptionHandler` (`META-INF/services/kotlinx.coroutines.CoroutineExceptionHandler`). Reading the bytecode of `ExceptionCollector.handleException`:

- `enabled == false` → returns `false`, falls through to the thread's handler. Nothing stashed.
- `enabled == true` **and no `runTest` currently active** → **`unprocessedExceptions.add(exception)`**, returns `false`, so it *also* falls through to the thread handler.
- `enabled` is set to `true` by `addOnExceptionCallback` and **is never reset** — one `runTest` anywhere earlier in the fork latches it for the life of the JVM.

`TestScopeImpl.enter()` then calls `registerExceptionCollectorLocked()`, which drains `unprocessedExceptions` into the new scope and throws `UncaughtExceptionsBeforeTest`. So the deliberate leak is charged to **the next `runTest` in the same JVM** — the identical mechanism as #101.

Measured, not theorised. Probe class (`@FixMethodOrder(NAME_ASCENDING)`, three methods: `runTest{}` to latch `enabled`, then a byte-for-byte copy of your positive control, then `runTest{}`):

```
<testsuite name="ZzLeakOrderProbeTest[jvm]" tests="3" failures="1">
  <testcase name="cVictimRunTest[jvm]">
    <failure type="kotlinx.coroutines.test.UncaughtExceptionsBeforeTest">
      ...at kotlinx.coroutines.test.TestScopeImpl.enter(TestScope.kt:231)
      Suppressed: java.io.IOException: deliberate positive-control leak
        at ZzLeakOrderProbeTest$bDeliberateLeakUnderRecorder$1$2.invokeSuspend
```

**Why the suite is green anyway: ordering luck.** From the XML timestamps, backend runs 28 classes in a stable order, and all three `runTest` users — `ListenerHandlerTest`, `ScopedShipperTest`, `PullDiagnosticsPublisherTest` — run *before* `ExecProcessLeakTest` (position 27 of 28). Only `ExecuteTaskTest` follows it, and it uses `runBlocking`. So the poisoned stash is simply never collected. Any future `runTest`-based class scanned after `tasks/ExecProcessLeakTest` turns this into a red build, with the failure landing on a stranger — which is the sentence this PR's own description opens with.

**Fix, verified:** drain the stash in the same test, right after `recordUncaught` returns.

```kotlin
val seen = recordUncaught { ... }
// kotlinx-coroutines-test's process-wide ExceptionCollector also STASHES the
// deliberate leak above and would charge it to the next runTest in this JVM as
// UncaughtExceptionsBeforeTest — #101's mechanism exactly. Drain it here so it
// cannot land on a stranger.
runCatching { runTest { } }
assertTrue(seen.isNotEmpty(), ...)
```

I ran the probe again with exactly that line added: `tests="3" failures="0"`. `enter()` clears `unprocessedExceptions` while draining, so one drain is enough and later `runTest`s are clean.

### 4. 🟡 The silent catches — split verdict, not a hedge

**Stderr pump (`ExecUtil.kt:105`): silent is right.** Keep it. That pump has no other realistic failure mode; a read error on a pipe or a write error to a broken `System.err` are cases where logging is impossible anyway. Logging the ordinary close would be noise on every subprocess, as you argue.

**Connection setup (`ExecUtil.kt:128`): silent is wrong.** Different consequence. `connection` is awaited in exactly one place — `ScriptManager.getScriptWithExit:95`, inside `withTimeout(config.executeTimeout)`. Swallowing an `IOException` there means `connection` is never completed, the caller blocks for the whole execute timeout, and surfaces as `InvalidScriptException("$name could not be initialized", TimeoutCancellationException)` with the real cause gone. Before this PR the `IOException` at least reached stderr via the default handler; after it, a genuine setup failure is indistinguishable from a hang and leaves no trace anywhere. A `hauler.debug`/`warn` in that catch costs one line and one log per failed startup (not per subprocess — this path is not the normal exit route). If you want the two truly distinguishable, a `@Volatile var streamsClosed` set immediately before `proc.errorStream.close()` is an exact discriminator for both catches: closed → silent, not closed → log.

### 5. The `CoroutineExceptionHandler` is safe

- `hauler` is `hauler().asAsync(GlobalScope)`; `AsyncHauler.error` is a **non-suspend** `fun` whose `emit` is `scope.launch { … }`. Non-blocking from the handler, so no deadlock at teardown — at worst a dropped log line if the JVM exits first.
- No recursion: emission goes to the global warehouse, which never spawns an `ExecProcess`.
- Nothing that previously surfaced usefully is now swallowed. The only thing it intercepts is what used to be an uncaught stack print misattributed to a random thread; it is now logged naming `ExecProcess`.
- Worth knowing (not a defect, arguably the point): the handler is on `parentScope`, so it also covers coroutines *callers* launch there — e.g. `ScriptManager`'s force-kill timer at `exec.parentScope.launch { delay(executeTimeout); exec.kill() }`.

### 6. Stability — 5 runs, not one

`./gradlew :backend:jvmTest :protocol:jvmTest -Dintegration=true`, XML deleted before each run, counted from fresh XML:

| run | exit | wall | XML | tests | failures | errors | skipped |
|---|---|---|---|---|---|---|---|
| 1 | 0 | 55s | 30 | 151 | 0 | 0 | 6 |
| 2 | 0 | 55s | 30 | 151 | 0 | 0 | 6 |
| 3 | 0 | 54s | 30 | 151 | 0 | 0 | 6 |
| 4 | 0 | 54s | 30 | 151 | 0 | 0 | 6 |
| 5 | 0 | 57s | 30 | 151 | 0 | 0 | 6 |

**5/5 green.** `ExecProcessLeakTest` costs 3.405 / 3.406 / 3.406s across runs — dead stable, ~6% of a 55s suite, and no sign of the OOM the unthrottled writer caused. `spotlessCheck` clean. CI on `eaf1afc` is green (`build-and-test` pass, 10m36s).

### 7. Does it fix the original symptom, and is there another leak source?

Yes on the symptom — mechanically, not just by inference. `ScopedShipperTest.requestPickupDelegatesToWrappedShipper` fails via `TestScopeImpl.enter()` reading the stash that `ExceptionCollector` filled from the escaping `IOException: Stream closed`. `handleCoroutineException` consults `context[CoroutineExceptionHandler]` **first** and returns without touching the platform handlers, so with the handler on `parentScope` nothing can reach the collector — and the RED/GREEN pair above shows the exception no longer escapes at all.

Other unguarded scopes I checked in the same blast radius (**not** merge blockers, and none demonstrated to fire in 5 suite runs — flagging so they are known):

- `ScriptManager.kt:79` — `GlobalScope.launch(callSign) { …; exec.exitCode.await(); lock.unlock() }` has no handler and no `runCatching`. A double-unlock `IllegalStateException` from `Mutex.unlock()` would leak exactly like #101 did. Nearest sibling to this bug; worth a follow-up issue.
- `KonstructionServiceImpl.kt:98,406`, `KonstructorImpl.kt:58`, `PullDiagnosticsPublisher.kt:163/177/203/217`, `BridgeLanguageServer` `EngineConnection.scope` — all `CoroutineScope(… + SupervisorJob())` with no handler. Same shape, latent.
- `KonstructionController.kt:88` is fine — `runCatching` inside.

### Verdict

Changes requested on **§3** only. The production change is correct, reproduced, and stable; the fix I am asking for is one line in the test, and I have already run it green. §4's connection-setup log is a strong suggestion rather than a gate — your call, but I would take it.

monkopedia-reviewer (approved):
## Re-verified on `e3c0e39` — approving

Fresh worktree, fresh measurements, nothing inherited from your run or from my own on `eaf1afc`.

### The thing you asked me to check specifically: does the drain mask a real leak?

No, and this is measured, not reasoned. **Case A — `ExecUtil.kt` reverted to `origin/main` with the drain fully in place:**

```
<testsuite name="ExecProcessLeakTest[jvm]" tests="2" failures="1"
message="java.lang.AssertionError: Running a subprocess to completion leaked 1 uncaught
          exception(s): IOException: Stream closed"
MARKER_handler=0  MARKER_debuglog=0  HANDLER_CLASS_PRESENT=0
```

The regression test still fails for the right reason with the drain present. Your reading of the mechanism is correct: `ExceptionCollector.handleException` **adds to `unprocessedExceptions` and returns `false`**, so it falls through to `propagateExceptionFinalResort` and the thread handler fires anyway — `seen` is populated at throw time, before the `finally` ever runs. The drain consumes the collector's stash; it cannot touch the recorder's list. Two independent paths, and case A is the empirical proof.

### The drain works, proven on the shipped helper rather than a copy

I did not replicate `recordUncaught` — I drove the real one. Probe class, `@FixMethodOrder(NAME_ASCENDING)`, both preconditions satisfied: `aaPrimeTheCollector` (`runTest {}`, latches `enabled`), `bbRunTheShippedGuard` (calls `ExecProcessLeakTest().uncaughtExceptionRecorderIsNotBlind()` directly), `ccVictimRunTest` (`runTest {}`).

| case | result |
|---|---|
| D — drain as shipped | `tests="3" failures="0"` |
| E — control, `runCatching { runTest { } }` deleted from the shipped helper | `ccVictimRunTest` **FAILED**: `kotlinx.coroutines.test.UncaughtExceptionsBeforeTest` |

Same binary, same order, one line apart. That is the before/after the fix needed.

### Everything else re-measured on `e3c0e39`

| check | result |
|---|---|
| RED (revert `ExecUtil.kt`) | fails with the `Stream closed` assertion; markers 0/0, no `…$inlined$CoroutineExceptionHandler$1.class` |
| GREEN | `tests="2" failures="0"`; markers 1/1, handler class present |
| self-guard broken (`{ _, _ -> }`) | `uncaughtExceptionRecorderIsNotBlind` FAILED — *"The uncaught-exception recorder saw nothing when handed a deliberate SupervisorJob leak…"* |
| `spotlessCheck` | clean |

Stale binaries excluded at class level on every case (`ExecUtil*.class` deleted first, both new marker strings counted — `"Uncaught failure in subprocess plumbing"` and `"Process streams closed before the channel attached"`).

**Stability, 5 runs on `e3c0e39`**, XML deleted before each, counted from fresh XML, `:backend:jvmTest` never `UP-TO-DATE`:

| run | exit | wall | XML | tests | fail | err | skip | `ExecProcessLeakTest` |
|---|---|---|---|---|---|---|---|---|
| 1 | 0 | 54s | 30 | 151 | 0 | 0 | 6 | 3.406s |
| 2 | 0 | 54s | 30 | 151 | 0 | 0 | 6 | 3.407s |
| 3 | 0 | 54s | 30 | 151 | 0 | 0 | 6 | 3.406s |
| 4 | 0 | 54s | 30 | 151 | 0 | 0 | 6 | 3.407s |
| 5 | 0 | 54s | 30 | 151 | 0 | 0 | 6 | 3.406s |

**5/5.** The drain adds nothing measurable — same 3.406s as before it existed. Worktree confirmed clean at `e3c0e39` with no scratch files (`DIRTY=[]`).

### Your two questions

**The `@Volatile streamsClosed` discriminator — I agree with you, debug is proportionate.** No objection, and I would not hold the merge for it. One caveat to bank rather than act on: `debug` is likely below the configured level in production, so a genuine mid-attach failure is *recoverable* rather than *visible* — the operator has to already suspect it and turn debug on. That is still strictly better than the pre-PR state, and given `asConnection` does no I/O at construction the path is close to theoretical. If it ever bites in the field, the discriminator promoting the non-shutdown case to `warn` is the upgrade.

**Which of the six is materially riskier — yes, and I owe you a correction on my own list first.** `KonstructionServiceImpl.kt:406` (`doCallback`) does **not** belong on it: I flagged the launch site without reading the body, and it already has a full `try { … } catch (t: Throwable) { hauler.error(...) }`. My error. Five, not six.

Of the remaining five, two are worth naming separately from the class:

1. **`KonstructionServiceImpl.launchRequest:98` — riskiest by reachability.** It is the fire-and-forget shell shared by *every* `request*` RPC method, on `CoroutineScope(Dispatchers.IO + SupervisorJob())` with no handler and no `try`. `withTimeout`'s own failure is a `CancellationException` and harmless, but anything else `block()` throws leaks. What makes it stand out is that the correct shape already exists **ten lines below it** in `doCallback` — so this is an inconsistency inside one file, not a design gap.
2. **`ScriptManager.kt:79` — riskiest by consequence.** `GlobalScope`, so no owning lifecycle could ever cancel or observe it, and the body ends in `lock.unlock()`, whose `IllegalStateException` on a non-held mutex is a real state-dependent failure mode. It also sits on the exact path this PR just fixed, so it fires in the same integration tests that exposed #101.

The other three (`KonstructorImpl:58` startup-only, `PullDiagnosticsPublisher`, `BridgeLanguageServer.EngineConnection.scope` — both LSP, opt-in) are fine as one class-level issue.

### On your three controls

Worth saying plainly: the two you discarded were discarded correctly, and the reason generalises past this PR. A `--tests` filtered run is not a faster full run — it is a JVM where `ExceptionCollector.enabled` never latches, so the mechanism under test is switched off. That is the same shape as the three void reproductions in the original PR description. You caught it twice in a row on yourself, which is the harder direction.

**Approving.** Merging on completed-SUCCESS CI.

PR #108, opened 2026-08-21
fix(lsp): treat an engine that exits immediately as unavailable, and bound respawns

### LINKED ISSUE #84: LSP self-heal is blind to an expired engine, and respawns unbounded on a dead one

monkopedia-coder:
Fixes #84

## The hole

An **expired** kotlin-lsp build passes every availability check there is. The binary exists, it is executable, and `ProcessBuilder.start()` returns normally — it simply prints `This build of intellij-server has expired` and exits. So `connection` was cached as live, the documented "degrade to inert" branch was never taken, and since `textDocumentHover` fires on every cursor move, each one respawned a fresh ~2GB JVM with **no ceiling**.

The issue records that taking prod down 2026-07-28 → 08-03, on a ~40-day expiry clock. This is the *expected* failure mode, not an exotic one.

## Two halves, matching the two halves of the issue

**1. Readiness.** A freshly spawned engine must still be running after a short probe window; otherwise it is reported unavailable and LSP stays off, exactly as it does for a missing binary.

Keyed on *"did it exit immediately"* rather than on JetBrains' expiry wording, deliberately: it catches a bad launcher, a missing JRE and a wrong argument the same way, and it cannot rot when their message text changes. A healthy engine stays up for its ~120s cold index, so the two are never confused — and `waitFor` returns the instant a process exits, so the full window is only ever paid by an engine that is actually working.

**2. Bounded respawn.** 3 attempts per 60s, then a 5-minute cooldown. The existing KDoc justified having no backoff because recovery is *"paced by request traffic"* — but the traffic here is the user's cursor.

`ensureConnection` becomes `internal` so the test can drive the spawn path directly, without needing a `KsrpcLanguageClient` forwarder for a subprocess that is never going to answer.

## The test measures the thing that actually hurt

A fake engine with the real failure shape — starts fine, exits instantly — that **counts its own launches**. "How many ~2GB JVMs would this have spawned" is a number, not an argument. It drives 8 attempts against a cap of 3; a count at or below the cap could not distinguish a working bound from a broken one.

## Demonstrated RED — one per half, and the first attempt was invalid

**The first control was not evidence and I am flagging it rather than quietly replacing it.** Reverting the whole file to `main` removed the API the test compiles against, so `BUILD FAILED` was a **compile error wearing a red test's clothes** — 2 seconds, 8 tasks, never reached the test task.

Each half is now disabled surgically, leaving the API intact:

```
readiness disabled, bound intact:
  anExpiredEngineIsTreatedAsUnavailable FAILED
  expected null, but was:<com.monkopedia.ksrpc.jsonrpc.internal.JsonRpcWriterBase@3a4621bd>

bound disabled, readiness intact:
  respawnsAreBoundedForADeadEngine FAILED
  A dead engine must stop being respawned. 8 attempts spawned 8 JVMs; the cap is 3.
```

Restoring each goes green. **Two halves needed two controls:** with only the first mutation, the bound test aborted on its per-call `assertNull` and never reached the launch count, so it would have shipped on a control that never exercised it.

## Verification

`spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true` — **154 tests, 0 failures, 6 skipped**, XML deleted first, tasks executed not `UP-TO-DATE`. The +3 against main's 151 is this PR's 2 tests plus the 1 `ExecuteTaskTest` case that arrived with #92.

## Why this is queued ahead of other work

The owner's instruction was **"#84 first"**, and #103 (LSP default ON) has already merged — which does not create the exposure, but does widen it from "someone who ticked a box" to anyone's first hover. This PR is the precondition recorded on #84 for the next `:8001` deploy. **Deployment remains manual and the owner's alone; nothing here deploys anything.**

## Notes for the reviewer

- Post your verdict with `gh pr review --approve | --request-changes | --comment --body-file` via the reviewer bot. **Even a "held for owner" outcome gets a review object** (`--comment` carries the prose) — a verdict posted only as an issue comment pins to no SHA and is invisible to every review-state check.
- Worth challenging: `READY_PROBE_MILLIS = 1500` is a guess. It must exceed how long an expired build takes to print and exit, and stay far below a healthy engine's lifetime. If you think a slow machine could see a healthy engine die inside that window for an unrelated reason, say so.
- Also worth challenging: the cooldown makes LSP unavailable for 5 minutes after 3 bad spawns, including if the engine is replaced with a good one during it. That is deliberate — the alternative is probing a possibly-dead engine on every hover — but it is a real tradeoff, not an oversight.

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/lsp/KotlinLspProcess.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/lsp/KotlinLspProcess.kt
@@ -53,6 +53,12 @@ class KotlinLspProcess private constructor(private val config: Config) {
     private var process: Process? = null
     private var connection: SingleChannelConnection<String>? = null
 
+    /** Spawn timestamps inside the current window, used to bound respawn attempts. */
+    private val recentSpawns = ArrayDeque<Long>()
+
+    /** While `now < this`, spawning is refused outright and the bridge stays inert. */
+    private var cooldownUntil = 0L
+
     /**
      * Lazily spawn (once) the warm subprocess and return a fresh subprocess-facing
      * [KsrpcLanguageServer] stub wired to [forwarder]. Returns `null` if the binary is
@@ -76,13 +82,21 @@ class KotlinLspProcess private constructor(private val config: Config) {
         }
     }
 
-    private suspend fun ensureConnection(): SingleChannelConnection<String>? {
+    /**
+     * Internal rather than private so a test can drive the spawn path directly, without
+     * needing a [KsrpcLanguageClient] forwarder for a subprocess that is never going to
+     * answer. Not part of the public surface.
+     */
+    internal suspend fun ensureConnection(): SingleChannelConnection<String>? {
         connection?.let { existing ->
             if (process?.isAlive == true) return existing
             // The warm process died; drop the stale handles and respawn below.
             shutdownLocked()
         }
         val binary = config.kotlinLspBinary ?: return null
+        val now = System.currentTimeMillis()
+        if (now < cooldownUntil) return null
+        if (!admitSpawn(now)) return null
         return try {
             val systemPath = config.kotlinLspSystemPath.absolutePath
             hauler.info("Spawning kotlin-lsp: ${binary.absolutePath} (system-path=$systemPath)")
@@ -95,6 +109,26 @@ class KotlinLspProcess private constructor(private val config: Config) {
                 .redirectOutput(ProcessBuilder.Redirect.PIPE)
                 .redirectError(ProcessBuilder.Redirect.INHERIT)
                 .start()
+            if (!isStillRunning(proc)) {
+                // An EXPIRED engine lands here (konstructor#84). It is present, executable,
+                // and `start()` returns normally — it just prints "This build of
+                // intellij-server has expired" and exits. Every other availability check
+                // passes, so without this the connection is cached as live and the bridge
+                // never degrades to inert.
+                //
+                // Deliberately "did it exit immediately" rather than "does stderr say
+                // expired": it catches a bad launcher, a missing JRE and a wrong argument
+                // the same way, and it does not depend on JetBrains' wording. A healthy
+                // engine stays up for its ~120s cold index, so it is never confused with
+                // one that dies in the first second.
+                hauler.error(
+                    "kotlin-lsp exited immediately (code ${proc.exitValue()}) — treating the " +
+                        "engine as unavailable; LSP stays off. An expired build looks exactly " +
+                        "like this."
+                )
+                runCatching { proc.destroyForcibly() }
+                return null
+            }
             val conn = (proc.inputStream to proc.outputStream).asLspConnection()
             process = proc
             connection = conn
@@ -104,6 +138,40 @@ class KotlinLspProcess private constructor(private val config: Config) {
             shutdownLocked()
             null
         }
+    }
+
+    /**
+     * True if [proc] is still alive after a short readiness window.
+     *
+     * Costs [READY_PROBE_MILLIS] once per spawn, not per request — and only on the spawn
+     * path, which already takes seconds. `waitFor` returns as soon as the process exits,
+     * so the full wait is only paid by an engine that is actually staying up.
+     */
+    private fun isStillRunning(proc: Process): Boolean =
+        !proc.waitFor(READY_PROBE_MILLIS, java.util.concurrent.TimeUnit.MILLISECONDS)
+
+    /**
+     * Bound respawns. Without this, a dead-but-startable engine is re-spawned on every
+     * editor event — and `textDocumentHover` fires on every cursor move, so "paced by
+     * request traffic" means paced by how fast someone moves the mouse. Each attempt is a
+     * fresh ~2GB JVM; #84 records that taking prod down for six days.
+     */
+    private suspend fun admitSpawn(now: Long): Boolean {
+        while (recentSpawns.isNotEmpty() && now - recentSpawns.first() > SPAWN_WINDOW_MILLIS) {
+            recentSpawns.removeFirst()
+        }
+        if (recentSpawns.size >= MAX_SPAWNS_PER_WINDOW) {
+            cooldownUntil = now + COOLDOWN_MILLIS
+            recentSpawns.clear()
+            hauler.error(
+                "kotlin-lsp failed to stay up $MAX_SPAWNS_PER_WINDOW times in " +
+                    "${SPAWN_WINDOW_MILLIS / 1000}s — no further spawns for " +
+                    "${COOLDOWN_MILLIS / 1000}s. LSP is off until then."
+            )
+            return false
+        }
+        recentSpawns.addLast(now)
+        return true
     }
 
     /** Tear the subprocess down (best-effort, bounded). Safe to call multiple times. */
@@ -125,6 +193,15 @@ class KotlinLspProcess private constructor(private val config: Config) {
 
     companion object {
         private const val GRACE_MILLIS = 1_000L
+
+        /** How long a freshly spawned engine must stay up to count as usable. */
+        internal const val READY_PROBE_MILLIS = 1_500L
+
+        /** Spawn attempts allowed inside [SPAWN_WINDOW_MILLIS] before the cooldown. */
+        internal const val MAX_SPAWNS_PER_WINDOW = 3
+
+        internal const val SPAWN_WINDOW_MILLIS = 60_000L
+        internal const val COOLDOWN_MILLIS = 5 * 60_000L
 
         // One warm subprocess per Config (i.e. per backend). The classpath is fixed, so
         // a single instance serves all konstructions (Phase 5 hardens multiplexing).
--- a/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/lsp/KotlinLspProcessExpiredEngineTest.kt
+++ b/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/lsp/KotlinLspProcessExpiredEngineTest.kt
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2022 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.monkopedia.konstructor.lsp
+
+import com.monkopedia.konstructor.Config
+import java.io.File
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+import kotlin.test.assertTrue
+import kotlinx.coroutines.runBlocking
+import org.junit.After
+import org.junit.Before
+
+/**
+ * Regression test for konstructor#84: an **expired** kotlin-lsp engine must be treated as
+ * unavailable, and a dead engine must not be respawned without limit.
+ *
+ * The trap this covers is that an expired EAP build passes every availability check there
+ * is — the binary exists, it is executable, and `ProcessBuilder.start()` returns normally.
+ * It simply prints "This build of intellij-server has expired" and exits. So the bridge
+ * cached the connection as live and never degraded to inert, and because
+ * `textDocumentHover` fires on every cursor move, each one respawned a fresh ~2GB JVM.
+ * #84 records that taking prod down for six days.
+ *
+ * The fake engine here reproduces exactly that shape — starts fine, exits immediately —
+ * and counts its own launches, so "how many JVMs would this have spawned" is measured
+ * rather than argued.
+ */
+class KotlinLspProcessExpiredEngineTest {
+
+    private lateinit var tempDir: File
+    private lateinit var launchLog: File
+    private lateinit var fakeEngine: File
+
+    @Before
+    fun setUp() {
+        tempDir = kotlin.io.path.createTempDirectory("lsp-expired-test-").toFile()
+        launchLog = File(tempDir, "launches.txt")
+        fakeEngine = File(tempDir, "fake-intellij-server").apply {
+            writeText(
+                """
+                #!/bin/bash
+                echo launch >> "${launchLog.absolutePath}"
+                echo "This build of intellij-server has expired" >&2
+                exit 1
+                """.trimIndent()
+            )
+            setExecutable(true)
+        }
+    }
+
+    @After
+    fun tearDown() {
+        tempDir.deleteRecursively()
+    }
+
+    private fun launches(): Int =
+        if (launchLog.exists()) launchLog.readLines().count { it.isNotBlank() } else 0
+
+    @Test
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
Independent review of `a50b7f0`. I did not write this change and took nothing in the PR body as evidence — every claim below is one I reproduced in my own detached worktree off the PR head.

## Verdict: approve

Both REDs reproduce. No correctness defect found. The two halves are genuinely independent and the second one backstops the first, which is what makes the weaker constant survivable. Findings 4–8 below are follow-up material, not merge blockers.

---

## 1. Both REDs reproduced, surgically, with the API left intact

Method: fresh detached worktree at `a50b7f0`, one half disabled at a time, the test file never touched.

**Control A — readiness disabled, bound intact.** `isStillRunning(proc)` replaced with `= true` (probe removed, `READY_PROBE_MILLIS` and every signature kept):

```
> Task :backend:compileKotlinJvm            <-- EXECUTED
> Task :backend:compileTestKotlinJvm UP-TO-DATE
> Task :backend:jvmTest FAILED

KotlinLspProcessExpiredEngineTest[jvm] > anExpiredEngineIsTreatedAsUnavailable[jvm] FAILED
  java.lang.AssertionError: An engine that starts and immediately exits must be reported as
  unavailable so the bridge degrades to inert. It launched 1 time(s).
  expected null, but was:<com.monkopedia.ksrpc.jsonrpc.internal.JsonRpcWriterBase@335b5620>

KotlinLspProcessExpiredEngineTest[jvm] > respawnsAreBoundedForADeadEngine[jvm] FAILED
  java.lang.AssertionError: a dead engine must never yield a connection
  expected null, but was:<com.monkopedia.ksrpc.jsonrpc.internal.JsonRpcWriterBase@51acdf2e>
```
Fresh XML: `tests="2" skipped="0" failures="2" errors="0"`.

**Control B — bound disabled, readiness intact.** `recentSpawns.size >= MAX_SPAWNS_PER_WINDOW` → `>= Int.MAX_VALUE` (the constant stays, the gate never trips):

```
> Task :backend:compileKotlinJvm            <-- EXECUTED
> Task :backend:jvmTest FAILED

KotlinLspProcessExpiredEngineTest[jvm] > respawnsAreBoundedForADeadEngine[jvm] FAILED
  java.lang.AssertionError: A dead engine must stop being respawned. 8 attempts spawned 8 JVMs;
  the cap is 3. Unbounded here is what took prod down (#84).
```
Fresh XML: `tests="2" skipped="0" failures="1" errors="0"`.

**That `failures="1"` is the load-bearing detail.** Only the bound test failed — the readiness half still held, so the per-call `assertNull` passed and the test actually reached the launch count. Control A produces `failures="2"`. The two controls therefore fail for *different* reasons, which is exactly the property the author's discarded first attempt lacked. The concern that the second test would ship on a control that never exercised it is answered by the count.

**Compile-error vs test-failure.** Both controls print `BUILD FAILED in 2s` — the same 2-second signature that made the first attempt invalid — and both are nevertheless genuine test failures:
- `:backend:compileKotlinJvm` is listed as **executed**, not `UP-TO-DATE` (the mutation really compiled);
- `:backend:compileTestKotlinJvm` / `:backend:jvmTestClasses` are `UP-TO-DATE` — the test source did **not** need recompiling, i.e. the API it binds against was intact, which is precisely what the reverted-whole-file control destroyed;
- `:backend:jvmTest` was **reached** and reported `FAILED`, with `9 executed, 45 up-to-date`;
- a fresh `TEST-…KotlinLspProcessExpiredEngineTest.xml` was written each time, carrying the assertion text quoted above.

The 2 seconds is warm upstream tasks, not an early abort. Restoring the file returns 2/2 green.

## 2. `READY_PROBE_MILLIS = 1500` — defensible, but it is the weaker half, and that is fine

**False negative (a healthy engine written off) is not a real risk.** For this to fire, an IntelliJ-platform JVM would have to exit within 1.5s of `start()`. It cannot do that and be healthy — a launcher that dies that fast (bad JRE, `--system-path` lock conflict, wrong argument) is a genuine unavailability, and reporting it as such is correct. A slow or loaded machine makes a healthy engine *slower* to become ready, not more likely to exit. So I see no path where a working engine gets permanently written off.

**False positive (an expired engine slipping through) is plausible and I cannot rule it out.** `bin/intellij-server` is a launcher script fronting a JVM, and the expiry check runs after platform startup. JVM boot plus platform init can exceed 1.5s on a loaded box, in which case the probe sees a live process and caches the connection. I have no engine binary on this machine, so I did not measure it and I am not going to assert a number. **Neither has anyone else** — the author calls it a guess and I agree with that self-assessment.

**What makes it mergeable anyway is the bound.** In the slipped-through case the flow degrades to exactly today's behaviour — connection cached, first hover fails, `withLiveDelegate` → `reconnect` → `ensureConnection` sees `process.isAlive == false` → respawn — except `admitSpawn` now caps it at 3 spawns per 60s and then 5 minutes of nothing. That is ~36 spawn attempts per hour instead of one per cursor move. The outage in #84 does not recur even when the probe is mistuned. The bound is the load-bearing half; the probe is the fast path.

Two things follow, both non-blocking:
- If someone with an expired binary ever measures the exit latency, raising the constant is nearly free — a healthy engine pays it once per *process* lifetime (the process is kept warm across konstructions and sessions), against a ~120s cold index. 1500ms is not obviously the right number and 5s would cost nothing noticeable.
- There is a **third** expiry shape this does not cover, recorded from the 2026-06-05 incident: an expired engine that *stays up* and never answers `initialize`. The probe passes it, and nothing in `BridgeLanguageServer` puts a timeout on the delegate `initialize` (`callTimeout` in `KonstructionServiceImpl` wraps the `lsp()` factory call, not the bridge's own methods), so the frontend's `client.initialize()` would hang. That is silent-LSP-death, not the respawn storm, so it is out of scope for #84 — but it is the case #84's "readiness handshake" alternative would have caught and the liveness probe does not. Worth a follow-up issue.

## 3. The 5-minute cooldown — acceptable, but there is genuinely no manual escape

Confirmed by reading, not by taking the PR's word: `cooldownUntil` is private, cleared only by the passage of time. `KotlinLspProcess.shutdown()` does not reset it — and `git grep` shows **`shutdown()` has no production callers at all**; `BridgeLanguageServer.shutdown()` is a different method that deliberately does not touch the shared process. `instances` is a static map that is never evicted, and `App` constructs exactly one `Config`, so toggling LSP off/on in Settings, closing and reopening a konstruction, or switching workspaces all land back on the same `KotlinLspProcess` with the same live cooldown.

So the only ways out are: wait ≤5 minutes, or restart the backend. Note that dropping a *good* binary at the same path does **not** need a restart otherwise (`kotlinLspBinary` is a `File` resolved at `Config` construction, and the path is re-exec'd each spawn) — so the cooldown is the only thing that makes an in-place engine swap wait.

I think that is an acceptable trade. It is bounded, it self-clears, and the alternative the author names — probing a possibly-dead engine on every hover — is the bug being fixed. The obvious cheap improvement is for `shutdown()` to clear `cooldownUntil`/`recentSpawns`, which is safe precisely *because* nothing calls it today, and would give a future admin/debug hook somewhere to hang off. Follow-up, not a blocker.

## 4. Happy-path cost — the claim checks out

`Process.waitFor(timeout, unit)` returns `false` on timeout and `true` the moment the process exits, so `isStillRunning` returns early for a dying engine and burns the full 1500ms only for one that stays up. It runs inside `ensureConnection`, which runs only when there is no live cached process: `connection?.let { if (process?.isAlive == true) return existing }` short-circuits every subsequent call. So it is **once per subprocess lifetime**, not once per request, and never on the hover/completion/diagnostics path. Confirmed by reading the call graph, and consistent with the 4 `This build of intellij-server has expired` lines that appear in the test output — 1 for the first test, 3 for the second, i.e. the cap, visible in the raw log.

One nit: `waitFor` blocks the calling *thread* for up to 1.5s inside a `suspend fun`, with the `Mutex` held, and there is no `withContext(Dispatchers.IO)`. It matches the pre-existing style (`shutdownLocked` does the same with `GRACE_MILLIS`) and the event is rare, so I am not asking for a change — but it is now 1.5x longer than the existing one.

## 5. `ensureConnection` → `internal` — acceptable

`internal` is module-scoped, and the backend's `jvmTest` compilation is a friend of `jvmMain`, so this widens nothing that ships. There *is* an alternative seam — `com.monkopedia.lsp.DefaultLanguageClient` is concrete and already subclassed in this file, so the test could have driven `connect(object : DefaultLanguageClient() {})` and asserted on the returned stub; under Control A that would have gone red too. I marginally prefer that as the more end-to-end assertion, but the author's reason (no forwarder for a subprocess that will never answer) is legitimate and this is a preference, not a defect.

The one thing I would add: `ensureConnection` mutates `recentSpawns`/`cooldownUntil`/`process`/`connection` and **requires the caller to hold `lock`** — `shutdownLocked` says so in its name, this does not say so anywhere. Now that an in-module caller can reach it, one KDoc line ("caller must hold [lock]") would keep that contract from rotting.

## 6. End-to-end inertness — yes, genuinely identical to having no binary

Traced rather than assumed:
- `BridgeLanguageServer.initialize` → `connectEngine()` → `KotlinLspProcess.connect()` → `ensureConnection()` returns null → `connection` is left **null**, `hauler.error("kotlin-lsp engine unavailable; bridge is inert")`, and it returns `InitializeResult(capabilities = ServerCapabilities())`.
- Every editor-facing forward call goes through `withLiveDelegate`, whose first line is `val dead = connection ?: return fallback`. With `connection` null it never enters the `try`, so it never throws, never calls `reconnect`, and therefore **never reaches the spawn path again**. `textDocumentHover` returns `null`, completion returns `null`, didOpen/didChange are no-ops.
- That is byte-for-byte the same path taken when `kotlinLspBinary` is unset (`connect()` bails at `if (!config.isKotlinLspAvailable) return null`), including the same log line.
- On the frontend, `EditorPane`'s `produceState` is keyed only on `lspEnabled` and `currentKonstruction`, so there is no retry loop; empty capabilities simply mean kodemirror asks for nothing. Nothing is surfaced to the user.

This is the part I most wanted to falsify and could not. Pre-fix, the storm existed precisely because `ensureConnection` handed back a live-looking connection over a dead process's pipes, so `connection` was non-null and every hover took the `catch` → `reconnect` → respawn path. Making `ensureConnection` return null cuts the loop at its root, not at a symptom.

Residual cost versus a missing binary: an expired engine still costs one spawn attempt plus a 1500ms probe per `initialize`, capped at 3 per 60s. Bounded and acceptable.

## 7. Concurrency — clean on every path

- Only production caller of `ensureConnection` is `connect()`, and it is `lock.withLock { ensureConnection() }`. All three new early returns (`now < cooldownUntil`, `!admitSpawn(now)`, `!isStillRunning(proc)`) are inside `ensureConnection`, hence inside the same critical section; the non-local `return null` from inside the `try` still unwinds through `withLock`'s finally. `recentSpawns` and `cooldownUntil` are never touched anywhere else. Verified by grep, not by inspection of the diff alone.
- `forConfig` is `@Synchronized` over `getOrPut`, so two callers cannot both construct for the same key. `Config` is a plain class with no `equals`/`hashCode`, so the map is identity-keyed — which is fine here because `App` builds exactly one `Config` (`App.kt:57`) and every `BridgeLanguageServer` is handed that same instance. Worth being aware of: if a second `Config` were ever constructed in-process, each would get its own `KotlinLspProcess` and its own independent spawn budget. Not reachable today.
- The test's direct `ensureConnection()` calls run lock-free, but sequentially inside `runBlocking`, so they do not prove or disprove anything about the lock. See the KDoc note in §5.

## 8. Smaller notes (all follow-up)

- `recentSpawns`/`cooldownUntil` use `System.currentTimeMillis()` for interval measurement. `nanoTime` is the monotonic source; a backwards wall-clock step could extend a cooldown. Pedantic on a slewed NTP host.
- On the readiness-fail path, `proc.destroyForcibly()` is called but the stdin/stdout pipes are not closed. `shutdownLocked` only closes them in the `isAlive` branch, so this is pre-existing, and at ≤3 leaked pairs per 5 minutes it is not a practical concern.
- `admitSpawn` is `suspend` with no suspension point.
- The tests leave two entries in the static `instances` map pointing at deleted temp dirs. Harmless in a test JVM.
- Mild flake surface: `anExpiredEngineIsTreatedAsUnavailable` fails if the fake `bash` engine ever takes >1500ms to fork, write a line and exit. On a pathologically loaded runner that is not impossible. I ran the suite three times and saw no instability (below), so I am recording it rather than objecting.

## Verification I ran myself

Full CI-equivalent suite (`spotlessCheck test :backend:jvmTest :protocol:jvmTest -Dintegration=true`), XML deleted with the `find … -exec rm -rf {} +` form before **each** run so the test tasks could not come back `UP-TO-DATE`:

| run | exit | fresh result XML files |
|-----|------|------------------------|
| 1 | 0 | 31 |
| 2 | 0 | 31 |
| 3 | 0 | 31 |

Tallied from run 3's fresh XML: **154 tests, 6 skipped, 0 failures, 0 errors** across 31 files — matching the number claimed in the PR body. Each run: 26 `spotless*` tasks executed, `:backend:jvmTest` and `:protocol:jvmTest` both executed (not `UP-TO-DATE`), `BUILD SUCCESSFUL in 57s`.

`KotlinLspProcessExpiredEngineTest` reports `tests="2" failures="0" time="0.031"` in every run — 31 milliseconds for both tests, which is also the empirical confirmation that `waitFor` returns the instant the process exits rather than burning the 1500ms window. And each run's log carries exactly **4** `This build of intellij-server has expired` lines from the fake engine's inherited stderr: 1 for the first test, 3 for the second. The cap is visible in the raw build output, independent of the assertion.

`spotlessCheck` passes on the head. Both new tests are in every run.

---

Approving. The change closes both halves of #84 as written, the controls are honest, and the failure mode it does not close (a hung expired engine) is a different, non-outage symptom that deserves its own issue rather than blocking this one.

PR #111, opened 2026-08-22
fix(backend): claim ids atomically so concurrent creates cannot collide

### LINKED ISSUE #102: create() is not atomic: concurrent creates mint the same id and silently overwrite each other's info.json
## What happens

`create()` on both `KonstructorImpl` (workspaces) and `WorkspaceImpl` (konstructions) allocates an id by *scanning* the directory for used ids and then, in a separate step, writing the new `info.json`. Nothing is held between the two. Two concurrent `create()` calls can therefore mint the **same id**, and the "id already used" guard does not reliably catch it:

```kotlin
// KonstructorImpl / WorkspaceImpl
val newItemWithId = newItem.copy(id = newItem.id.ifEmpty { generateId() })  // scan -> pick
writeNewInfo(newItemWithId.infoFile, newItemWithId.id, config.json, newItemWithId)

// InfoStore.writeNewInfo
if (targetInfo.exists() || targetInfo.parentFile.exists()) throw IllegalArgumentException(...)
targetInfo.parentFile.mkdirs()          // <- gap: both racers can pass the check above
targetInfo.outputStream().use { ... }   //    before either reaches this line
```

Two racers that both pass the `exists()` check before either `mkdirs()` both proceed, and the second `encodeToStream` **overwrites the first record's `info.json`**. Both callers are returned a `Space`/`Konstruction` reporting success. From then on the losing client holds an id that resolves to *someone else's* object — every subsequent read, rename, compile and delete on that handle silently addresses the wrong record. Nothing errors.

## Reproduction

8 concurrent `create()` calls released from a common gate, 40 rounds, fresh temp data dir per round (each racer taking its own `konstructor.get(ws)` handle, exactly as separate ksrpc calls do):

| | rounds | rounds losing >=1 record | records lost | rounds where **every** caller got success and records were still lost |
| --- | --- | --- | --- | --- |
| `KonstructorImpl.create` (workspaces) | 40 | 24 | 41 | 2 |
| `WorkspaceImpl.create` (konstructions) | 40 | 24 | 40 | 0 |

A fully silent round, verbatim:

```
WS round 2: ok=8 err=0 returnedIds=[2, 0, 1, 1, 2, 4, 3, 0] distinct=5 onDisk=5
            diskNames=[0=w1, 1=w2, 2=w0, 3=w6, 4=w5] errs=[]
```

Eight creates, **zero exceptions**, five records on disk. `w3` was told it is id `1`; id `1` on disk is `w2`. `w4` was told id `2`; that is `w0`. `w7` was told id `0`; that is `w1`. Three clients are now driving another client's object with no error anywhere.

The majority of rounds are "loud" (`IllegalArgumentException: N has been already used` to some racers, which is at least visible), but the loud rounds still lose records: a racer that got an exception is a failed create, whereas a racer that got a *success* pointing at a record another caller overwrote is silent corruption.

Two other symptoms fell out of the same runs:

- `list()` reads a **torn** `info.json` — `JsonDecodingException: Expected colon ':', but had 'EOF' instead at path: $.konstruction` — because the scan can open a file another `create()` is midway through writing. One torn file fails the **whole** listing, not just that entry.
- `PathController[workspaceId, id]` does `File(workspaceDir, id).also { it.mkdirs() }`, so a directory can exist with no `info.json`. `listInfo` skips such a directory (so `firstFreeId` re-offers the id) while `writeNewInfo` treats it as taken — that id becomes permanently unallocatable, failing every `create()` that reaches it.

## Scope / exposure

- `KonstructorImpl` is a **single instance shared by every connection** (`App.createChannel()` serializes one `service`), and its `Mutex` guards only the `konstructionLookup` map in `konstruction()` — not `create`, `list` or `delete`.
- `WorkspaceImpl` is constructed fresh per `Konstructor.get(id)` call, so a per-instance lock would not help it either.
- Frontend calls arrive from `viewModelScope.launch` per dialog confirm (`NavigationDialogViewModel.mutate`), with no client-side serialization; two tabs, two users, or a double-fired create race directly.

## Not introduced by #100

#100 moved this code into `InfoStore.writeNewInfo`/`firstFreeId` without changing the sequence. Same probe with the production files reverted to the pre-#100 commit: 23/39 and 27/38 rounds losing records (37 and 52 records). Statistically indistinguishable from the post-#100 numbers above — pre-existing, and not widened. (If anything the hash-set id lookup shortens the scan-to-claim window slightly.)

## Suggested direction

Make claiming an id atomic with allocating it, rather than checking-then-writing:

1. Use `File.mkdir()` (not `mkdirs()`) as the claim — it is atomic and returns `false` if the directory already exists — and loop `firstFreeId` -> `mkdir` until one succeeds, before writing `info.json`. That closes the window without a lock and is multi-process safe.
2. Write `info.json` to a temp file in the same directory and `Files.move(..., ATOMIC_MOVE)` it into place, so `list()` can never observe a partially written file.
3. Failing (1), a per-container `Mutex` held across scan+write would fix the in-process case, but `WorkspaceImpl`'s per-call construction means it would have to live somewhere shared (e.g. keyed off `config.dataDir` on `KonstructorImpl`).

Reproduction harness available on request; it is a ~120-line JUnit test using `TestEnvironment` and a `CompletableDeferred` start gate.

_Filed from the independent review of #100._

monkopedia-coder:
Fixes #102

## The defect was silent, which is the whole problem

`create()` picked the lowest free id and *then* wrote. Between those two steps another caller could pick the same number. Measured on `main`: **eight concurrent creates, zero exceptions, five records on disk.** Three callers walked away holding an id that resolves to someone else's object, and nothing anywhere reported a fault.

## The fix

`mkdir` — single, not `mkdirs` — is now the claim primitive. On POSIX exactly one concurrent caller can create a given directory, so the winner is decided by the kernel instead of by a check-then-create window.

Two details matter as much as the primitive itself:

**Only auto-allocated ids retry.** A caller who supplied an explicit id and lost still gets `IllegalArgumentException` — retrying would hand them a *different object* than the one they asked for. Losing a race for an id we picked ourselves is not a failure, so it re-derives and tries again.

**Free ids now come from directory names, not from decoded `info.json`.** This fixes a second bug living in the same window: `PathController.get()` creates directories eagerly on a **read** path, and the old allocator skipped info-less directories — so it offered those ids forever while the claim rejected them forever. Permanently unallocatable. It is also what makes the retry *converge* rather than spin on the same number, because a racer's claim becomes visible the instant it lands.

## Demonstrated RED

Atomic `mkdir` swapped back to `exists()`-then-`mkdirs()` — the pre-fix shape — with the API left intact so a compile error cannot masquerade as a red test:

```
concurrentCreatesAllGetDistinctIds FAILED
round 0: 8 concurrent creates returned 4 DISTINCT ids ([3, 3, 3, 3, 3, 2, 1, 0])
 — duplicates mean two callers hold the same id and one silently overwrote the other.
expected:<8> but was:<4>
```

Five of eight callers got id `3`. Restoring goes green.

## Why the test races real callers

A single-caller test passes against the broken code — the window only opens when two callers are inside it. So the test releases 8 coroutines off a shared gate, 25 rounds, and asserts on the **result set**: every returned id distinct, and one record on disk per success. It has to assert on the aggregate rather than on an error, because **the failure mode produces no exception to catch**.

Two companion tests cover the halves that are easy to regress in opposite directions: an explicitly-requested id that is taken must still fail (not silently retry), and an id whose directory exists without `info.json` must not be re-offered.

## Verification

`spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true` run twice — **159 tests, 0 failures, 6 skipped** both times, XML deleted first, tasks executed not `UP-TO-DATE`.

## Notes for the reviewer

- Post your verdict as a **review object** (`gh pr review --approve | --request-changes | --comment --body-file`) via the reviewer bot — even a held-for-owner outcome gets one. A verdict left as an issue comment pins to no SHA and is invisible to every review-state check.
- Worth challenging: `ID_CLAIM_ATTEMPTS = 32` is a bound on retries, not a proof of termination. With N concurrent creators each losing at most N-1 times, 32 covers far more concurrency than this service will ever see — but it is a bound, and if you think the exhaustion path should degrade differently than `IllegalStateException`, say so.
- Also worth challenging: deriving used ids from directory names means a stray non-item directory under the data dir now consumes an id. That is deliberate (it is the only way to make the retry converge) but it is a behaviour change worth a second opinion.

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/InfoStore.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/InfoStore.kt
@@ -133,6 +133,10 @@ internal inline fun <reified T> createWithClaimedId(
         // claim is already visible and the next pass picks a different candidate — this
         // converges rather than spinning on the same number.
     }
+    // Loud, not silent: callContext logs it and the caller sees the failure. Exhaustion
+    // is reachable under sustained concurrent creation (measured: ~1% at 4 concurrent
+    // writers, 6% at 8) — NOT, as an earlier version of this comment claimed, bounded by
+    // "N creators lose at most N-1 times". That holds only for a single burst.
     throw IllegalStateException(
         "Could not claim a free id under ${parent.absolutePath} after $ID_CLAIM_ATTEMPTS " +
             "attempts; something is creating items faster than ids can be allocated."
@@ -156,13 +160,22 @@ internal const val ID_CLAIM_ATTEMPTS = 32
  * file is a sibling so the move stays within one filesystem, which is what makes it atomic.
  */
 internal inline fun <reified T> writeInfo(targetInfo: File, json: Json, value: T) {
-    val tmp = File(targetInfo.parentFile, "${targetInfo.name}.tmp")
+    // The temp name must be UNIQUE per write, and it must NOT be deleted on success.
+    // A fixed name plus `finally { delete() }` made two concurrent writers to the same
+    // info.json collide: writer A's cleanup unlinked writer B's in-flight temp, and
+    // 628 of 1200 writes threw NoSuchFileException on the move. No corruption — the
+    // target always decoded — but a rename that used to succeed started failing.
+    val tmp = File.createTempFile("${targetInfo.name}.", ".tmp", targetInfo.parentFile)
+    var moved = false
     try {
         tmp.outputStream().use { output ->
             json.encodeToStream(value, output)
         }
         Files.move(tmp.toPath(), targetInfo.toPath(), ATOMIC_MOVE, REPLACE_EXISTING)
+        moved = true
     } finally {
-        tmp.delete()
+        // On success the temp no longer exists under this name; deleting unconditionally
+        // is what let one writer reach into another's.
+        if (!moved) tmp.delete()
     }
 }
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/InfoStore.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/InfoStore.kt
@@ -18,6 +18,9 @@
 package com.monkopedia.konstructor
 
 import java.io.File
+import java.nio.file.Files
+import java.nio.file.StandardCopyOption.ATOMIC_MOVE
+import java.nio.file.StandardCopyOption.REPLACE_EXISTING
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.decodeFromStream
@@ -139,8 +142,27 @@ internal inline fun <reified T> createWithClaimedId(
 /** How many times to re-derive a free id when losing a concurrent claim. */
 internal const val ID_CLAIM_ATTEMPTS = 32
 
+/**
+ * Write [value] to [targetInfo] so a concurrent reader never sees it half-written.
+ *
+ * `outputStream()` TRUNCATES on open, so writing in place leaves a window where the file
+ * exists but is incomplete — and [listInfo] decodes every entry, so one torn file fails
+ * the WHOLE listing. Measured on the in-place version: 18 of 1167 concurrent listings
+ * failed (1.5%), rising to 15% with larger payloads, with errors like
+ * `JsonDecodingException: Expected colon ':', but had 'EOF' instead at path: $.id`.
+ *
+ * Writing to a sibling temp file and moving it into place with [ATOMIC_MOVE] closes that:
+ * a reader sees either the previous file or the complete new one, never a prefix. The temp
+ * file is a sibling so the move stays within one filesystem, which is what makes it atomic.
+ */
 internal inline fun <reified T> writeInfo(targetInfo: File, json: Json, value: T) {
-    targetInfo.outputStream().use { output ->
-        json.encodeToStream(value, output)
+    val tmp = File(targetInfo.parentFile, "${targetInfo.name}.tmp")
+    try {
+        tmp.outputStream().use { output ->
+            json.encodeToStream(value, output)
+        }
+        Files.move(tmp.toPath(), targetInfo.toPath(), ATOMIC_MOVE, REPLACE_EXISTING)
+    } finally {
+        tmp.delete()
     }
 }
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/InfoStore.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/InfoStore.kt
@@ -52,6 +52,19 @@ internal inline fun <reified T> File.listInfo(json: Json): List<T> =
     } ?: emptyList()
 
 /**
+ * Every immediate subdirectory name under [this], i.e. every id that is TAKEN.
+ *
+ * Deliberately based on directory names rather than on successfully decoded [INFO_JSON]
+ * files. A directory whose info has not been written yet — because a concurrent create is
+ * mid-flight, or because [com.monkopedia.konstructor.PathController] created it eagerly on
+ * a read path — still owns its id. Deriving free ids from decoded info instead re-offered
+ * those ids forever: [listInfo] skipped them, so they never looked used, while the claim
+ * below always rejected them. See konstructor#102.
+ */
+internal fun File.usedIds(): Collection<String> =
+    listFiles()?.filter { it.isDirectory }?.map { it.name } ?: emptyList()
+
+/**
  * The lowest non-negative integer, as a string, that is not in [usedIds].
  *
  * Membership is tested against a hash set, so allocating an id is linear in the number
@@ -67,17 +80,66 @@ internal fun firstFreeId(usedIds: Collection<String>): String {
 }
 
 /**
- * Claim [targetInfo] for a new item and write [value] to it.
+ * Attempts to claim [id] under [parent], atomically.
  *
- * Both the file and its directory must be absent — an existing directory means the id is
- * taken even when its [INFO_JSON] was never written — otherwise this throws
- * [IllegalArgumentException] naming [id].
+ * `mkdir` (single, not `mkdirs`) is the claim primitive because on POSIX exactly one
+ * concurrent caller can create a given directory — everyone else gets `false`. The
+ * previous `exists()`-then-`mkdirs()` pair was a TOCTOU window wide enough that eight
+ * concurrent creates produced five records with zero exceptions (konstructor#102).
+ *
+ * Returns the claimed directory, or `null` if someone else already owns the id.
  */
-internal inline fun <reified T> writeNewInfo(targetInfo: File, id: String, json: Json, value: T) {
-    if (targetInfo.exists() || targetInfo.parentFile.exists()) {
-        throw IllegalArgumentException("$id has been used already")
+internal fun tryClaimId(parent: File, id: String): File? {
+    parent.mkdirs()
+    val dir = File(parent, id)
+    return if (dir.mkdir()) dir else null
+}
+
+/**
+ * Create a new item under [parent] and write its [INFO_JSON], claiming the id atomically.
+ *
+ * If [requestedId] is non-empty the caller chose it, and losing the claim is an ERROR —
+ * they asked for a specific id and it is taken, which is exactly what
+ * [IllegalArgumentException] should tell them. If it is empty the id is ours to pick, so
+ * losing a race is not a failure: re-derive the lowest free id and try again. Retrying a
+ * caller-supplied id would silently hand them a different object than they asked for.
+ *
+ * [buildValue] receives the id that was actually claimed, so callers can stamp it into the
+ * payload they persist.
+ */
+internal inline fun <reified T> createWithClaimedId(
+    parent: File,
+    requestedId: String,
+    json: Json,
+    buildValue: (String) -> T
+): String {
+    if (requestedId.isNotEmpty()) {
+        val dir = tryClaimId(parent, requestedId)
+            ?: throw IllegalArgumentException("$requestedId has been used already")
+        writeInfo(File(dir, INFO_JSON), json, buildValue(requestedId))
+        return requestedId
     }
-    targetInfo.parentFile.mkdirs()
+    repeat(ID_CLAIM_ATTEMPTS) {
+        val candidate = firstFreeId(parent.usedIds())
+        val dir = tryClaimId(parent, candidate)
+        if (dir != null) {
+            writeInfo(File(dir, INFO_JSON), json, buildValue(candidate))
+            return candidate
+        }
+        // Lost the race for this id. `usedIds` reads directory names, so the winner's
+        // claim is already visible and the next pass picks a different candidate — this
+        // converges rather than spinning on the same number.
+    }
+    throw IllegalStateException(
+        "Could not claim a free id under ${parent.absolutePath} after $ID_CLAIM_ATTEMPTS " +
+            "attempts; something is creating items faster than ids can be allocated."
+    )
+}
+
+/** How many times to re-derive a free id when losing a concurrent claim. */
+internal const val ID_CLAIM_ATTEMPTS = 32
+
+internal inline fun <reified T> writeInfo(targetInfo: File, json: Json, value: T) {
     targetInfo.outputStream().use { output ->
         json.encodeToStream(value, output)
     }
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionController.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionController.kt
@@ -88,9 +88,11 @@ class KonstructionControllerImpl(
             GlobalScope.launch(saveContext + callSign) {
                 runCatching {
                     contentFileLock.withLock {
-                        paths.infoFile.outputStream().use { output ->
-                            config.json.encodeToStream(value, output)
-                        }
+                        // Atomic: this is the SAME file WorkspaceImpl.list() decodes, and
+                        // this setter fires on rename, dirty-state and target updates —
+                        // far more often than create writes it. In-place, 28% of
+                        // concurrent listings saw a torn file (#102).
+                        writeInfo(paths.infoFile, config.json, value)
                     }
                 }.onFailure { hauler.error("Failed to persist info for $workspaceId/$id", it) }
                 // Always set one more time after write to settle out any race conditions.
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructorImpl.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructorImpl.kt
@@ -85,15 +85,13 @@ class KonstructorImpl(private val config: Config) :
     }
 
     override suspend fun create(newItem: Space): Space = callContext("create") {
-        val newItemWithId = newItem.copy(
-            id = newItem.id.ifEmpty { generateId() }
-        )
-        writeNewInfo(newItemWithId.infoFile, newItemWithId.id, config.json, newItemWithId)
-        newItemWithId
-    }
-
-    private suspend fun generateId(): String = callContext("generateId") {
-        firstFreeId(list().map { it.id })
+        // The id is chosen and claimed in one atomic step; see createWithClaimedId. Doing
+        // it as "pick a free id, then write" let two concurrent creates pick the same id
+        // and silently overwrite each other (konstructor#102).
+        val claimedId = createWithClaimedId(config.dataDir, newItem.id, config.json) { id ->
+            newItem.copy(id = id)
+        }
+        newItem.copy(id = claimedId)
     }
 
     override suspend fun delete(item: Space): Unit = callContext("delete") {
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/WorkspaceImpl.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/WorkspaceImpl.kt
@@ -49,9 +49,9 @@ class WorkspaceImpl(private val config: Config, private val workspaceId: String)
         val info = infoFile.inputStream().use { input ->
             config.json.decodeFromStream<Space>(input)
         }.copy(name = name)
-        infoFile.outputStream().use { output ->
-            config.json.encodeToStream(info, output)
-        }
+        // Atomic, for the same reason create() is — an in-place rewrite is visible to a
+        // concurrent reader as a truncated file. See writeInfo.
+        writeInfo(infoFile, config.json, info)
     }
 
     override suspend fun create(newItem: Konstruction): Konstruction = callContext("create") {
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/WorkspaceImpl.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/WorkspaceImpl.kt
@@ -58,19 +58,12 @@ class WorkspaceImpl(private val config: Config, private val workspaceId: String)
         if (newItem.workspaceId != workspaceId) {
             throw IllegalArgumentException("Trying to create item in wrong workspace")
         }
-        val newItemWithId = newItem.copy(
-            id = newItem.id.ifEmpty { generateId() }
-        )
-        writeNewInfo(
-            newItemWithId.infoFile,
-            newItemWithId.id,
-            config.json,
-            KonstructionInfo(newItemWithId, CLEAN, emptyList())
-        )
-        newItemWithId
+        // Atomic claim — see createWithClaimedId and konstructor#102.
+        val claimedId = createWithClaimedId(workspaceDir, newItem.id, config.json) { id ->
+            KonstructionInfo(newItem.copy(id = id), CLEAN, emptyList())
+        }
+        newItem.copy(id = claimedId)
     }
-
-    private suspend fun generateId(): String = firstFreeId(list().map { it.id })
 
     override suspend fun delete(item: Konstruction): Unit = callContext("delete") {
         val targetInfo = item.infoFile
--- a/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/ConcurrentCreateTest.kt
+++ b/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/ConcurrentCreateTest.kt
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:OptIn(ExperimentalSerializationApi::class)
+
 package com.monkopedia.konstructor
 
 import com.monkopedia.konstructor.common.Space
@@ -27,6 +29,8 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
+import kotlinx.serialization.ExperimentalSerializationApi
+import kotlinx.serialization.json.decodeFromStream
 import org.junit.After
 import org.junit.Before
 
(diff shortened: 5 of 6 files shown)


monkopedia-reviewer (changes requested):
# Independent review — PR #111 @ `ee05d18`

I did not write this PR. Nothing below is taken from the description — every claim is
re-measured in my own worktree off the head, through the fleet build slot.

**Verdict: request changes.** The id race is genuinely *closed*, not narrowed, and I could
not break it at any load I threw at it. Two things stop me approving as-is, and neither is
about the `mkdir` primitive:

1. **`Fixes #102` closes an issue whose second documented symptom is still live**, and I
   measured it on this head: a `list()` overlapping a `create()` decodes a half-written
   `info.json` and takes down the **whole** listing. Same exception text as the issue.
2. **The `ID_CLAIM_ATTEMPTS = 32` justification is measurably wrong.** Not the constant —
   the reasoning. Exhaustion is reachable with **4** concurrent creators, not 32.

Details, with numbers, below.

---

## 1. The RED reproduces — surgically, and it is the right RED

I swapped only the claim primitive back to the pre-fix check-then-create shape, leaving
`tryClaimId`'s signature intact so a compile error could not masquerade as a red test:

```kotlin
internal fun tryClaimId(parent: File, id: String): File? {
    parent.mkdirs()
    val dir = File(parent, id)
    if (dir.exists()) return null   // RED: check-then-create
    dir.mkdirs()
    return dir
}
```

`:backend:compileKotlinJvm` **executed**, `:backend:jvmTest` was **reached** and ran
(3 tests), `53 actionable tasks: 53 executed`, `BUILD FAILED in 1m 40s` — not a
two-second configuration failure. From the fresh JUnit XML:

```
ConcurrentCreateTest > concurrentCreatesAllGetDistinctIds FAILED
java.lang.AssertionError: round 0: 8 concurrent creates returned 3 DISTINCT ids
([2, 2, 2, 2, 2, 0, 0, 1]) — duplicates mean two callers hold the same id and one
silently overwrote the other. expected:<8> but was:<3>
```

Five of eight callers walked away holding id `2`. **Zero exceptions.** The other two tests
in the class passed against the broken code, which is the correct shape — only the racing
one can see this.

So the test genuinely races. That answers the instrument-blindness question directly:
this is not a racing test that quietly serialises.

---

## 2. Is the race actually closed? Yes — I tried hard to break it

I wrote my own probe (raw platform threads + a `CyclicBarrier`, a tighter release than the
`CompletableDeferred` gate) against the **unmodified** PR code, asserting not just
distinctness but that the claimed set is exactly `0..N-1` with no gaps:

| racers x rounds | duplicate ids | records lost | outcome |
|---|---|---|---|
| 8 x 25 (the PR's own test) | 0 | 0 | pass |
| **32 x 100** | **0** | **0** | **pass** |
| 64 x 50 | 0 | 0 | fails only on the retry bound (below) |
| 128 x 20 | 0 | 0 | fails only on the retry bound (below) |

And under *sustained* load (N writers each doing K back-to-back creates, no barrier),
1600 creates at the top end:

```
writers=2  each=20 -> ok=40   EXHAUSTED=0   DUPLICATE_IDS=0  dirsOnDisk=40
writers=2  each=50 -> ok=100  EXHAUSTED=0   DUPLICATE_IDS=0  dirsOnDisk=100
writers=4  each=20 -> ok=80   EXHAUSTED=0   DUPLICATE_IDS=0  dirsOnDisk=80
writers=4  each=50 -> ok=198  EXHAUSTED=2   DUPLICATE_IDS=0  dirsOnDisk=198
writers=8  each=20 -> ok=154  EXHAUSTED=6   DUPLICATE_IDS=0  dirsOnDisk=154
writers=8  each=50 -> ok=375  EXHAUSTED=25  DUPLICATE_IDS=0  dirsOnDisk=375
writers=16 each=20 -> ok=281  EXHAUSTED=39  DUPLICATE_IDS=0  dirsOnDisk=281
writers=16 each=50 -> ok=685  EXHAUSTED=115 DUPLICATE_IDS=0  dirsOnDisk=685
writers=32 each=20 -> ok=545  EXHAUSTED=95  DUPLICATE_IDS=0  dirsOnDisk=545
writers=32 each=50 -> ok=1291 EXHAUSTED=309 DUPLICATE_IDS=0  dirsOnDisk=1291
```

**`DUPLICATE_IDS=0` in every row, and `dirsOnDisk` equals successes in every row** — no
duplicate id and no lost record anywhere, at any concurrency I could produce. The
`firstFreeId -> tryClaimId` sequence has no remaining ordering hazard I can find: a lost
claim leaves the winner's directory visible to `usedIds()` (names, not decoded info), so
the candidate is strictly monotonic across retries and the loop converges rather than
spinning. Interleaving with `delete` is also safe, since `deleteRecursively` removes the
directory last and `mkdir` cannot succeed while it exists.

**Conclusion: closed, not narrowed.** Every failure I could force is the *loud* retry
bound, never silent divergence.

---

## 3. Retry-only-for-auto-allocated-ids — correct, in both directions

- Explicit id that loses: `tryClaimId` returns `null` -> `IllegalArgumentException`, no
  retry. Confirmed by `anExplicitlyRequestedIdStillFailsWhenTaken` and by the pre-existing
  `KonstructorImplTest.testCreateDuplicateThrows`, and the loser does not overwrite the
  winner.
- The subtler direction: `buildValue` is invoked with `candidate`, and the function returns
  that same `candidate`; both call sites (`KonstructorImpl.create`,
  `WorkspaceImpl.create`) then rebuild with `newItem.copy(id = claimedId)`. The persisted
  payload and the returned handle carry the id that was actually claimed — a retry can
  never hand back a stale candidate. Nothing an auto-allocating caller could have had
  reason to expect is violated, because it supplied no id.

---

## 4. Directory-names-not-info: no existing install shifts. I checked the real one

Two facts narrow this behaviour change well below what the description implies:

- `firstFreeId` only tests membership for `"0"`, `"1"`, `"2"`… A stray directory only
  consumes an id if its **name is literally a small integer**. Non-numeric strays are free.
- I enumerated the live install on `adolin` (read-only `ls`):

```
~/.konstructor:  0 1 2 3 4 5 6   __lsp_prewarm__  kotlin-lsp-system   (+ lib.jar log.bin log.txt)
```

Every workspace dir `0..6` has an `info.json`. Every konstruction dir inside every one of
those seven workspaces has an `info.json` (46 of them; I listed each). The only non-item
directories are `__lsp_prewarm__` and `kotlin-lsp-system`, both non-numeric, both ignored
by `firstFreeId`.

**Existing installs' ids do not shift. Not the workspace ids, not one konstruction id.**
Next free workspace id is `7` before and after.

I also confirmed `PathController.get()` does create eagerly on a read path
(`File(workspaceDir, id).also { it.mkdirs() }`, reached from `KonstructionControllerImpl`
and `KonstructionLspWorkspace`) — so the case the change is defending against is real, and
under the old code such a directory made its id *permanently unallocatable*. This change
strictly improves that. Worth saying out loud though: the PR mitigates
`PathController`'s mkdir-on-read rather than fixing it, so a mere *lookup* of a
non-existent id still burns that id. Fine for #102; probably worth its own issue.

---

## 5. `ID_CLAIM_ATTEMPTS = 32` — the constant is fine, the stated reasoning is not

The description argues: *"With N concurrent creators each losing at most N-1 times, 32
covers far more concurrency than this service will ever see."*

That bound holds only for a **burst** where each creator makes one item. I confirmed it
does hold there — instrumenting the algorithm, max attempts used was exactly `8` at 8
racers and `28` at 32 racers, never reaching 32.

But under **sustained** load the other creators keep claiming while a slow caller
rescans, so the losses are not bounded by N at all. Measured above: exhaustion appears at
**4 concurrent writers** (2 failures in 200 creates) and reaches 6% at 8 writers. The
`firstFreeId` scan is O(#items) and gets slower as the directory grows, so the caller that
falls behind keeps falling behind.

In practice this is unreachable for Konstructor — a create is a human clicking a dialog,
not four clients in a tight loop — so **I am not asking for the constant to change**, and
`IllegalStateException` is the right degradation: loud, and `callContext` logs every
throwable it sees, so it lands in `log.txt`/`log.bin`. (Note the frontend's
`NavigationDialogViewModel.mutate` swallows it, so the *user* sees the dialog close and
nothing created — that is pre-existing frontend behaviour, not this PR's.)

What I am asking is that the code comment not assert a bound the code does not have.
Either correct the comment, or retry-until-no-progress instead of a fixed count.
**Non-blocking.**

---

## 6. Removed symbols — clean

`git grep writeNewInfo` and `git grep generateId` (real `git grep`, not the shell's
ugrep) return **zero** hits across the whole tree. `WorkspaceImpl.getName`/`setName` still
compile against the surviving `INFO_JSON` constant — both my RED and GREEN runs compiled
`:backend:compileKotlinJvm` and `:backend:compileTestKotlinJvm` cleanly, and the full
suite passes (below).

---

## 7. BLOCKING: `Fixes #102` closes an issue whose symptom 2 is still live

Issue #102 does not only report duplicate ids. It reports, verbatim:

> `list()` reads a **torn** `info.json` — `JsonDecodingException: Expected colon ':', but
> had 'EOF' instead at path: $.konstruction` — because the scan can open a file another
> `create()` is midway through writing. One torn file fails the **whole** listing, not
> just that entry.

and its "Suggested direction" item 2 names the fix (temp file + `Files.move(ATOMIC_MOVE)`).

I measured this **on this head**, with a reader calling `listInfo` while creates run:

```
TORN-READ realistic payload : listings=1167  FAILED_LISTINGS=18   (1.5%)
    sample: JsonDecodingException: Expected start of the object '{', but had 'EOF' instead at path: $
TORN-READ larger payload    : listings=965   FAILED_LISTINGS=148  (15%)
    sample: JsonDecodingException: Expected colon ':', but had 'EOF' instead at path: $.id
```

Same failure mode, same exception class, the second sample is nearly the issue's exact
string. `writeInfo` opens with `File.outputStream()`, which **truncates on open**, so
between that and the write completing a concurrent `listInfo` sees `exists() == true` and
decodes an empty or partial file — and because the decode is inside `mapNotNull` with no
per-entry guard, one torn file throws out of the entire `list()`.

This is pre-existing, not introduced here, and severity is low (the frontend swallows it
and the workspace list momentarily goes empty). I would not block a PR for it on its own.
I am blocking because **this PR closes the issue that reports it**, and because
`writeInfo` is a function this PR creates — it is the natural and cheapest place. It is
also worth noting `listInfo`'s own docstring, kept by this PR, promises "a half-created
... directory must not fail the whole listing"; the code does not hold that promise for a
half-written *file*.

Either resolution is fine by me:

- **(a)** Write it atomically — roughly:
  ```kotlin
  internal inline fun <reified T> writeInfo(targetInfo: File, json: Json, value: T) {
      val tmp = File(targetInfo.parentFile, "${targetInfo.name}.tmp")
      tmp.outputStream().use { json.encodeToStream(value, it) }
      Files.move(tmp.toPath(), targetInfo.toPath(), REPLACE_EXISTING, ATOMIC_MOVE)
  }
  ```
  (and, cheaply, let `listInfo` skip an entry it cannot decode instead of failing the whole
  listing — that also covers a genuinely corrupt file on disk).
- **(b)** Or drop `Fixes #102` to a plain reference, file the remaining symptom as its own
  issue, and let #102 close when both halves land.

I have no preference between them. What I do not want is #102 closing with a measured-live
symptom it documents.

---

## Verification I ran

- `spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true`, run **3 times** on a pristine tree
  (`git status --porcelain` = 0 changes at `ee05d18`), JUnit XML deleted with the
  `find … -exec rm -rf {} +` form before each run and confirmed at 0 files:

  | run | `:backend:jvmTest` / `:protocol:jvmTest` | tests | failures | errors | skipped | wall |
  |---|---|---|---|---|---|---|
  | 1 | executed (not UP-TO-DATE) | 159 | 0 | 0 | 6 | 1m |
  | 2 | executed (not UP-TO-DATE) | 159 | 0 | 0 | 6 | 57s |
  | 3 | executed (not UP-TO-DATE) | 159 | 0 | 0 | 6 | 57s |

  32 fresh XML files each run, none stale. `BUILD SUCCESSFUL` all three, `spotlessCheck`
  included. **The new `ConcurrentCreateTest` did not flake in 3/3 runs** (nor in the
  several extra targeted runs I did while probing).
- CI `build-and-test` on `ee05d18`: **success** (11m54s), confirmed against the head SHA.

---

Everything about the actual fix is right: the primitive, the asymmetry between chosen and
supplied ids, deriving used ids from names, and a test that genuinely races and genuinely
goes red. Close out symptom 2 (or the `Fixes` keyword) and correct the retry-bound comment
and I will approve immediately.

monkopedia-reviewer (changes requested):
# Re-review — PR #111 @ `41680ea`

**The blocking finding is fixed, and I verified the fix rather than taking it.** But the
invariant in the commit title — *"a listing never sees a prefix"* — is still false on this
head, and I measured it with the real class. One more writer of the same file was missed.

**Verdict: request changes** (one blocking, one cheap fix-alongside, plus your two
questions answered).

---

## What I re-verified, and it holds

**Your RED reproduces.** I reverted only `writeInfo`'s body to the in-place write,
signature intact. `:backend:compileKotlinJvm` executed, `:backend:jvmTest` reached,
`53 actionable tasks: 53 executed`, `BUILD FAILED in 1m 37s`:

```
ConcurrentCreateTest > listingNeverSeesAHalfWrittenInfoFile FAILED
java.lang.AssertionError: 39 of 127 concurrent listings saw a torn info.json. An atomic
move leaves no window, so any failure is a regression to in-place writes.
expected:<0> but was:<39>
```

Your test detects the defect and the zero-failures-plus-`listings > 100` shape is right —
a probe that measured nothing cannot pass.

**`writeInfo` is genuinely fixed.** My own probe from last round, re-pointed at this head:

| path | listings | failed |
|---|---|---|
| `writeInfo`, realistic payload | 1326 | **0** |
| `writeInfo`, 2000-char payload | 1571 | **0** |

Against `ee05d18` the same probe gave 18/1167 and 148/965. Zero now, at both payload sizes.

**The id race is still closed after the delta** — 32 racers x 100 rounds, `badRounds=0`.

**Suite on a pristine tree at `41680ea`**, XML deleted first, tasks executed:
2/2 runs, **160 tests, 0 failures, 0 errors, 6 skipped** (matches your number). CI
`build-and-test` on `41680ea`: success (11m11s). `listingNeverSeesAHalfWrittenInfoFile`
did not flake in either run.

---

## BLOCKING: `KonstructionController.info` still writes `info.json` in place — and it is the bigger one

`KonstructionController.kt:91`:

```kotlin
paths.infoFile.outputStream().use { output ->
    config.json.encodeToStream(value, output)
}
```

`paths.infoFile` is `<dataDir>/<ws>/<id>/info.json` — **exactly the file
`WorkspaceImpl.list()` decodes** as `KonstructionInfo` through `listInfo`. The setter fires
from four places in `KonstructionServiceImpl` (rename, dirty-state transitions, target-list
updates after compile), so it rewrites that file far more often than `create` ever writes
it.

Measured on this head using the **real `KonstructionControllerImpl`**, not a replica — a
reader calling `wsDir.listInfo<KonstructionInfo>()` while `controller.info = …` runs:

```
P3B KonstructionController.info pad=0    listings=1250 FAILED=353  (28%)
P3B KonstructionController.info pad=2000 listings=1242 FAILED=435  (35%)
    JsonDecodingException: Expected start of the object '{', but had 'EOF' instead at path: $
```

**28% at a realistic payload** — an order of magnitude worse than the 1.5% you just closed
on the `writeInfo` path.

And the tell that this is the *reported* instance, not a new one I invented: the torn-read
sample in #102 is

> `JsonDecodingException: Expected colon ':', but had 'EOF' instead at path: **$.konstruction**`

`$.konstruction` is a field of `KonstructionInfo`. The file the issue actually caught torn
is the one written by this setter, not by `createWithClaimedId`.

The fix is one line, same package so no import:

```kotlin
runCatching {
    contentFileLock.withLock {
        writeInfo(paths.infoFile, config.json, value)
    }
}.onFailure { hauler.error("Failed to persist info for $workspaceId/$id", it) }
```

I am blocking on the same grounds as last round, and I want to be explicit that I am not
moving the goalposts: it is the *same* symptom, in the *same* file type, read by the *same*
`listInfo` call, and it is what `Fixes #102` will close. The commit message asserts the
invariant globally; right now it holds for two of the three writers.

---

## Fix alongside (2 lines, not worth its own round): the temp name is fixed, and `delete()` is unconditional

```kotlin
val tmp = File(targetInfo.parentFile, "${targetInfo.name}.tmp")
…
} finally { tmp.delete() }
```

One temp path per target means two writers of the *same* `info.json` share it, and the
`finally` deletes it even after a successful move — so writer A's cleanup can unlink
writer B's in-flight temp, and B's `Files.move` then fails.

Measured, 4 threads x 300 writes to one `info.json`:

```
P3C WRITE_ERRORS=628/1200  finalFileDecodes=true  leftoverFiles=[info.json]
    sample: NoSuchFileException: …/0/info.json.tmp -> …/0/info.json
```

The good half: **no corruption** — the target always decodes, and no stale `.tmp` is left
behind. The bad half: **52% of concurrent same-file writes throw.**

Exposure is narrow — `createWithClaimedId` always writes into a directory it exclusively
claimed, so the only reachable case today is two concurrent `WorkspaceImpl.setName` calls
on the *same* workspace, which is the path this commit just routed through `writeInfo`.
Before it, both renames "succeeded" (last writer won); now one can throw
`NoSuchFileException`. Loud beats silent, so this is not urgent — but it is a new failure
mode introduced here, and a unique temp name plus deleting only on failure removes it:

```kotlin
val tmp = File.createTempFile(targetInfo.name, ".tmp", targetInfo.parentFile)
try {
    tmp.outputStream().use { json.encodeToStream(value, it) }
    Files.move(tmp.toPath(), targetInfo.toPath(), ATOMIC_MOVE, REPLACE_EXISTING)
} catch (t: Throwable) {
    tmp.delete()
    throw t
}
```

Worth doing while `KonstructionController` is in your hand. If you disagree, say so — the
"loud failure is fine here" argument you made about `listInfo` applies with some force
here too, and I would accept it.

---

## Your two questions

**1. The `listInfo` skip-undecodable guard — you are right, I withdraw it.**

Not conceding to be agreeable; your reasoning is better than mine. Once every writer is
atomic there is no transient torn state left for the guard to skip, so the only thing it
could ever swallow is a genuinely corrupt file — and silently dropping an item from a
listing is precisely how corruption becomes invisible, which is the entire defect class
this review series exists for. A listing that quietly returns 10 of 11 konstructions is
worse than one that fails loudly. Don't add it.

One caveat that cuts *toward* the blocking item rather than against your call: that
argument only holds *once every writer is atomic*. Today, with
`KonstructionController` still in place, a torn file is transient rather than corrupt —
which is an argument for fixing the writer, not for adding the guard.

**2. The retry-bound comment — correct it on this PR, in the same push as the above.**

You are re-touching the branch anyway, so there is no clean-verification argument for
splitting it. I would not have asked you to push a comment-only commit on top of an
approval (that would stale the review for no behaviour change), but that is moot now.
Either wording works: correct the claim to "a burst where each creator makes one item",
or drop the bound-shaped justification entirely and say the constant is a backstop.

---

## Where this leaves us

Nothing I found is wrong with what you wrote. The atomic move is correct, the sibling-temp
reasoning is correct, routing `setName` through it was the right catch, and your RED is
stronger than mine was. The gap is coverage, not correctness: one writer of the listed file
was missed, and it happens to be the noisiest one and the one the issue actually caught.

Land `KonstructionController` through `writeInfo` (+ the temp-name tidy and the comment)
and I will re-verify against the new SHA and approve. I still hold the claim.

---

*Everything above measured in a worktree off `41680ea`, through the fleet build slot;
probe sources available on request.*

monkopedia-reviewer (approved):
# Approve — PR #111 @ `3c58e16`

All three are fixed, and I verified each one against the real code rather than taking the
description. I said last round that landing these three would get an approval; they landed,
so here it is.

---

## 1. `KonstructionController` — fixed, and the RED is conclusive

I reverted **only** that one line back to the in-place write and re-ran my probe against
the real `KonstructionControllerImpl`:

| | listings | failed |
|---|---|---|
| reverted, realistic payload | 1299 | **451 (35%)** |
| reverted, 2000-char payload | 1012 | 276 (27%) |
| **`3c58e16` as shipped, realistic** | **2225** | **0** |
| **`3c58e16` as shipped, padded** | **2275** | **0** |

And the error the revert produces, at the realistic payload, is this:

```
JsonDecodingException: Expected colon ':', but had 'EOF' instead at path: $.konstruction
```

That is #102's reported string, character for character, `$.konstruction` and all —
reproduced by reverting one line, and gone on your head. That closes the loop on the
argument we were having two rounds ago about which writer the issue actually caught.

## 2. The temp-name regression — gone

4 writers x 300 writes to one `info.json`: **`WRITE_ERRORS=0`**, down from 628/1200. Target
still decodes, and `leftovers=[info.json]` — no temp file left behind, so
`if (!moved) tmp.delete()` is doing the right thing on both paths.

I also checked something the switch to `File.createTempFile` could have changed silently:
`Files.move` replaces the target inode, so the surviving file inherits the *temp* file's
mode. If those differed, every rewritten `info.json` on a live install would quietly change
permissions. Measured — `rw-r--r--` before the rewrite, after the rewrite, and for an
ordinary new file. Identical. No silent mode change.

## 3. The retry-bound comment now states the measured behaviour

Matches my numbers (~1% at 4 concurrent writers, 6% at 8) and correctly scopes the N−1
bound to a single burst.

## Everything else still holds

- **Id race**: 32 racers x 100 rounds, `badRounds=0`. Unchanged by the delta.
- **`writeInfo` torn reads**: 0 in 1501 and 1059 listings.
- **Writer enumeration** — I grepped for every write of any `info.json` in `jvmMain`.
  There are exactly three: `InfoStore.createWithClaimedId` (x2 call sites),
  `WorkspaceImpl.setName`, `KonstructionController.info`. **All three now go through
  `writeInfo`.** That is the complete set, so the invariant in the commit title is now
  actually true.
- **#102 discharges honestly now.** All three symptoms it documents are addressed:
  duplicate ids (round 1), the torn `list()` read (rounds 2 and 3), and
  `PathController`-created directories making an id permanently unallocatable (round 1,
  via `usedIds`). `Fixes #102` is earned.
- **Suite on a pristine tree at `3c58e16`**, XML deleted first, tasks executed, 3 runs:
  **161 tests / 0 failures / 0 errors / 6 skipped** every time. No flake in the three
  concurrency tests across 3 runs plus the several targeted runs I did while probing.
- **CI `build-and-test` on `3c58e16`: success** (11m29s), confirmed against the head SHA.

---

## One thing I am NOT blocking on, filed as #114

The fix in item 1 has no test. I proved that rather than assuming it: with that one line
reverted, **all five `ConcurrentCreateTest` tests pass** while listings are failing at 35%.
The three tests this PR adds all exercise `writeInfo` or `createWithClaimedId` directly;
none goes through the controller.

That is worth closing — it is exactly how this writer got missed in the first place — but I
told you last round I would approve on these three, and moving the line after that would be
its own kind of bad faith. So it is #114, with the probe written out as a ready-to-paste
test and both sets of numbers. **Please do not push it to this branch**; a commit on top
would stale this approval for no behaviour change.

---

## On your notes

**The `listInfo<Space>` detour.** Recording that was the right call and it is the most
useful thing in your message. A run where the RED fires correctly *and* a secondary
assertion is wrong is the hardest state to read, because the failure looks like evidence
for the thing you were already expecting. "Only reading both signals separated them" is the
general lesson, and the fix — decoding the target directly, with the reason written into
the test body so the next person does not re-derive it — is better than just correcting the
call.

**The claim.** No harm done; I re-asserted it and it held. The rule that bit you is worth
narrowing though: release-on-verdict is right for a review that ends, but a
CHANGES_REQUESTED is not an ending — the reviewer is still on the hook for the next head.
Releasing on a *terminal* verdict (approve, or an explicit hand-back) rather than on any
verdict would have avoided it. I am releasing it myself now.

---

Approved. Merging on green.

PR #112, opened 2026-08-22
fix(lsp): readiness means ANSWERED — bound the engine handshake and every engine request (#109)

### LINKED ISSUE #109: An expired engine that stays UP but never answers initialize hangs the editor silently
Found during the #108 review. #84 covered two of the three shapes an unusable engine can take; this is the third, and #108 does **not** fix it.

## The three shapes

| # | shape | covered by |
| --- | --- | --- |
| 1 | binary missing / not executable / env unset | pre-existing `isKotlinLspAvailable` |
| 2 | starts, prints its expiry notice, **exits** | **#108** (readiness probe + bounded respawn) |
| 3 | starts, **stays up**, never answers `initialize` | **nothing** |

Shape 3 is not hypothetical — it is the 2026-06-05 symptom.

## Why #108 does not catch it

The readiness probe asks one question: *is the process still alive after a short window?* Shape 3 answers **yes**. The process is running, the pipes are open, `connection` is set live, and the bridge takes the fully-connected path. Everything downstream then waits on an engine that will never reply.

**And nothing bounds that wait.** `callTimeout` wraps only the `lsp()` sub-service factory, not the delegate calls made through it — so the bridge's `initialize` toward the engine has no timeout, and the frontend's `client.initialize()` hangs behind it.

## Why it is worse than the outage it sits next to

#84's failure was loud: prod fell over, someone noticed within days. **This one is silent.** The editor simply never gets diagnostics or completions. LSP appears to be on, the settings toggle says on, no error surfaces, nothing is logged as failed. Since #103 made LSP default-ON, the population exposed to a silent no-op is now everyone rather than whoever opted in.

## Direction

Make readiness mean *"answered"*, not *"still breathing"*: complete the LSP `initialize` handshake behind a timeout as part of the readiness check, and treat a timeout exactly as #108 treats an immediate exit — report unavailable, degrade to inert, and let the existing respawn bound absorb the retries. That collapses all three shapes onto one definition of usable, instead of adding a third special case.

A timeout on the delegate calls is worth having regardless of how readiness is defined — an engine that stops answering *after* a successful handshake produces the same hang, and no probe at startup can prevent that.

## Test note

The existing fake-engine harness in `KotlinLspProcessExpiredEngineTest` extends to this directly: a script that sleeps instead of exiting reproduces shape 3, and the test should assert the call returns (unavailable) rather than blocking. Assert on **completion within a bound**, not on a flag — the failure mode here is a hang, and a test that waits forever fails by timing out the whole suite rather than by reporting this.

monkopedia-coder:
Fixes #109

## The shape

An unusable kotlin-lsp engine takes three shapes. Two were covered:

| # | shape | covered by |
| --- | --- | --- |
| 1 | binary missing / not executable / env unset | `isKotlinLspAvailable` |
| 2 | starts, prints its expiry notice, **exits** | #108 |
| 3 | starts, **stays up**, never answers `initialize` | **this PR** |

#108's readiness probe asks one question — *is the process still alive after ~1500ms?* — and shape 3 answers **yes**. So `connection` was cached as live, the bridge took the fully-connected path, and everything downstream parked on an engine that would never reply. `callTimeout` wraps only the `lsp()` sub-service factory, not the delegate calls made through it, so the bridge's `initialize` toward the engine had no deadline and the frontend's `client.initialize()` hung behind it.

Worse than the outage next to it because it is **silent**: nothing is logged as failed, no error surfaces, the settings toggle still reads on, and the editor simply never produces diagnostics or completions. Since LSP went default-ON the exposed population is everyone.

## The change

**Readiness means *answered*, not *still breathing*** — and rather than adding a third special case, all three shapes collapse onto one definition of usable.

- `KotlinLspProcess.connect` owns the LSP `initialize` handshake and runs it behind `Config.kotlinLspCallTimeout`. A timeout is treated exactly as #108 treats an immediate exit: log it, destroy the process so it is not cached, return `null`, and let the bounded-respawn logic absorb the retries.
- It therefore returns an `EngineSession` (the stub plus the engine's own `InitializeResult`) instead of a bare stub — holding one is proof the engine *answered*, not merely that it is running. The bridge's "engine unavailable ⇒ inert server" path already existed and is unchanged; `handshakeEngine` splits into `engineParams` (the rewrite the engine needs) and `adoptEngine` (wrap + stand up the pull publisher), both shared by first-initialize and post-crash reconnect exactly as before.
- **Every request-shaped engine call carries the same deadline**, because an engine that stops answering *after* a successful handshake produces the identical hang and no startup probe can prevent it: `guardEngine` / `withLiveDelegate` cover initialize, completion, completionItemResolve, hover and signatureHelp, and `PullDiagnosticsPublisher` covers `textDocument/diagnostic`. Notifications are not covered and need no cover — they complete on the write.
- `Config.kotlinLspCallTimeout` defaults to 30s, sourced from `KONSTRUCTOR_KOTLIN_LSP_TIMEOUT` / `konstructor.kotlinLspTimeout` like the other LSP knobs. A healthy engine answers `initialize` in well under a second (indexing happens afterwards) and a warm pull in ~8s, so only an already-broken engine reaches it.

### Review round 2

**The pull seam was the one unbounded engine request** — and the seam that matters most. `PullDiagnosticsPublisher`'s 4s cold-index retry is a *cadence between completed pulls*, not a deadline: it advances only if a pull returns, so a wedged engine parks `pullOnce` — and, behind its per-doc lock, every later trigger for that doc — for the session. That is this issue's exact symptom at the seam a user notices FIRST, and the one with **no self-heal**, since reconnect only ever fires from a request-shaped call and a user who is only typing makes none. The publisher gains a `pullTimeout` knob beside its existing timing knobs and the bridge passes the same deadline; a pull that hits it is an ordinary failed pull (logged, re-tried on the next cadence), which is strictly better than the silence it replaces. The `Config` KDoc that claimed pulls ran "on their own budget" is corrected rather than left asserting a bound the code lacked.

**The respawn cap could never trip.** A failed attempt costs the 1.5s probe plus the 30s deadline, so under a 3-spawns-per-60s window the first spawn aged out before the third arrived — `recentSpawns` peaked at 2 and the cooldown was unreachable. Replaced with a **consecutive-failure count**, reset only by an answered handshake: time-independent, so no deadline can outrun it, and more accurate anyway (a healthy engine that crashes and respawns cleanly should not count against a cap meant to stop hammering a broken one). It also deletes the deque and the window constant. Both bound tests now assert **equality** with the cap rather than `<=` — "at most the cap" is satisfied by a cap that never trips, which is exactly how a bound ends up measuring nothing.

Also from review: the reconnect test's fake seam now swallows a throwing `initialize` to null as the real `connect` does (that call moved outside any guard, so a propagating fake would pin a contract production does not have).

**Deliberately not changed — worst-case editor latency.** The deadline bounds each engine request, not the whole `withLiveDelegate`, so a call that wedges → reconnects into an engine that also wedges → retries costs up to 3 × the deadline. A single outer bound would have to cancel a `reconnect` mid-flight, which either skips `connect`'s discard of the unresponsive engine (re-creating this very bug) or lets a CancellationException escape an LSP method — the cascade `guardEngine` exists to prevent (#54 / ksrpc#228). The worst case is self-limiting: the wedged engine is discarded on the way through and the cap stops re-spawning it. This is now stated in the `withLiveDelegate` KDoc rather than left for the next reader to discover.

## Tests

The fake-engine harness in `KotlinLspProcessExpiredEngineTest` gains a script that **sleeps instead of exiting** — a real subprocess, real stdio, real ksrpc, answering nothing. The readiness assertion bounds the **whole call** with `withTimeoutOrNull` rather than checking a flag: the defect *is* a hang, so a test that merely waited would fail by timing out the entire suite instead of reporting this. It then proves the hung process is discarded rather than cached, and a second test proves the respawn cap trips for the hanging shape too.

**Demonstrated REDs**, each by disabling only the mechanism under test with its API intact (so every run reached `:backend:jvmTest` rather than failing to compile):

| disabled | failure |
| --- | --- |
| handshake deadline | `connect() did not return within 30s against an engine that answers nothing. Readiness has to mean ANSWERED, not still-breathing…` |
| pull deadline | `a pull that never answers must hit its deadline and be re-tried; the loop only made 1 attempt(s), i.e. it is parked on the first one` |
| respawn cap | `8 attempts must spawn exactly 3 JVMs and then stop expected:<3> but was:<8>` — both engine shapes |

Green, twice, from freshly deleted-and-regenerated XML with the compile tasks executed (not `UP-TO-DATE`): `./gradlew spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true` → **backend 140 tests / 6 skipped / 0 failures, protocol 17 tests / 0 failures** on both runs.

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/Config.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/Config.kt
@@ -19,6 +19,7 @@ import com.monkopedia.konstructor.tasks.LibsJar
 import java.io.File
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.minutes
+import kotlin.time.Duration.Companion.seconds
 import kotlinx.serialization.json.Json
 
 class Config(
@@ -41,6 +42,20 @@ class Config(
         System.getenv("KONSTRUCTOR_KOTLIN_LSP")
             ?: System.getProperty("konstructor.kotlinLsp")
         )?.let(::File),
+    /**
+     * Deadline for a single request made to the kotlin-lsp subprocess — the `initialize`
+     * handshake that decides whether a freshly spawned engine is usable at all, and every
+     * delegated request after it.
+     *
+     * An engine that STAYS UP but never answers is indistinguishable from a healthy one
+     * until something bounds the wait, and the editor simply hangs with LSP apparently on
+     * (#109). A healthy engine answers `initialize` in well under a second — indexing
+     * happens afterwards, and diagnostic pulls are driven by
+     * [com.monkopedia.konstructor.lsp.PullDiagnosticsPublisher] on their own budget — so
+     * this is generous by an order of magnitude and only an engine that is already broken
+     * ever reaches it.
+     */
+    val kotlinLspCallTimeout: Duration = 30.seconds,
     /**
      * Whether the script host advertises STL caching to running scripts
      * ([com.monkopedia.konstructor.lib.HostService.supportsCaching]). Turning it off
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/Config.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/Config.kt
@@ -43,19 +43,32 @@ class Config(
             ?: System.getProperty("konstructor.kotlinLsp")
         )?.let(::File),
     /**
-     * Deadline for a single request made to the kotlin-lsp subprocess — the `initialize`
-     * handshake that decides whether a freshly spawned engine is usable at all, and every
-     * delegated request after it.
+     * Deadline for a single request made to the kotlin-lsp subprocess. It covers EVERY
+     * request-shaped call: the `initialize` handshake that decides whether a freshly spawned
+     * engine is usable at all, the delegated editor requests (completion, hover,
+     * signatureHelp), and the `textDocument/diagnostic` PULL that
+     * [com.monkopedia.konstructor.lsp.PullDiagnosticsPublisher] drives. Notifications are not
+     * covered and need no cover — they complete on the write.
      *
-     * An engine that STAYS UP but never answers is indistinguishable from a healthy one
-     * until something bounds the wait, and the editor simply hangs with LSP apparently on
-     * (#109). A healthy engine answers `initialize` in well under a second — indexing
-     * happens afterwards, and diagnostic pulls are driven by
-     * [com.monkopedia.konstructor.lsp.PullDiagnosticsPublisher] on their own budget — so
-     * this is generous by an order of magnitude and only an engine that is already broken
-     * ever reaches it.
+     * An engine that STAYS UP but never answers is indistinguishable from a healthy one until
+     * something bounds the wait, and the editor simply hangs with LSP apparently on (#109).
+     * The pull seam matters most, because it is the one a user notices first and the one with
+     * no self-heal: the publisher's cold-index retry is a *cadence between completed pulls*,
+     * not a deadline, so an unbounded pull parks it for the session.
+     *
+     * A healthy engine answers `initialize` in well under a second (indexing happens
+     * afterwards) and a warm diagnostic pull in ~8s, so this is generous by design and only an
+     * engine that is already broken reaches it. A pull that does hit the deadline is an
+     * ordinary failed pull: it is logged and re-tried on the next cadence, which is strictly
+     * better than the silence it replaces.
+     *
+     * Sourced (in seconds) from `KONSTRUCTOR_KOTLIN_LSP_TIMEOUT`, or the
+     * `konstructor.kotlinLspTimeout` system property, like the other LSP knobs.
      */
-    val kotlinLspCallTimeout: Duration = 30.seconds,
+    val kotlinLspCallTimeout: Duration = (
+        System.getenv("KONSTRUCTOR_KOTLIN_LSP_TIMEOUT")
+            ?: System.getProperty("konstructor.kotlinLspTimeout")
+        )?.toLongOrNull()?.seconds ?: 30.seconds,
     /**
      * Whether the script host advertises STL caching to running scripts
      * ([com.monkopedia.konstructor.lib.HostService.supportsCaching]). Turning it off
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/lsp/BridgeLanguageServer.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/lsp/BridgeLanguageServer.kt
@@ -66,6 +66,7 @@ import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.job
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
+import kotlinx.coroutines.withTimeout
 import kotlinx.serialization.json.JsonNull
 
 /**
@@ -167,7 +168,7 @@ open class BridgeLanguageServer(
         /**
          * The event-driven pull→push publisher (#43). Non-null ONLY when the engine advertised
          * pull-mode ([ServerCapabilities.diagnosticProvider] != null); otherwise we run no pull
-         * machinery at all. Set by [handshakeEngine], which is where the capabilities land.
+         * machinery at all. Set by [adoptEngine], which is where the capabilities land.
          */
         var diagnostics: PullDiagnosticsPublisher? = null
     }
@@ -268,15 +269,17 @@ open class BridgeLanguageServer(
     }
 
     /**
-     * Obtain a fresh subprocess-facing stub wired to a new [Forwarder] for this session — the sole
-     * seam through which the bridge acquires an engine delegate, used by both the first
-     * [initialize] and the post-crash [reconnect] (#54). [KotlinLspProcess.connect] respawns a dead
-     * subprocess before handing back a stub, which is what makes reconnect self-healing. Returns
-     * `null` when the engine is unavailable. `open` so tests can substitute a fake engine (no real
-     * subprocess) to exercise the crash→respawn transition.
+     * Obtain a fresh [EngineSession] — a subprocess-facing stub wired to a new [Forwarder] for this
+     * session, already through the `initialize` handshake with [params] — the sole seam through
+     * which the bridge acquires an engine delegate, used by both the first [initialize] and the
+     * post-crash [reconnect] (#54). [KotlinLspProcess.connect] respawns a dead subprocess before
+     * handing back a session, which is what makes reconnect self-healing. Returns `null` when the
+     * engine is unavailable, which now includes an engine that stays up but never answers the
+     * handshake (#109). `open` so tests can substitute a fake engine (no real subprocess) to
+     * exercise the crash→respawn transition.
      */
-    protected open suspend fun connectEngine(): KsrpcLanguageServer? =
-        KotlinLspProcess.forConfig(config).connect(Forwarder())
+    protected open suspend fun connectEngine(params: InitializeParams): EngineSession? =
+        KotlinLspProcess.forConfig(config).connect(Forwarder(), params)
 
     override suspend fun initialize(params: InitializeParams): InitializeResult {
         // Stash the editor's params so a post-crash [reconnect] can replay the identical
@@ -285,53 +288,50 @@ open class BridgeLanguageServer(
         // Synthesize the per-konstruction workspace (wrapped .kt + workspace.json) and
         // spawn/connect the warm subprocess. NO blocking client-bound request here.
         lspWorkspace.synthesize()
-        val server = connectEngine()
-        if (server == null) {
+        val session = connectEngine(engineParams(params))
+        if (session == null) {
             hauler.error("kotlin-lsp engine unavailable; bridge is inert")
             return InitializeResult(capabilities = ServerCapabilities())
         }
-        val fresh = EngineConnection(server)
-        connection = fresh
-        return handshakeEngine(fresh, params)
+        connection = adoptEngine(session)
+        return session.initializeResult
     }
 
     /**
-     * Drive a freshly-[connect][KotlinLspProcess.connect]ed [connection] through the engine
-     * handshake and build its pull publisher, returning the engine's [InitializeResult]. Shared by
-     * the first [initialize] and by [reconnect] after a crash, so both replay the IDENTICAL init.
+     * The editor's [InitializeParams] rewritten for the engine. Built by both the first
+     * [initialize] and by [reconnect], so the respawned subprocess is driven through the
+     * IDENTICAL init the first one saw.
+     *
+     * Rooted at the synthesized workspace so the engine loads our workspace.json + classpath.
+     * The editor's LSPClient advertises EMPTY client capabilities; newer kotlin-lsp builds
+     * (LS-262.6274.0+) gate features on the client declaring support — with no
+     * `textDocument.completion` the engine returns no completions, and with no
+     * `textDocument.diagnostic` it omits the pull `diagnosticProvider` entirely (so
+     * [PullDiagnosticsPublisher] never runs). The bridge DOES support both (it forwards
+     * completion and runs the pull→push loop), so we advertise them on the client's behalf
+     * before forwarding.
      */
-    private suspend fun handshakeEngine(
-        connection: EngineConnection,
-        params: InitializeParams
-    ): InitializeResult {
-        // Drive the subprocess through the same handshake, but rooted at the synthesized
-        // workspace so it loads our workspace.json + classpath. The editor's LSPClient
-        // advertises EMPTY client capabilities; newer kotlin-lsp builds (LS-262.6274.0+)
-        // gate features on the client declaring support — with no `textDocument.completion`
-        // the engine returns no completions, and with no `textDocument.diagnostic` it omits
-        // the pull `diagnosticProvider` entirely (so [PullDiagnosticsPublisher] never runs).
-        // The bridge DOES support both (it forwards completion and runs the pull→push loop),
-        // so we advertise them on the client's behalf before forwarding.
-        val result = guardEngine(
-            "initialize",
-            InitializeResult(capabilities = ServerCapabilities())
-        ) {
-            connection.server.initialize(
-                params.copy(
-                    rootUri = lspWorkspace.rootUri,
-                    workspaceFolders = null,
-                    capabilities = params.capabilities.withBridgeFeatures()
-                )
-            )
+    private fun engineParams(params: InitializeParams): InitializeParams = params.copy(
+        rootUri = lspWorkspace.rootUri,
+        workspaceFolders = null,
+        capabilities = params.capabilities.withBridgeFeatures()
+    )
+
+    /**
+     * Take ownership of a freshly-[connect][KotlinLspProcess.connect]ed [session]: wrap its stub in
+     * an [EngineConnection] and build the pull publisher the engine's capabilities call for.
+     * Shared by the first [initialize] and by [reconnect] after a crash.
+     */
+    private fun adoptEngine(session: EngineSession): EngineConnection =
+        EngineConnection(session.server).apply {
+            // Detect pull-mode authoritatively (#43): a non-null diagnosticProvider in the
+            // engine's capabilities is the only signal (there is no push capability flag). Only
+            // then do we stand up the event-driven pull→push publisher; otherwise no pull
+            // machinery runs at all.
+            diagnostics = PullDiagnosticsPublisher
+                .pullProviderOf(session.initializeResult.capabilities)
+                ?.let { provider -> createPublisher(this, provider) }
         }
-        // Detect pull-mode authoritatively (#43): a non-null diagnosticProvider in the
-        // engine's capabilities is the only signal (there is no push capability flag). Only
-        // then do we stand up the event-driven pull→push publisher; otherwise no pull
-        // machinery runs at all.
-        connection.diagnostics = PullDiagnosticsPublisher.pullProviderOf(result.capabilities)
-            ?.let { provider -> createPublisher(connection, provider) }
-        return result
-    }
 
     /**
      * Build the [PullDiagnosticsPublisher] wired to THIS [connection]'s engine stub and the
@@ -472,10 +472,15 @@ open class BridgeLanguageServer(
      * silently stops compiling/executing. So every forward call degrades to inert here: the
      * worst a dead engine can do is "no completion / no hover", never kill the session.
      * Cooperative cancellation is re-thrown; everything else returns [fallback].
+     *
+     * Bounded by [Config.kotlinLspCallTimeout], because an engine that stops answering
+     * mid-session hangs the editor exactly as one that never answered `initialize` does (#109),
+     * and no probe at startup can prevent that. A deadline turns "never returns" into an
+     * ordinary engine failure, which everything above already knows how to degrade.
      */
     private suspend fun <T> guardEngine(label: String, fallback: T, block: suspend () -> T): T =
         try {
-            block()
+            withTimeout(config.kotlinLspCallTimeout) { block() }
         } catch (t: Throwable) {
             // Propagate ONLY a genuine cancellation of *this* coroutine (cooperative cancellation).
             // A FOREIGN CancellationException — the dead kotlin-lsp subprocess leg's MultiChannel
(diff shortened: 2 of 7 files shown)


monkopedia-reviewer (changes requested):
Independent review of `f01917b`. I did not write this; every claim below I confirmed or refuted myself in a throwaway worktree off the PR head.

**Verdict: request changes** — for one thing only, and it is small. The handshake half is correct, well-tested and defensible. The "**and every delegate call**" half of the title is not literally true, and the one call it misses is precisely the one whose hang reproduces #109's *silent* symptom.

---

## What I verified and agree with

**The handshake fix is real.** `KotlinLspProcess.connect` now owns `initialize` behind `withTimeout(config.kotlinLspCallTimeout)`, treats a timeout exactly as #108 treats an immediate exit (log → `shutdownLocked()` → `null`), and hands out `EngineSession` so the type itself carries "this engine answered". Collapsing all three shapes onto one definition rather than adding a third special case is the right call.

**The RED reproduces, fails fast, and names itself.** I disabled *only* the handshake deadline — one line, API intact, so the build reached the test rather than failing to compile:

```
-  withTimeout(config.kotlinLspCallTimeout) { server.initialize(params) }
+  server.initialize(params)
```

```
KotlinLspProcessExpiredEngineTest[jvm] > anEngineThatNeverAnswersInitializeIsTreatedAsUnavailable[jvm] FAILED
    java.lang.AssertionError: connect() did not return within 30s against an engine that
    answers nothing. Readiness has to mean ANSWERED, not still-breathing: a process that
    stays up passes the liveness probe, and the editor then hangs behind an `initialize`
    that never comes back (#109).

3 tests completed, 1 failed
> Task :backend:jvmTest FAILED
BUILD FAILED in 10m 2s
```

Failing case took **30.06s** (the test's own `HANG_BUDGET`), the other two 0.005s each — so it fails on its own clock with its own message, not by starving the suite.

`:backend:compileKotlinJvm` and `:backend:jvmTest` both executed (not `UP-TO-DATE`). The bounding is genuine — `withTimeoutOrNull(HANG_BUDGET)` in the test, so a regression yields a named assertion in ~30s, not a suite timeout. `assertEquals(2, launches())` after `ensureConnection()` correctly proves the corpse was discarded rather than cached. Positive proof the LSP tests really ran: the fake engine's own `This build of intellij-server has expired` appears 4× in the log (1 for the exit-shape test + exactly 3 for the bounded-respawn test — the cap, measured).

**Degradation is genuinely inert.** Traced end-to-end: after a handshake timeout `connection` stays `null`; `initialize` returns empty `ServerCapabilities`; every editor-facing override (`didOpen`, `didChange`, `didClose`, `completion`, `completionItemResolve`, `hover`, `signatureHelp`) short-circuits on `connection == null` *before* any logging, so there is no per-keystroke spew and no retry loop inside the session. Exactly one error line, at connect time. The frontend's `client.initialize()` runs inside `produceState`, so the 30s worst case delays the LSP extension rather than blocking the editor.

**`Config.kt` is clean.** `Config` is a plain class (not `data`, not `@Serializable`), so no equality/hashCode change — which matters, because `KotlinLspProcess.instances` keys a map on it. The new param is inserted mid-constructor, but every one of the 8 call sites in the tree passes everything past position 1 by name, so nothing breaks. Default 30s, opt-out-shaped.

---

## Blocking: the pull-diagnostics call is still unbounded

`@KsNotification` matters here, so the enumeration has to distinguish the two kinds. I checked the annotations in `lsp-ksrpc` 1.2.0 and the writer in `ksrpc-jsonrpc` (`JsonRpcWriterBase.execute`: `isNotify` → `comm.send(...)`, `return null`): `initialized`, `didOpen`, `didChange`, `didClose` are notifications — the call awaits the *write*, never a reply. So the deadline on `didOpen`/`didChange` only guards pipe backpressure, and its absence on `initialized`/`didClose` is a correspondingly small residual (a wedged engine that stops draining stdin could park them once the 64K pipe buffer fills; both sit under `runCatching`, and one of them holds `reconnectLock`). The calls that can park forever on their own are the **request**-shaped ones:

| engine-facing call | kind | site | bounded? |
| --- | --- | --- | --- |
| `initialize` | request | `KotlinLspProcess.connect` | ✅ `withTimeout` |
| `textDocument/completion` | request | `withLiveDelegate` | ✅ |
| `completionItem/resolve` | request | `withLiveDelegate` | ✅ |
| `textDocument/hover` | request | `withLiveDelegate` | ✅ |
| `textDocument/signatureHelp` | request | `withLiveDelegate` | ✅ |
| **`textDocument/diagnostic`** | **request** | **`createPublisher`'s `pull` seam, `BridgeLanguageServer.kt:352`** | ❌ **none** |
| `initialized` ×2, `didOpen` ×2, `didChange`, `didClose` ×2 | notification | various | n/a (fire-and-forget) |

The pull seam calls `connection.server.textDocumentDiagnostic(...)` directly — it goes through neither `guardEngine` nor `withLiveDelegate`. `git grep withTimeout` over `backend/.../lsp/` returns three hits, none in `PullDiagnosticsPublisher`; there is no `@KsTimeout` anywhere in lsp-kotlin either (`git grep KsTimeout` in that repo: empty). So:

> `Config.kt`: "diagnostic pulls are driven by `PullDiagnosticsPublisher` on their own budget"

is not accurate. The publisher has a **retry cadence** (`coldIndexRetry = 4s`), not a **deadline** — and the cadence only advances *between completed pulls*. If the engine wedges on a pull, `pullOnce` never returns, `pullUntilPublished` parks forever, and you get: no diagnostics, nothing logged as failed, LSP toggle still on — the exact #109 symptom, at the exact seam whose failure the user notices first. Worse, nothing self-heals it: `withLiveDelegate`'s reconnect only fires from `completion`/`hover`/`signatureHelp`, and `didOpen`/`didChange` cannot fire it because notifications don't fail. A user who just types sits there silently.

By the PR's own standard — *"it is four lines here, so it is in scope rather than deferred"* — this belongs in the PR. A `withTimeout(config.kotlinLspCallTimeout)` inside the `pull` lambda degrades correctly for free: it throws, `pullOnce`'s existing `catch` calls `coroutineContext.ensureActive()` (the outer coroutine is untouched by an inner `withTimeout`), logs `pull diagnostics failed`, returns `FAILED`, and the 4s retry loop carries on.

And it is safe against cold start: the publisher's own KDoc records that during the ~120s cold index a pull *"may transiently fail or come back empty"* — i.e. the engine answers pulls promptly even when cold; that is why the 4s retry exists at all. A deadline there does not decapitate a healthy-but-cold engine.

Your own fixture already demonstrates the gap, incidentally: `FakeEngine.textDocumentDiagnostic` in `BridgeLanguageServerReconnectTest` parks on a `CompletableDeferred` that is never completed, and the #80 test asserts that the *only* thing which ever ends that pull is `dead.scope.cancel()`. Nothing in the pull path has a countdown.

One caveat on scope, so I'm not over-prescribing: a bare deadline makes the wedge **visible and bounded** (error logged, `FAILED`, retry in 4s) but it does not **self-heal** — the pull path has no route into `withLiveDelegate`'s single-flight reconnect, so recovery would still wait for the user to hover. So: **either** bound the pull seam (visible beats silent, and it is the same four lines), **or**, if the full answer is bigger than this PR wants to be, correct the `Config` KDoc + the "every delegate call" title and file the follow-up. Either one closes this for me — what I can't sign off is the claim standing while the seam is open.

---

## Non-blocking findings

**1. The 3-spawns/60s cooldown cannot trip for this shape.** `SPAWN_WINDOW_MILLIS = 60_000` but a failed handshake now costs `READY_PROBE_MILLIS` + 30s ≈ 31.5s, so consecutive attempts are ~31.5s apart and the window prunes the oldest entry before the third ever lands: spawns at t=0, 31.5, 63 → at t=63 the t=0 entry is 63s old and dropped, `recentSpawns.size` never reaches 3. So "let the existing bounded-respawn logic absorb the retries" holds for the *discard-and-respawn* half but not for the *cap* half — the 5-minute cooldown is unreachable on the hang path. The cap needs the 3rd spawn inside the window, i.e. `2 × (READY_PROBE + deadline) < 60s` → deadline < ~28s; 30s misses it by a hair. This is nothing like #84 (attempts are gated on bridge init, not on cursor moves, and each corpse is destroyed so nothing accumulates, so it is ~1 JVM per 31.5s of continuous retrying, not one per mouse move), so I am not blocking on it — but it wants either a note in the KDoc or a slightly smaller default.

**2. Worst-case editor-facing latency is now ~90s, not 30s.** The deadline bounds each leg rather than the whole `withLiveDelegate`: 30s (first attempt) + up to 31.5s (`reconnect` → `connect`, which re-runs the bounded handshake) + 30s (`guardEngine` post-reconnect retry). Still a strict improvement over unbounded, but if you want "a hover can't take more than X", the timeout wants to be on the outside.

**3. `kotlinLspCallTimeout` has no env/system-property source,** unlike `kotlinLspBinary` (`KONSTRUCTOR_KOTLIN_LSP`) and `kotlinLspSystemPath` (`KONSTRUCTOR_KOTLIN_LSP_SYSTEM_PATH`). Ops cannot widen it on a slow box without a rebuild. Cheap to add, matches the file's own convention.

**4. Small test-fidelity nit.** `FakeEngineBridge.connectEngine` says it "Mirrors `KotlinLspProcess.connect`", but the real one swallows a throwing `initialize` into `null` (`catch (t: Throwable)`) while the fake propagates it. That matters because the PR moved the engine `initialize` from *inside* `guardEngine` to *outside* any guard in `BridgeLanguageServer.initialize` — it is safe today only because `connect` catches internally, and the fake seam would not catch a regression there.

---

## Measurements

Own worktree off `f01917b`; XML deleted with `find … -name '*.xml' -exec rm -rf {} +` before every run; counts read from the freshly regenerated XML; `:backend:compileKotlinJvm` + `:backend:jvmTest` confirmed executed (not `UP-TO-DATE`) on the runs that had work to do, and 29 fresh backend XML files after each deletion.

`./gradlew spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true`, **3/3 green**:

| run | result | backend | protocol | backend suite | `KotlinLspProcessExpiredEngineTest` |
| --- | --- | --- | --- | --- | --- |
| 1 | BUILD SUCCESSFUL 12m22s (75/75 executed) | 138 tests / 6 skipped / 0 fail | 17 / 0 fail | 59.1s | 4.04s |
| 2 | BUILD SUCCESSFUL 10m06s | 138 / 6 / 0 | 17 / 0 | 60.5s | 4.27s |
| 3 | BUILD SUCCESSFUL 10m42s | 138 / 6 / 0 | 17 / 0 | 59.2s | 4.04s |

No flake across three runs of a test that spawns real subprocesses — tight spread (4.03/4.26/4.03s).

**Added CI cost: ~4.0s** — `anEngineThatNeverAnswersInitializeIsTreatedAsUnavailable` is 4.03s of a ~60s backend suite (1.5s liveness probe + 1s `ENGINE_DEADLINE` + 1.5s probe on the respawn). That is honest and cheap; nothing here quietly adds tens of seconds. I also checked for leaked fakes: `ps` shows **0** stray `sleep 600` after both the green and the RED runs, so the `destroy()`/`destroyForcibly()` path does reap the fake engine's process group.

**CI on the PR head:** `build-and-test` **pass** (20m21s). No other reviews on the PR.

Environment: own worktree off `f01917b`, JDK 21, all Gradle through the fleet build slot.

monkopedia-reviewer (approved):
## Re-review of `d66f39f` — approve

Independent re-measurement of the new head (the prior CHANGES_REQUESTED pinned `f01917b`, so nothing had a verdict here). Own worktree off `d66f39f`; every Gradle invocation through the fleet build slot. Both blocking items from the previous round are genuinely resolved, and I reproduced the evidence rather than reading it.

### 1. The pull seam is bounded by a deadline, not a cadence — RED reproduced

`PullDiagnosticsPublisher.pullOnce` now wraps the pull itself:

```kotlin
withTimeout(pullTimeout) { pull(uri, previousResultId) }
```

with the bridge passing `pullTimeout = config.kotlinLspCallTimeout` at the `textDocument/diagnostic` seam. The `catch` around it is `coroutineContext.ensureActive()`, not `catch (CancellationException)`, so the `TimeoutCancellationException` is handled as an ordinary failed pull (the outer job is still active) and the retry loop keeps going — which is the behaviour the test asserts.

Surgically removed only the deadline (`withTimeout(pullTimeout) { pull(...) }` → `pull(...)`, `pullTimeout` parameter and every call signature left intact) and ran the class:

```
> Task :backend:compileTestKotlinJvm
> Task :backend:jvmTest FAILED
PullDiagnosticsPublisherTest[jvm] > a pull the engine never answers is bounded and re-tried, not parked[jvm] FAILED
    java.lang.AssertionError: a pull that never answers must hit its deadline and be re-tried;
    the loop only made 1 attempt(s), i.e. it is parked on the first one
11 tests completed, 1 failed
```

Reached `:backend:jvmTest` and ran 11 tests — a real red, not a compile failure. Restored, green again.

### 2. The respawn cap actually trips — measured, on a shape the PR does not test

The author's two cap tests assert **equality** with `MAX_CONSECUTIVE_FAILURES`, which is what makes them load-bearing: an unreachable cap cannot satisfy `assertEquals(3, launches())` after 8 attempts. Both pass.

To avoid taking that on the same evidence the author used, I wrote a throwaway test for a **fourth** engine shape neither test covers — one that survives the 1500 ms liveness probe and then exits *during* the handshake (`sleep 2; exit 1`), with a 10 s deadline — and counted real JVM launches:

```
REVIEWPROBE attempt=0 session=null launches=1
REVIEWPROBE attempt=1 session=null launches=2
REVIEWPROBE attempt=2 session=null launches=3
REVIEWPROBE attempt=3..7 session=null launches=3
```

8 attempts → exactly 3 spawns, then the cooldown. 34.5 s for the three attempts = 3 × (1.5 s probe + 10 s deadline), i.e. a mid-handshake death surfaces as a *timeout*, not a throw, so it lands in the branch that increments the counter. The cap is reachable, and time-independent as claimed.

### 3. Delegate-call inventory, re-enumerated from the head

13 engine call sites, 9 distinct methods (`grep` over `lsp/*.kt`, cross-checked against `@KsNotification` in `lsp-jvm-1.2.0.jar` via `javap -v`):

| call | site | bound |
| --- | --- | --- |
| `initialize` | `KotlinLspProcess.kt:108` | `withTimeout(kotlinLspCallTimeout)` in `connect` |
| `textDocument/diagnostic` | `Bridge:352` | `pullTimeout` = same deadline |
| `textDocument/completion` | `Bridge:649` | `withLiveDelegate` |
| `completionItem/resolve` | `Bridge:667` | `withLiveDelegate` |
| `textDocument/hover` | `Bridge:684` | `withLiveDelegate` |
| `textDocument/signatureHelp` | `Bridge:720` | `withLiveDelegate` |
| `initialized` ×2, `didOpen` ×2, `didChange`, `didClose` ×2 | 381/595, 401/617, 434, 462/792 | `@KsNotification` — complete on the write |

The prior round's list of request-shaped calls was complete and still is. The bridge never sends `shutdown`/`exit` to the engine (deliberate — shared keep-warm process), so those requests do not exist here. Cancelling a request mid-flight is safe on this transport: `JsonRpcWriterBase` matches responses through an id-keyed `MultiChannel` and implements `CancellationSupport`/`sendCancel`, so a timed-out call cannot desync the channel.

### 4. Cold start

The deadline is **unchanged at 30 s** — the author made the cap reachable by replacing the window with a failure count rather than by lowering the deadline, so nothing crossed into cold-index territory. `initialize` precedes indexing; the ~120 s cold index is a post-`initialized` cost.

Two consequences did change, both worth an ops line rather than a block:

- The cap is now reachable **by slowness as well as by brokenness**. Three consecutive starts where `initialize` exceeds 30 s now arm a 5-minute cooldown (LSP off) where previously they just retried forever.
- The cold *pull* is now bounded too. The premise — a cold pull returns empty/failed rather than blocking, which is precisely why the 4 s cold-index cadence exists — is consistent with the publisher's design. If it is ever wrong, the failure mode is "timeout, log, retry every ~34 s until the index warms": noisier, still converging.

I could not measure real cold-start latency: the engine binary is not on this host (it lives on adolin).

### 5. Prose vs. code

`Config.kt`'s KDoc no longer claims pulls run "on their own budget" and now describes what the code does; the PR title's "every engine request" is literally true against the inventory above. Two small mismatches remain, neither blocking:

- **`withLiveDelegate` KDoc, `BridgeLanguageServer.kt:516`** — "costs up to 3 × `Config.kotlinLspCallTimeout`". The actual chain is first attempt (30 s) + `reconnect` [1.5 s probe + 30 s handshake + `guardEngine("reconnect didOpen")` 30 s] + post-reconnect retry (30 s) — up to **4 ×** the deadline plus the probe, and `KotlinLspProcess.connect` holds a process-wide mutex, so a concurrent konstruction's handshake can queue another ~31.5 s on top. 3× is the realistic figure (the replayed `didOpen` is a notification that returns on the write); the stated bound is not the strict one.
- **`KotlinLspProcessExpiredEngineTest.kt:167`** still says "(bounded by `MAX_SPAWNS_PER_WINDOW`)" — a constant this commit deletes. `git grep` confirms it is the only surviving reference.

### 6. Is there a fifth rung? Yes — two, both narrower than what was fixed

**The bound got narrower in one direction.** The old window counted *spawns*; the new counter counts *consecutive failures, reset by an answered handshake*. So an engine that answers `initialize` and then dies — repeatedly — resets the counter every cycle and is respawned without limit, paced by request traffic. That shape *was* bounded before (3 spawns/60 s regardless of outcome) and is not now. It is materially less severe than #84 (each cycle costs ≥1.5 s of probe, and the corpse is destroyed before the next spawn, so it is ~40 JVM launches/min rather than mouse-move rate) and the KDoc argues the trade deliberately — but the honest statement is that the cap now catches "never becomes usable" and no longer catches "usable, then dead, forever", not that it is strictly better.

**The pull seam is bounded but still not self-healing.** A wedged engine now produces a logged, retried failure instead of a silent park — a real improvement, and the loop recovers the moment the engine answers again — but nothing at that seam triggers a reconnect, and structurally cannot: the publisher runs inside the `EngineConnection.scope` that `reconnect` cancels. So a user who is only typing still sees "LSP on, no diagnostics" indefinitely, just no longer silently. The PR says as much; recording it as the answer to "can the engine be unusable while the code believes it is fine" — the code no longer believes it is fine, but the user's symptom is unchanged until they hover.

Minor, same area: in `connect`, the generic `catch (Throwable)` (a handshake that *throws* rather than times out) neither increments `consecutiveFailures` nor calls `shutdownLocked()`. Measured above: a dying process times out rather than throwing, so the realistic path is covered; the throwing path is reachable mainly by a protocol/serialization error, where the still-alive process is kept and the same broken channel is retried. And `kotlinLspCallTimeout` silently accepts `0` (→ `withTimeout(0)` → LSP permanently off) and silently ignores a malformed env value.

### 7. Verification — 3 clean runs, and the added CI cost

`./gradlew spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true`, XML deleted with the `find … -exec rm -rf {} +` form before each run, `:backend:jvmTest` executed (never `UP-TO-DATE`), counts read from the freshly regenerated XML:

| run | result | backend | protocol | test-exec time | wall |
| --- | --- | --- | --- | --- | --- |
| 1 | BUILD SUCCESSFUL | 140 tests / 0 fail / 6 skip | 17 / 0 fail | 67.1 s | 651 s |
| 2 | BUILD SUCCESSFUL | 140 / 0 / 6 | 17 / 0 | 67.3 s | 651 s |
| 3 | BUILD SUCCESSFUL | 140 / 0 / 6 | 17 / 0 | 67.0 s | 650 s |

(An earlier fourth run carried my throwaway probe test — 141 tests, also green.) Wall clock is dominated by the frontend/wasm chain that `spotlessCheck` pulls in, not by the tests.

**Added CI cost:** `KotlinLspProcessExpiredEngineTest` goes from 2 tests at ~0.01 s to 4 tests at **11.56 s** — `theRespawnCapTripsForAnEngineThatNeverAnswers` 7.54 s (new here) and `anEngineThatNeverAnswersInitializeIsTreatedAsUnavailable` 4.01 s (from `f01917b`). That is real subprocess time (3 × probe + deadline, then the cooldown) and it is +21 % on backend jvmTest execution (11.5 s added to a ~55.6 s suite, now 67.0-67.3 s across three runs) and ~0.9 % of the 21m34s CI job. The new pull test costs ~0 — it runs on `runTest`'s virtual clock, which is the right call and worth keeping in mind if the engine-shape tests ever need trimming.

CI: `build-and-test` **success** on `d66f39f` (run 32556003065, 21m34s).

### Verdict

Approve. The pull seam is bounded by a deadline and I reproduced its red; the cap trips, measured on a shape the PR does not itself test; the delegate inventory is complete and every request-shaped call is covered; the deadline was not lowered, so cold start is not squeezed. The remaining items — the 3×/4× arithmetic in the `withLiveDelegate` KDoc, the stale `MAX_SPAWNS_PER_WINDOW` comment, and the crash-after-handshake loop the new counter no longer bounds — are follow-up material, not merge blockers.

PR #118, opened 2026-08-22
test(backend): cover the info.json writer the suite was blind to

### LINKED ISSUE #114: Regression test missing for KonstructionController's atomic info.json write (#102 symptom the suite cannot see)
Found during the independent review of #111 (head `3c58e16`). Not a defect in that PR — it
landed correct and verified — but the fix it contains has **no test**, and I measured that
the suite is blind to its regression.

## What is untested

`KonstructionController.info`'s setter writes the konstruction's own `info.json`
(`KonstructionController.kt:95`). #111 routed it through `writeInfo` so it is atomic,
because that file is exactly what `WorkspaceImpl.list()` decodes as `KonstructionInfo`
through `listInfo`, and the setter fires on rename, dirty-state changes and target updates.

The three tests #111 added all exercise `writeInfo` (or `createWithClaimedId`) directly.
None goes through the controller.

## Evidence the suite cannot see a regression here

I reverted **only** that one line back to the in-place write, on `3c58e16`, and ran the
whole `ConcurrentCreateTest` class:

```
ConcurrentCreateTest  tests=5  failures=0  errors=0
    pass concurrentWritersToOneFileDoNotBreakEachOther
    pass anExplicitlyRequestedIdStillFailsWhenTaken
    pass concurrentCreatesAllGetDistinctIds
    pass listingNeverSeesAHalfWrittenInfoFile
    pass anIdWhoseDirectoryExistsWithoutInfoIsNotReoffered
```

**Five of five green** — while a reviewer probe against the same build measured
**451 of 1299 concurrent listings failing (35%)**, with the error being #102's verbatim
string:

```
JsonDecodingException: Expected colon ':', but had 'EOF' instead at path: $.konstruction
```

So the exact defect #102 reported can be reintroduced by a one-line edit and the suite
stays green. That is worth closing, because this is precisely how the writer got missed in
the first place: #111's first two rounds fixed two writers of `info.json` and left this one
in place, and nothing failed.

## Suggested test

This is the reviewer probe that caught it, trimmed to an assertion. It uses the real
`KonstructionControllerImpl` with `Dispatchers.Unconfined` so the write happens inline.

```kotlin
@Test
fun aListingNeverSeesTheControllersInfoFileHalfWritten() {
    val root = kotlin.io.path.createTempDirectory("controller-torn-").toFile()
    val config = Config(root)
    val wsDir = File(root, "ws1").also { it.mkdirs() }
    writeInfo(File(wsDir, INFO_JSON), config.json, Space(id = "ws1", name = "ws"))
    val kDir = File(wsDir, "k1").also { it.mkdirs() }
    val base = KonstructionInfo(
        Konstruction(name = "k", workspaceId = "ws1", id = "k1"), CLEAN, emptyList()
    )
    writeInfo(File(kDir, INFO_JSON), config.json, base)

    val controller = KonstructionControllerImpl(config, "ws1", "k1", Dispatchers.Unconfined)
    val pool = Executors.newFixedThreadPool(1)
    val listings = AtomicInteger()
    val failures = AtomicInteger()
    val stop = AtomicBoolean(false)
    val reader = pool.submit {
        while (!stop.get()) {
            listings.incrementAndGet()
            runCatching { wsDir.listInfo<KonstructionInfo>(config.json) }
                .onFailure { failures.incrementAndGet() }
        }
    }
    val padding = "n".repeat(2_000)
    repeat(800) { n ->
        controller.info = base.copy(konstruction = base.konstruction.copy(name = "$padding-$n"))
    }
    stop.set(true)
    reader.get(30, TimeUnit.SECONDS)
    pool.shutdownNow()

    assertEquals(
        0,
        failures.get(),
        "${failures.get()} of ${listings.get()} listings saw a torn info.json — the " +
            "controller's write is no longer atomic (konstructor#102)."
    )
    assertTrue(listings.get() > 100, "the probe only proves something if it listed a lot")
    root.deleteRecursively()
}
```

Measured on `3c58e16` as shipped: **0 failures in 2225 listings** (and 0 in 2275 with the
2000-char payload). With the line reverted: 451/1299 and 276/1012. It separates the two
states cleanly, and the `listings > 100` guard stops it passing trivially if it ever stops
racing.

monkopedia-coder:
Fixes #114

## The gap

PR #111 routed `KonstructionController.info`'s setter through `writeInfo` — that setter writes the konstruction's own `info.json`, which is exactly what `WorkspaceImpl.list()` decodes, and it fires on rename, dirty-state changes and target updates. It is the writer issue #102 actually reported, identifiable from the reported error naming `at path: $.konstruction`.

**It shipped with no test, and the review measured that the suite cannot see its regression:** revert that one line and all five `ConcurrentCreateTest` cases stay green while 451 of 1299 concurrent listings fail (35%).

That blindness is not incidental — **it is how the writer was missed for two rounds.** #111's earlier passes fixed `createWithClaimedId` and `WorkspaceImpl.setName`, both exercised directly by tests, and nothing failed. The one writer nothing drove was the one the issue was about.

## What the test does

Drives the **real `KonstructionControllerImpl`** on `Dispatchers.Unconfined`, so the setter's write happens inline — otherwise the reader loop outruns the writes and the probe races nothing at all. A background reader lists the workspace continuously while 800 padded writes go through the controller.

It asserts **zero** torn listings rather than a reduced rate, because an atomic move leaves no window whatsoever, and it asserts the reader got past a hundred listings so a probe that measured nothing cannot pass trivially.

## Demonstrated RED

Reverting only the controller's write back to the in-place form, API and every other test intact:

```
aListingNeverSeesTheControllersInfoFileHalfWritten FAILED
552 of 1466 listings saw a torn info.json — the controller's write is no longer
atomic (konstructor#102). expected:<0> but was:<552>
```

The other five cases in that class stayed green throughout, which is the point.

## Verification

`spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true` — **162 tests, 0 failures, 6 skipped**, twice before the rebase and once after, XML deleted first with fresh mtimes confirmed.

## Notes for the reviewer

- Test-only; no production change. The behaviour it guards merged in #111 (`56ff9c0`).
- Post your verdict as a **review object** (`gh pr review …`) via the reviewer bot — even a held outcome. A verdict as a plain comment pins to no SHA.
- `main` is unprotected, so `--auto` is not a CI gate — poll `gh pr checks` to completed SUCCESS, then plain `--squash --delete-branch`.
- Worth challenging: `Dispatchers.Unconfined` makes the write inline and therefore deterministic, but it also means this does not exercise the real `newSingleThreadContext` save path. If you think a version that races the actual async setter would catch more, say so — I chose determinism over fidelity here and that is arguable.

### DIFF
--- a/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/ConcurrentCreateTest.kt
+++ b/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/ConcurrentCreateTest.kt
@@ -17,8 +17,15 @@
 
 package com.monkopedia.konstructor
 
+import com.monkopedia.konstructor.common.DirtyState
+import com.monkopedia.konstructor.common.Konstruction
+import com.monkopedia.konstructor.common.KonstructionInfo
 import com.monkopedia.konstructor.common.Space
 import java.io.File
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicInteger
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertFailsWith
@@ -229,12 +236,69 @@ class ConcurrentCreateTest {
         )
     }
 
+    @Test
+    fun aListingNeverSeesTheControllersInfoFileHalfWritten() {
+        // The writer that #102 actually reported, and the one the rest of this class is
+        // BLIND to. Measured during the #111 review: revert KonstructionController's write
+        // to the in-place form and all five other tests here still pass, while 451 of 1299
+        // concurrent listings fail with #102's verbatim string —
+        // `Expected colon ':', but had 'EOF' instead at path: $konstruction`. The other
+        // tests exercise writeInfo and createWithClaimedId directly; none goes through the
+        // controller, which is exactly how this writer was missed for two rounds.
+        val root = env.tempDir
+        val wsDir = File(root, "ws1").also { it.mkdirs() }
+        writeInfo(File(wsDir, INFO_JSON), env.config.json, Space(id = "ws1", name = "ws"))
+        val kDir = File(wsDir, "k1").also { it.mkdirs() }
+        val base = KonstructionInfo(
+            Konstruction(name = "k", workspaceId = "ws1", id = "k1"),
+            DirtyState.CLEAN,
+            emptyList()
+        )
+        writeInfo(File(kDir, INFO_JSON), env.config.json, base)
+
+        // Unconfined so the setter's write happens inline rather than on its own thread —
+        // otherwise the loop below outruns the writes and the probe races nothing.
+        val controller =
+            KonstructionControllerImpl(env.config, "ws1", "k1", Dispatchers.Unconfined)
+        val listings = AtomicInteger()
+        val failures = AtomicInteger()
+        val stop = AtomicBoolean(false)
+        val pool = Executors.newFixedThreadPool(1)
+        try {
+            pool.submit {
+                while (!stop.get()) {
+                    listings.incrementAndGet()
+                    runCatching { wsDir.listInfo<KonstructionInfo>(env.config.json) }
+                        .onFailure { failures.incrementAndGet() }
+                }
+            }
+            val padding = "n".repeat(2_000)
+            repeat(CONTROLLER_WRITES) { n ->
+                controller.info =
+                    base.copy(konstruction = base.konstruction.copy(name = "$padding-$n"))
+            }
+        } finally {
+            stop.set(true)
+            pool.shutdown()
+            pool.awaitTermination(30, TimeUnit.SECONDS)
+        }
+
+        assertEquals(
+            0,
+            failures.get(),
+            "${failures.get()} of ${listings.get()} listings saw a torn info.json — the " +
+                "controller's write is no longer atomic (konstructor#102)."
+        )
+        assertTrue(listings.get() > 100, "the probe only proves something if it listed a lot")
+    }
+
     private companion object {
         const val RACERS = 8
         const val ROUNDS = 25
         const val REWRITES = 400
         const val WRITERS = 4
         const val WRITES_EACH = 300
+        const val CONTROLLER_WRITES = 800
 
         /** Big enough that a truncated write is very likely to land mid-token. */
         val PADDING = "n".repeat(2_000)


monkopedia-reviewer (approved):
## Approve — the test is sensitive to the real regression, and the guard holds up under contention

Independent review at `746d654`, on a fresh worktree off the PR head. I reproduced the RED
myself rather than taking the PR body's numbers, and I measured the two things the PR itself
flags as arguable: the `listings > 100` guard, and the `Dispatchers.Unconfined` tradeoff.

### 1. RED reproduced

Reverted **only** the controller's write back to the pre-#111 in-place form (the exact hunk
`56ff9c0` replaced), API and all other tests intact:

```kotlin
paths.infoFile.outputStream().use { output ->
    config.json.encodeToStream(value, output)
}
```

```
ConcurrentCreateTest[jvm] > aListingNeverSeesTheControllersInfoFileHalfWritten[jvm] FAILED
    java.lang.AssertionError: 233 of 968 listings saw a torn info.json — the controller's
    write is no longer atomic (konstructor#102). expected:<0> but was:<233>
6 tests completed, 1 failed
```

24% of listings torn; **the other five cases in the class stayed green**, which is the whole
point of the PR. A second, independent RED later in the session measured 470 of 1398 (34%).
`:backend:jvmMainClasses`/`:backend:jvmJar` re-executed (not `UP-TO-DATE`) and
`:backend:jvmTest` actually ran — "6 tests completed, 1 failed", XML mtime matching the run's
own end timestamp — so this is a red test, not a compile error wearing one's clothes.

Restoring the atomic write goes green again.

### 2. The `listings > 100` guard is robust — measured, not assumed

I instrumented the test to print the counter and ran it across four contention regimes,
including one that models the 2-vCPU `ubuntu-latest` runner:

| regime | listings (3 runs) |
|---|---|
| 16 cores, normal | 2316 / 2236 / 2685 |
| `taskset -c 0,1` (2 cores) | 2231 / 2208 / 2438 |
| 2 cores + 4 competing spinners (3× oversubscribed) | 2318 / 1983 |
| `taskset -c 3`, whole toolchain on **one** core | 2021 / 2452 / 2065 |

Guard is 100, so the margin is **20–27×, and it does not degrade under load.** That is not
luck: the reader's window is bounded by how long the writer's 800 writes take, so contention
slows both sides together and the ratio is roughly invariant. Starving the reader below 100
would require starving the writer with it, which shortens the window it needs to survive. I
don't think this is a flake risk on CI.

### 3. `Dispatchers.Unconfined` — I think the tradeoff is right, and I tested the worry

The concern worth having is not fidelity for its own sake, it is whether the async path has a
tearing mode this test structurally cannot see. I don't believe it does:

- The tear is produced by `outputStream()` truncating on open, and observed by a *different*
  thread decoding the file. Which thread performs the write does not change the window.
- The production path serialises every setter write onto one `newSingleThreadContext` thread
  *and* behind `contentFileLock`, so it introduces no writer/writer interleaving that inline
  execution omits.
- Injecting the dispatcher is already this codebase's convention for controller tests —
  `KonstructionControllerImplTest` and `KonstructionControllerConcurrencyTest` both construct
  `KonstructionControllerImpl(..., Dispatchers.Unconfined)`. It is not a new seam invented here.

I also ran the experiment the choice invites: patch the setter to do its write inside
`withContext(Dispatchers.IO)` — a plausible future refactor — and see whether the test still
catches #102.

| setter variant | result | listings |
|---|---|---|
| in-place, inline (control) | **FAIL** 470 torn | 1398 |
| in-place, inside `withContext(Dispatchers.IO)` | **FAIL** 46 torn | 224 |
| atomic, inside `withContext(Dispatchers.IO)` | pass | 318 |

So the test is **not** silently defanged by adding a suspension point — good. But note the
second row: the probe's headroom collapses from ~1400 listings to ~224, i.e. from a 14× margin
over the guard to 2.2×. Which leads to the one thing I'd like changed, below.

### 4. Non-blocking, but please consider: pin the invariant the whole design rests on

The test's sensitivity depends on an invariant it never asserts — that `controller.info = …`
performs its write *before returning*. Today that is true and the numbers above show it. But
nothing in the test checks it, and the failure mode if it stops being true is exactly the one
issue #114 exists to end. A conflating/debouncing setter (a very plausible refactor: this
setter fires on every dirty-state change) would drop most of the 800 writes on the floor, and
the probe would race almost nothing while still reporting a healthy-looking listing count.

One cheap line makes that loud instead of silent — after the write loop:

```kotlin
val landed = File(kDir, INFO_JSON).inputStream()
    .use { env.config.json.decodeFromStream<KonstructionInfo>(it) }
assertEquals(
    "$padding-${CONTROLLER_WRITES - 1}",
    landed.konstruction.name,
    "the last controller write must have landed before we asserted — if this fails the " +
        "setter is no longer writing inline and this probe is racing nothing"
)
```

Deterministic under `Unconfined`, zero added runtime, and it converts "probe measured nothing"
from a confusing `listings` failure (or a pass) into a message that names the cause.

### 5. Smaller notes

- `val padding = "n".repeat(2_000)` duplicates the companion's `PADDING`, which already carries
  the KDoc explaining *why* 2 000 ("big enough that a truncated write is very likely to land
  mid-token"). Reuse the constant so that rationale stays attached.
- `assertTrue(listings.get() > 100, "the probe only proves something if it listed a lot")` drops
  the count; the sibling `listingNeverSeesAHalfWrittenInfoFile` ends its identical message with
  `: $listings`. That number is the entire diagnostic when the guard trips — please include it.
- The header comment quotes the torn-decode path as `` `$konstruction` ``; #102 and #114 both
  quote `$.konstruction`. The sentence's job is verbatim identification, so the dot matters.
- `pool.submit { … }`'s `Future` is discarded. Harmless today because `runCatching` inside the
  loop swallows everything, but if that body is ever changed the reader thread could die
  silently and the probe would read zero. Not worth changing now, just noting the shape.
- Cleanup checks out: `pool.shutdown()` + bounded `awaitTermination` in `finally`, no
  `newSingleThreadContext` thread is created because `saveContext` is injected, and after the
  runs there were **0** leftover `/tmp/concurrent-create-*` directories.

### 6. Placement

`ConcurrentCreateTest` already outgrew "id claiming" — it holds `listingNeverSeesAHalfWritten-
InfoFile` and `concurrentWritersToOneFileDoNotBreakEachOther`, so it reads as the #102
regression class and the new case sits naturally beside its `writeInfo`-level sibling. Fine as
placed. Two small things for the next reader: the class KDoc still says only "concurrent creates
must not mint the same id", which now under-describes the file; and
`KonstructionControllerConcurrencyTest` already has this exact fixture in its `@Before`
(`TestEnvironment` + `KonstructionControllerImpl(..., Dispatchers.Unconfined)`), which the new
test hand-rolls. Either home is defensible — just don't leave the KDoc pointing at half the class.

### 7. Cost, stability, and the merge gate

Three full `spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true` runs, XML
deleted first each time and mtimes checked against the clock:

| run | wall | result | new test |
|---|---|---|---|
| 1 | 12m49s | 169 tests, **0 failures**, 6 skipped | 0.094s |
| 2 | 11m08s | 169 tests, **0 failures**, 6 skipped | 0.095s |
| 3 | 11m11s | 169 tests, 1 failure, 6 skipped — see below | 0.095s |

**Added CI cost is ~0.1 s.** `ConcurrentCreateTest` was green in all three, as was every other
case in it; the new test never varied by more than a millisecond.

Run 3's single failure is **not this PR**:

```
FailedRenderRetryTest > reRequestOfAFailedTargetRunsTheRenderAgain FAILED
java.io.FileNotFoundException: /tmp/konstructor-test-…/ws1/k1/result.json (No such file or directory)
    at KonstructionControllerImpl$lastRenderResult$2.invokeSuspend(KonstructionController.kt:154)
```

That is the same class currently failing CI **on `main`** at `332c472` (run 32559116186, same
method), and it is why this PR's own CI run is red — its failure there is
`FailedRenderRetryTest.alreadyCleanTargetThatFailsBecomesDirtyAgain`, the sibling method.
`main` was green at `56ff9c0` and `7f96989`, and went red with `332c472` (#113 / #104). So:

⚠️ **Do not merge yet, and do not merge on a "CI is red anyway" argument.** `main` is
unprotected, so `--auto` is not a gate; `gh pr checks 118` must reach completed SUCCESS on its
own. This branch cannot get there until the inherited `FailedRenderRetryTest` flake on `main`
is fixed and this branch is rebased. That is a merge-sequencing block, not a defect in this
change — approving on the merits.

monkopedia-reviewer (approved):
## Re-verify on the new base — APPROVE (independent, second pass)

Approved before at `746d654`; re-verified here at `cdafdf2` after the rebase onto the green `main` (`5ec8579`, the #113 revert). The only reason this got a second look is that tonight's #113 failure was exactly an approval earned on one base not holding on another, so a carried-over approval is not self-evidently valid. It holds. Everything below was re-measured on the new base.

### 1. The rebase is provably content-identical

- `git patch-id --stable` is the same for both heads: `4290a1889734e81c33bfafe2150b90e7cdd0c33c` at `746d654` and at `cdafdf2`.
- `git diff 746d654 cdafdf2` is byte-for-byte equal to `git diff 332c472 5ec8579` — i.e. the entire head-to-head delta is the base's #113 revert and nothing else. Diffed the two patch files directly; no difference.
- One commit, one file, `+64 −0`: `backend/src/jvmTest/.../ConcurrentCreateTest.kt`. No production change; nothing crept in.
- Parent of `cdafdf2` is `5ec8579` = `origin/main`. `MERGEABLE`.

### 2. Green on the new base — twice, matching the author

Own worktree at `cdafdf2`, `./gradlew spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true`, test-results XML deleted before each run:

| | result | tests | failures | errors | skipped |
|---|---|---|---|---|---|
| run 1 | BUILD SUCCESSFUL, 12m30s | 165 | 0 | 0 | 6 |
| run 2 (`--rerun-tasks`) | BUILD SUCCESSFUL, 12m26s | 165 | 0 | 0 | 6 |

Exactly the author's **165 / 0 / 6**. The new test genuinely ran both times, not skipped: `aListingNeverSeesTheControllersInfoFileHalfWritten[jvm]` at `time="0.091"` and `time="0.097"`. XML mtimes were seconds behind the wall clock at read time (23:44:58 vs 23:45:10; 00:10:25 vs 00:10:39), 32 result files, 75/76 tasks executed — real runs, not stale output from a denied slot. ~0.1s added cost, as measured before.

### 3. The RED still bites on the new base — and the blind spot is still real

One surgical change on top of `cdafdf2`: the controller's setter back to the in-place form (`writeInfo(...)` → `paths.infoFile.outputStream().use { config.json.encodeToStream(value, it) }`), matching the historic pre-atomic code at `8fdf760`. `writeInfo` itself and every signature untouched.

```
ConcurrentCreateTest[jvm] > aListingNeverSeesTheControllersInfoFileHalfWritten[jvm] FAILED
    java.lang.AssertionError at ConcurrentCreateTest.kt:286
java.lang.AssertionError: 374 of 1220 listings saw a torn info.json — the controller's
write is no longer atomic (konstructor#102). expected:<0> but was:<374>
```

30.7% torn, in line with the 24%/34% measured pre-rebase. A real test failure, not a compile error: 6 tests in the suite, the failing one at `0.071s`, fresh XML.

The part worth stating plainly: in that same RED run **the other five tests in the class all passed**. That is the PR's whole thesis reproduced on the new base — revert the controller's writer and the pre-existing suite is silent; only this test speaks. `listings.get()` was 1220 against the `> 100` guard, ~12× margin here on top of the 20–27× measured under core pinning.

### Not blocking

- The engine-expiry noise (`This build of intellij-server has expired`) shows up in both green runs and accounts for the 6 skips. Pre-existing on `main`, unrelated to this PR.
- Design points settled in the prior review (`Dispatchers.Unconfined`, the `listings > 100` guard, placement in `ConcurrentCreateTest`) were not re-litigated; I have nothing to add against them.

Merging. `Fixes #114` closes the issue.

PR #121, opened 2026-08-23
fix(konstruct): don't read result.json when the render was skipped (#120)

### LINKED ISSUE #120: konstruct() reads result.json unconditionally, including when it skipped the render
Isolated during the #119 revert review, and **it is a defect on `main` right now** — independent of #104, which is why it is filed separately rather than left inside that issue's re-land notes.

## The defect

`KonstructionServiceImpl.konstruct()` ends by reading the last render result to broadcast `onTaskComplete`:

```kotlin
konstructionController.lastRenderResult().also { result ->
    broadcast { onTaskComplete(result) }
}
```

That read is **outside** the dirty-state guard. When the guard decides nothing needs rendering, `render()` never runs — and if the konstruction has never rendered at all, there is no `result.json`, so `KonstructionControllerImpl.lastRenderResult()` throws `FileNotFoundException` out of an ordinary `konstruct()` call.

Instrumented at the moment of failure, the script directory is `[compile.json, content.csgs, content.kt, info.json, out]` — no `renders/`, no `result.json`. The render genuinely never happened.

## It is not caused by #113, and the revert did not remove it

Verified by the #119 reviewer: that line is **byte-identical before and after #113** (`332c472`) and remains so after the revert (`5ec8579`). #113 merely added tests that reach the guard-skip path. **The revert removed the test that exposed the read, not the read.**

So it sits on `main` today, reachable by any caller that requests a konstruct on a target that is already clean and has never rendered.

## ⚠️ The obvious fix is a trap — measured, not guessed

Making `lastRenderResult()` return `null` when the file is absent, and having `konstruct()` return an empty success, **compiles and looks correct**. I wrote exactly that. Against the #113 tests it **still failed 2 of 3 loaded runs**, now as an assertion failure on a sibling method instead of an exception.

It converts a loud crash into a **quiet wrong answer**: `konstruct()` returns a fabricated success for work that did not happen, and `onTaskComplete` either lies or silently does not fire. Whatever the fix is, it has to decide what `konstruct()` *means* when it skips a render — not just stop the read from throwing.

Two coherent shapes, both needing a decision rather than a patch:
- **Report honestly**: a distinct "nothing to do" outcome the caller can tell apart from "rendered successfully". Costs a signature change.
- **Do not reach the read**: return early on the guard-skip path, before the broadcast. Simplest, but changes whether `onTaskComplete` fires for a no-op konstruct, which the frontend may rely on.

## Reproducing

It needs the guard to skip on a konstruction with no prior render. Under the full suite with CPU load this happened intermittently; the class alone under the same load did not reproduce in 5 runs, so **suite-level interference is involved and is itself unexplained** (tracked on #104). A deliberate reproduction — construct the state directly rather than waiting for timing — is worth more than a load loop, and would let this be fixed without depending on #104's mystery.

monkopedia-coder:
Fixes #120

## The defect

`konstruct()` ended by reading the last render result to broadcast `onTaskComplete`, **outside** the dirty-state guard. When the guard skips the render and the konstruction has never rendered, there is no `result.json`, so an ordinary `konstruct()` died with `FileNotFoundException`.

## Deterministic reproduction (no load, no timing)

`WorkspaceImpl.create()` writes a konstruction as `CLEAN` with no targets. So a `konstruct()` that arrives before anything is compiled always skips the guard and always reads a file that was never written. The new test constructs exactly that state directly — the old reproduction needed the full suite under CPU pressure, and this one does not depend on the unexplained suite-level interference tracked in the separate #104 report.

RED, with the fix surgically disabled (guard condition forced true, API unchanged so the test still compiles):

```
KonstructionServiceKonstructTest > skippedRenderWithNoPriorRenderReportsFailureRatherThanThrowing FAILED
    java.io.FileNotFoundException: /tmp/konstructor-test-.../ws1/k1/result.json (No such file or directory)
      at KonstructionControllerImpl$lastRenderResult$2.invokeSuspend(KonstructionController.kt:154)
      at KonstructionServiceImpl$konstruct...
2 tests completed, 1 failed
```

The sibling test (`skippedRenderReportsThePriorRenderResult`) passed in the same RED run, which is what shows the failure is the defect rather than a broken build.

## Which shape, and why

The issue offers two shapes. I measured what the callers need before picking.

**"Return early on the skip path, before the broadcast" — rejected: it breaks the frontend.** `KonstructionViewModel` sets `UiState.EXECUTING` and then relies on `onTaskComplete` to settle it back to `DEFAULT`:

- `setTargetEnabled()` — comment in-tree: *"onRenderChanged will populate renderPaths, and onTaskComplete resets _state to DEFAULT"*.
- `save()` → `requestKonstructs()` — fire-and-forget, so the return value is not even seen; only `onTaskComplete` (or an `onInfoChanged` that the skip path also does not send) clears the spinner.

Suppressing the broadcast for a no-op konstruct would hang the UI in EXECUTING. So the broadcast has to keep firing.

**"Report honestly" — taken, with the read made conditional rather than the result fabricated:**

- **A render result exists** → keep returning it. On the skip path the requested targets are already up to date, so the render that made them clean is the honest answer, and callers rely on re-requesting an already-built target still reporting SUCCESS (`BuildAndDownloadStlTest.testMultipleTargetsBuildAndDownload` builds `myCube` then `mySphere`; the second call skips because the first render built both, and asserts SUCCESS). Locked down by the second new test.
- **No render result at all** → `FAILURE` with a message saying nothing was built, `taskArguments` empty. This is the trap the issue warns about, avoided: an empty `SUCCESS` here would be a fabricated success for work that did not happen, and the caller's postcondition (the requested target is built) is genuinely not met. The message surfaces in the frontend's messages pane via `_messages.value = result.messages`.

No protocol/signature change was needed for that, so `TaskStatus` is untouched. **Fallback if a reviewer disagrees:** add a distinct third `TaskStatus` (e.g. `SKIPPED`) for the no-prior-render case. I did not, because `FAILURE` + message already distinguishes it from a successful render, tells the caller *why*, and does not force a new branch into every existing `status ==` check — while a `SKIPPED` value would have to be returned *only* in the no-prior-render sub-case to avoid breaking the already-built-target contract above, making the enum value oddly asymmetric.

## Implementation

`KonstructionController.hasRenderResult()` (a `renderResultFile.exists()` check, mirroring the existing `renderFile()` exists-checks) gates the read. `lastRenderResult()` stays non-null — making it nullable would have forced a `!!` at the in-guard call site right after `render()`, where a result always exists.

## Scope

Only the unconditional read. Nothing here touches the failed-target-staying-CLEAN behaviour or the suite interference described in #104 — that work was reverted and is being handled separately.

## Verification

`./gradlew spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true` → BUILD SUCCESSFUL, both test tasks executed (not UP-TO-DATE). Fresh XML: backend 149 tests / 6 skipped / 0 failures / 0 errors, protocol 17 / 0 / 0 / 0.

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionController.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionController.kt
@@ -56,6 +56,11 @@ interface KonstructionController {
     suspend fun compile()
     suspend fun lastCompileResult(): TaskResult
     suspend fun render(targets: List<String>): List<String>
+
+    /** Whether a render has ever completed, i.e. whether [lastRenderResult] has one to read. */
+    suspend fun hasRenderResult(): Boolean
+
+    /** The result of the last [render]; only valid when [hasRenderResult] is true. */
     suspend fun lastRenderResult(): TaskResult
     suspend fun renderFile(target: String): File?
 }
@@ -139,6 +144,10 @@ class KonstructionControllerImpl(
             }
             return executedTargets
         }
+    }
+
+    override suspend fun hasRenderResult(): Boolean = withContext(Dispatchers.IO) {
+        paths.renderResultFile.exists()
     }
 
     override suspend fun lastRenderResult(): TaskResult = withContext(Dispatchers.IO) {
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionServiceImpl.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionServiceImpl.kt
@@ -35,7 +35,9 @@ import com.monkopedia.konstructor.common.KonstructionListener
 import com.monkopedia.konstructor.common.KonstructionRender
 import com.monkopedia.konstructor.common.KonstructionService
 import com.monkopedia.konstructor.common.KonstructionTarget
+import com.monkopedia.konstructor.common.TaskMessage
 import com.monkopedia.konstructor.common.TaskResult
+import com.monkopedia.konstructor.common.TaskStatus.FAILURE
 import com.monkopedia.konstructor.common.TaskStatus.SUCCESS
 import com.monkopedia.konstructor.logging.LoggingService
 import com.monkopedia.konstructor.logging.WarehouseWrapper
@@ -61,6 +63,10 @@ import kotlinx.coroutines.withTimeout
 
 /** Upper bound on a single compile/konstruct/listener-callback before it's cancelled. */
 private val callTimeout = 120.seconds
+
+/** Reported by `konstruct` when there was nothing to render and nothing had ever been built. */
+private const val NOTHING_RENDERED =
+    "Nothing was built: this konstruction has no render yet. Compile it first."
 
 class KonstructionServiceImpl(
     private val config: Config,
@@ -300,12 +306,33 @@ class KonstructionServiceImpl(
                     }
                 }
             }
-            konstructionController.lastRenderResult().also { result ->
+            renderResult().also { result ->
                 broadcast {
                     onTaskComplete(result)
                 }
             }
         }
+
+    /**
+     * The result to report for a konstruct.
+     *
+     * Normally that is the last render's result — on the skip path the requested targets are
+     * already up to date, so the render that made them clean is the honest answer, and callers
+     * rely on an already-built target still reporting SUCCESS.
+     *
+     * A konstruction that has never rendered has no `result.json` at all, though, and reading
+     * it threw [java.io.FileNotFoundException] out of an ordinary `konstruct()` (#120). It
+     * can't be reported as an empty success either: nothing was built, and claiming otherwise
+     * turns the crash into a quiet wrong answer. Say what actually happened instead.
+     */
+    private suspend fun renderResult(): TaskResult = if (konstructionController.hasRenderResult()) {
+        konstructionController.lastRenderResult()
+    } else {
+        TaskResult(
+            status = FAILURE,
+            messages = listOf(TaskMessage(NOTHING_RENDERED))
+        )
+    }
 
     override suspend fun requestKonstruct(target: String): Unit =
         callContext("requestKonstruct", baseCallSign = konstructionController.callSign) {
--- a/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/KonstructionServiceKonstructTest.kt
+++ b/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/KonstructionServiceKonstructTest.kt
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2022 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+@file:OptIn(ExperimentalSerializationApi::class)
+
+package com.monkopedia.konstructor
+
+import com.monkopedia.konstructor.common.DirtyState
+import com.monkopedia.konstructor.common.Konstruction
+import com.monkopedia.konstructor.common.KonstructionInfo
+import com.monkopedia.konstructor.common.TaskResult
+import com.monkopedia.konstructor.common.TaskStatus
+import com.monkopedia.konstructor.logging.WarehouseWrapper
+import com.monkopedia.konstructor.testutil.FakeKonstructionListener
+import com.monkopedia.konstructor.testutil.TestEnvironment
+import java.io.File
+import kotlin.test.AfterTest
+import kotlin.test.BeforeTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeout
+import kotlinx.serialization.ExperimentalSerializationApi
+import kotlinx.serialization.json.encodeToStream
+
+/**
+ * What `konstruct` reports when its dirty-state guard decides there is nothing to render.
+ *
+ * A konstruction is created CLEAN with no targets ([WorkspaceImpl.create]), so a konstruct
+ * request that arrives before anything has been compiled skips the render — deterministically,
+ * with no timing involved. `konstruct` used to read `result.json` regardless, which in that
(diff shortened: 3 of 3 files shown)


monkopedia-coder:
Correcting one incidental line in the PR description before it propagates — it does not affect the change.

The report notes the verify run logged `This build of intellij-server has expired` ×4 and attributes it to "the known kotlin-lsp engine expiry". **It is not a real engine.** That string is emitted by the fake-engine shell script inside `KotlinLspProcessExpiredEngineTest`:

```
backend/src/jvmTest/.../KotlinLspProcessExpiredEngineTest.kt:71
    echo "This build of intellij-server has expired" >&2
```

The count is its signature: **4 = 1 + the 3-spawn cap** from #108's bounded respawn. It is the test proving the bound works, exactly as designed.

Measured on this machine: `/home/jmonk/kotlin-lsp` does not exist, `KONSTRUCTOR_KOTLIN_LSP` is unset, and `find / -name intellij-server` returns nothing. There is no engine here to expire — the 6 skips in every run are `BridgeLanguageServerIntegrationTest` (5) plus the soak test, taking the ordinary no-binary path.

Flagging it because **this is the second time tonight an agent has read that log line as evidence of a real expired engine**, and the first time I nearly relayed it upward as "prod's engine has probably expired too" — which would have been a manufactured alarm about the deployed instance built on a test fixture's output. The deployed engine's expiry status remains genuinely unknown and is recorded on #84; nothing in any local test run tells us anything about it.

monkopedia-reviewer (approved):
Independent review at `7bd1ed3` (own worktree, nothing built in the dirty main checkout). I did not write this change and checked the claims rather than the description.

## The trap was not walked into

The failure mode #120 warns about is a skip path that manufactures `TaskResult(emptyList(), SUCCESS, emptyList())` — compiles, stops the crash, and reports a success for work that never ran. This PR does not do that. On the skip path `konstruct()` now answers:

- **prior render exists** → that render's result, unchanged (`hasRenderResult()` true → the same `lastRenderResult()` read as before);
- **no render ever** → `TaskResult(taskArguments = emptyList(), status = FAILURE, messages = [TaskMessage("Nothing was built: …")])`.

I checked callers can tell the two apart rather than taking it on faith. `status` is the discriminator and every consumer reads it (`KonstructionViewModel.build/setTargetEnabled` surface `result.messages`; `JsBridge` serializes the whole `TaskResult`; the e2e tests assert `status == SUCCESS`). Nothing coerces `FAILURE` back into "rendered fine". That is an honest answer to "what does konstruct mean here": the postcondition *the requested target is built* is genuinely unmet, and the caller is told so with a reason.

## `onTaskComplete` still fires — and it has to

Verified independently of the PR body. `KonstructionViewModel` sets `UiState.EXECUTING` and only `onTaskComplete` clears it:

- `KonstructionViewModel.kt:372` — `onTaskComplete` is the sole `_state.value = UiState.DEFAULT` for the execute path;
- `setTargetEnabled` (:525) and `save()`/`onInfoChanged` → `requestKonstructs` (:317, :450) are fire-and-forget — the return value is never read, so a return-early fix that suppressed the broadcast would have pinned the spinner in EXECUTING forever. That is the "hung UI worse than the crash" outcome, and the chosen shape avoids it. The new test asserts the broadcast payload equals the return value in *both* skip sub-cases, so a future refactor cannot quietly drop it.

## Normal path is unchanged

The only edit at the call site is `konstructionController.lastRenderResult()` → `renderResult()`, and `renderResult()` delegates to exactly that call whenever the file exists. `render()` writes `result.json` before returning (`KonstructionController.kt:135-138`), so on the render-actually-ran path `hasRenderResult()` is always true and the behaviour is identical bar one `File.exists()`. The in-guard `lastRenderResult()` at :273 is left alone — correctly, since it runs immediately after `render()`.

## RED reproduced myself, and it is deterministic

Surgical revert (call site back to the unguarded read, API and test untouched so the class still compiles), 3 runs:

```
=== RED RUN 1/2/3 ===
KonstructionServiceKonstructTest[jvm] > skippedRenderWithNoPriorRenderReportsFailureRatherThanThrowing[jvm] FAILED
    java.io.FileNotFoundException at FileInputStream.java:-2
2 tests completed, 1 failed
<testsuite … tests="2" skipped="0" failures="1" errors="0" timestamp="2026-08-23T04:10:56.807Z">
```

`3/3 FAILED`, no load, no timing, ~0.23s per run — and the sibling test passed in the same runs, so it is a genuine test failure and not a compile error wearing one (`:backend:jvmTest` executed, fresh XML, `tests=2`). Fix restored → **5/5 PASS** (`tests="2" failures="0"`, XML timestamps 04:13:35 → 04:13:41). This is a real deterministic reproduction, not the load-luck one from last night: it builds the CLEAN/no-targets/no-render state directly.

## Scope is clean

Three files, +162/−1. The dirty-state guard, the target-state bookkeeping and the CLEAN-on-failure behaviour are untouched — nothing here re-lands #104 or pokes the suite interference recorded there. `lastRenderResult()`'s contract is unchanged (still non-null, still throws when absent; the KDoc just names the precondition), so the other two callers — `KonstructionControllerSoakTest:150` and `SubprocessFailureTest:142` — are not modified and not loosened; `KonstructionControllerImpl` is the only implementation, so no fake needed updating.

## Verification

`./gradlew spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true`, own worktree, JDK 21, all through the fleet build slot:

| run | condition | result | backend | protocol |
|---|---|---|---|---|
| 1 | normal | BUILD SUCCESSFUL 12m27s | 149 tests, 0 fail, 0 err, 6 skip | 17 / 0 / 0 |
| 2 | normal | BUILD SUCCESSFUL 10m50s | 149 / 0 / 0 / 6 skip | 17 / 0 / 0 |
| 3 | 4 CPU spinners | BUILD SUCCESSFUL 10m51s | 149 / 0 / 0 / 6 skip | 17 / 0 / 0 |

Test XML deleted before each run and mtimes checked against the clock; `:backend:jvmTest` executed (not UP-TO-DATE) every time. CI `build-and-test` on `7bd1ed3`: **pass** (21m31s).

(The `This build of intellij-server has expired` ×4 in every run is `KotlinLspProcessExpiredEngineTest`'s fake engine hitting the #108 respawn bound, as already noted in the thread — not a real engine, and not related to this change.)

## Non-blocking observations

1. **The message is honest but nearly invisible.** `TaskMessage(NOTHING_RENDERED)` has `line = null`, so `toDiagnostics()` (`EditorDiagnostics.kt:41`) and the cursor-line `activeMessage` (`KonstructionViewModel.kt:84`) both `mapNotNull` it away. What the user actually sees is the red `"1 message(s)"` status bar at `EditorPane.kt:178` — a signal that something is wrong, without the sentence explaining it. Better than a crash and better than a fake success, but if a line-less message should be readable, the status-bar fallback is the place.
2. **Residual, pre-existing, deliberately out of scope:** with a prior render present, the skip path still reports that render's SUCCESS even when the requested target is not in its `taskArguments`. That is the contract `BuildAndDownloadStlTest.testMultipleTargetsBuildAndDownload` depends on and the PR preserves it on purpose — worth noting it is a milder second instance of "answers for work this call did not do", if it ever deserves its own issue.
3. `hasRenderResult()` + `lastRenderResult()` is a two-step TOCTOU. Nothing deletes `result.json`, so it is safe today; the KDoc precondition is what carries it.

None of these block. Approving.

Issue #117, opened 2026-08-22
info.dirtyState is set CLEAN after every render, even one that failed


monkopedia-reviewer:
Deferred out of #113 (the #104 retry fix), which called it out and deliberately left it alone. Pre-existing, not introduced there.

## The behaviour

After a render, `KonstructionServiceImpl.konstruct()` writes:

```kotlin
val newInfo = info.copy(
    dirtyState = CLEAN,     // :293 — unconditional, even when the render FAILED
    targets = targets
)
```

`info.dirtyState` goes `CLEAN` whatever the render reported. Since #113 the **per-target** states are accurate (a target that failed at execute time is `NEEDS_EXEC`), and the second clause of the `:263` guard is what carries the retry, so nothing is stuck. But the konstruction-level flag now says "clean" about a konstruction whose last render failed.

## Why the obvious fix is wrong — read this before touching it

The tempting rewrite is "`dirtyState = NEEDS_EXEC` if any target is dirty". That breaks two ways:

1. **It would infinite-loop the frontend.** `KonstructionViewModel.onInfoChanged()` fires `requestKonstructs(...)` whenever `info.dirtyState == NEEDS_EXEC`. A permanently-failing target would then drive render → fail → info change → re-request → render, forever. The unconditional `CLEAN` is currently what prevents that, so it is load-bearing rather than merely sloppy.
2. **It would pin `dirtyState` to `NEEDS_EXEC` forever** for any script with a declared-but-never-exported target: such a target is never attempted, so it holds `NEEDS_EXEC` indefinitely by design (that is what makes it explicitly requestable).

So this needs the frontend's auto-build trigger reworked alongside it — e.g. distinguishing "dirty because never built" from "dirty because it just failed, do not auto-retry" — not a one-line change in `konstruct()`.

## Why it is worth fixing anyway

`ScriptHostImpl.findScript()` gates a cross-konstruction import on `controller.info.dirtyState != CLEAN` to decide whether to recompile. A konstruction whose render failed reads `CLEAN` there, which is at least misleading, and any future consumer that treats `dirtyState` as "the last build was good" will be wrong.

Issue #116, opened 2026-08-22
A failed export that was not the requested target keeps broadcasting its stale STL


monkopedia-reviewer:
Found while re-reviewing #113 (the #104 retry fix). **Not a regression** — `main` behaves identically, both before and after #113 — so it did not block that merge. Filing so it is not lost.

## The behaviour

Two sets that look alike are not:

* `KonstructionControllerImpl.render(targets)` deletes the `.stl` of the **requested** targets only (`KonstructionController.kt:128`).
* The `onRenderChanged` broadcast loop in `KonstructionServiceImpl.konstruct()` covers every **attempted** target — the script's exports plus the valid requested extras (`:303`).

So an export that was *not* the requested target and then fails at execute time still has its previous `.stl` sitting on disk. `konstructed()` finds it, and the broadcast hands the editor that **stale path instead of `null`**.

Consequences in `KonstructionViewModel`:

* the viewport keeps drawing geometry for a target that just failed;
* `renderPaths` keeps the key, so `setTargetEnabled()`'s `if (renderPaths.value.containsKey(name)) return` short-circuit closes the toggle-off/on rebuild route for that target.

The dirty state is correct (`NEEDS_EXEC`) and the primary `requestKonstruct` retry works, so this is cosmetic-plus-one-blocked-route, not a stuck render.

## Measured

Script with exports `alpha` and `beta` plus a declared-but-not-exported `gamma`. Render 1 (requesting `alpha`) builds everything. `beta` is then broken out of band, and render 2 is entered by requesting `gamma`, so `beta` is attempted but never requested:

```
states={alpha=CLEAN, beta=NEEDS_EXEC, gamma=CLEAN}
renders={alpha=…/alpha.stl, beta=…/beta.stl, gamma=…/gamma.stl}
beta.stl still exists=true
konstructed(beta)=model/ws1/k1/renders/beta.stl
```

`beta` failed, is correctly dirty, and is still advertised to the editor with a model.

## Direction

Make the two sets agree. Either delete the STL of every target the render is about to *attempt* rather than only the requested ones, or have the broadcast consult what actually built and send `null` for an attempted-but-unbuilt target regardless of whether its file happens to still exist.

Note the constraint #113 ran into: narrowing the broadcast loop to the built list alone is **not** the fix — that drops the `onRenderChanged(target, null)` for a target whose STL *was* deleted, which is the very thing that clears the editor and frees `renderPaths`.

## Coverage

[...]

Issue #115, opened 2026-08-22
The respawn cap no longer bounds an engine that initializes then dies, repeatedly


monkopedia-coder:
Found by the independent re-review of PR #112 (merged as `7f96989`). Not a defect in that PR's stated goal — it does what it claims — but **it narrowed a guarantee that #108 had, and the reviewer said so plainly rather than letting the merge imply otherwise.**

## The trade

#108 bounded respawns by TIME: at most 3 spawns in any 60-second window, then a 5-minute cooldown. #112 replaced that with a **consecutive-failure count, reset by an answered handshake** — a better fix for what it was aimed at, because it is time-independent and no future deadline can outrun it, and because a healthy engine that crashes and respawns cleanly should not count against a cap meant to stop hammering a broken one.

But the reset is exactly what re-opens a hole. **An engine that initializes successfully and then dies, over and over, resets the counter on every cycle and is respawned without limit.** The old window cap bounded that shape; the new counter does not.

## Severity — lower than #84, but real

Each cycle costs at least the 1.5s probe, and the corpse is destroyed before the next spawn, so this is not #84's cursor-rate JVM storm. It is a slow leak rather than a flood. The reviewer noted the KDoc argues the trade deliberately — which is right — but the honest summary is **"not strictly better"**, and that belongs in the tracker rather than only in a review comment nobody will find.

## The shape that reaches it

An engine that answers `initialize` and then exits during real work: an expired build whose check runs late, an OOM under a large workspace, a crash on a specific file. Each death looks like a fresh start to the counter.

## Direction

Keep the failure counter (it is the better primitive) and add a separate bound on *spawn rate* that a successful handshake does not reset — the two answer different questions: "is this engine broken right now" versus "are we spawning JVMs faster than is reasonable". A test must **force the cap to fire**, asserting equality with the bound rather than `<=`; per the #112 review, "at most the cap" is satisfied by a cap that never trips, which is how the previous one shipped unreachable.

## Second, related gap from the same review: the pull seam is bounded but not self-healing

`textDocument/diagnostic` now has a deadline, so it can no longer park forever. But **nothing on that path can trigger a reconnect** — the publisher lives in the scope that reconnect cancels — so a user who is only typing (no hover, no completion) still gets "LSP is on, no diagnostics ever appear". It is now *logged* rather than silent, which is a genuine improvement, but the state is still terminal for that session.

Worth deciding as one question with the above, since both are "the engine is unusable and the code believes otherwise".

[...]

PR #123, opened 2026-08-23
Re-land: keep a failed render target dirty so re-running retries it (#104)

### LINKED ISSUE #104: A failed render target stays CLEAN, so re-running never retries it

### LINKED ISSUE #122: A slow info persist resurrects stale dirty state, so konstruct() silently skips the render
## The defect

`KonstructionControllerImpl.info`'s setter publishes the new value synchronously and then
launches a persist job that ends with a **second** assignment:

```kotlin
set(value) {
    if (infoImpl == value) return
    // Optimistically set now, to have the info available immediately.
    infoImpl = value
    GlobalScope.launch(saveContext + callSign) {
        runCatching { contentFileLock.withLock { writeInfo(paths.infoFile, config.json, value) } }
            .onFailure { ... }
        // Always set one more time after write to settle out any race conditions.
        infoImpl = value
    }
}
```

That trailing line does the opposite of what its comment claims. The job captures the value it
was *launched* with, so when it finally runs it writes that value back over whatever has been
assigned since — **a lost update**.

## Why it fires in practice: compile() holds the same lock

The persist job takes `contentFileLock`. `KonstructionControllerImpl.compile()` holds
`contentFileLock` for the entire kotlinc run. So the ordinary edit-then-build sequence lines up
almost perfectly to lose the update:

| step | thread | effect |
| --- | --- | --- |
| `set(content)` | caller | `infoImpl = NEEDS_COMPILE`; launches job A |
| `compile()` | caller | takes `contentFileLock`, runs kotlinc (seconds) |
| job A | saveContext | blocked on `contentFileLock` for the whole compile |
| `compile()` | caller | releases lock, publishes `infoImpl = NEEDS_EXEC`; launches job B |
| job A | saveContext | finally writes, then **`infoImpl = NEEDS_COMPILE`** ← clobbers |
| `konstruct()` | caller | reads `NEEDS_COMPILE`, guard fails, **skips `render()`** |
| job B | saveContext | writes, restores `NEEDS_EXEC` |

The window is between compile()'s publish and job B running, so it widens under CPU load — the
saveContext dispatcher is a single thread.

## User-visible harm

A `konstruct()` landing in that window silently does nothing: the dirty-state guard in
`KonstructionServiceImpl.konstruct()` sees a state that says "not compiled yet" and skips the
render entirely. Since #121 that path reports the honest `Nothing was built: this konstruction
has no render yet. Compile it first.` FAILURE — for a konstruction that *did* just compile
successfully. Before #121 it threw `FileNotFoundException: .../result.json`.

## Measured evidence

Instrumented `KonstructionServiceImpl.konstruct()` and the test that trips it
(`FailedRenderRetryTest`, from the reverted #113 branch), full `:backend:jvmTest -Dintegration=true`
under four CPU spinners. Failing run:

```
DBG104 konstruct req=[bad] info=KonstructionInfo(..., dirtyState=NEEDS_COMPILE, targets=[])
DBG104 first=TaskResult(taskArguments=[], status=FAILURE,
        messages=[Nothing was built: this konstruction has no render yet. Compile it first.])
DBG104 info=KonstructionInfo(..., dirtyState=NEEDS_EXEC, targets=[])
DBG104 compileResult={"status":"SUCCESS"}
```

`compile()` returned SUCCESS (so it had published `NEEDS_EXEC`), `konstruct()` then read
`NEEDS_COMPILE`, and immediately afterwards the same field read `NEEDS_EXEC` again. A value that
goes `NEEDS_EXEC` -> `NEEDS_COMPILE` -> `NEEDS_EXEC` with no intervening `set()` can only come
from the stale trailing assignment: `infoImpl` has exactly two writers, `ensureLoaded()` (which
returns early once initialised) and this setter.

The same log shows a second, harmless instance in a different test, so it is not a one-off.

## Fix

Delete the trailing assignment. The value is already published synchronously before the launch,
so the line is redundant as well as harmful, and the file writes stay correctly ordered because
`saveContext` is a single thread and the jobs are queued in setter order.

## Relationship to #104

This is what actually made `main` red after #113 and forced the #119 revert. #121 fixed the
proximate crash (`konstruct()` reading `result.json` on the skip path) but not the reason the
render was skipped, so #104's re-land still fails ~1 run in 3 without this.

monkopedia-coder:
Fixes #104
Fixes #122

Re-land of PR #113 (`332c472`), which was reverted in PR #119 (`5ec8579`) because it made `main`
intermittently red. Two commits, and the split is the point:

| commit | what |
| --- | --- |
| `fix(info): stop a slow persist from resurrecting stale dirty state` | the prerequisite — a pre-existing lost-update race that has nothing to do with #104, filed as #122 |
| `fix(render): keep a failed target dirty so re-running retries it` | #113 restored **verbatim** |

## The re-land itself is unmodified

The second commit is `332c472` restored by `git revert 5ec8579`, verified diff-identical by
comparing its `+`/`-` lines against `git diff 332c472^ 332c472` — same 7 files, same
405 insertions / 25 deletions. The only merge resolution was the `hasRenderResult()` declaration
that PR #121 added next to `render()` in the `KonstructionController` interface; the changed line
(`render()`'s return type) is byte-identical to #113's.

Its three-way dirty-state decision — built → CLEAN, attempted-and-failed → NEEDS_EXEC, not
attempted → unchanged — is untouched, as reviewed twice on the original PR.

## Why the same code is expected to behave differently now

The working hypothesis going in was that PR #121 (`a2d73a3`) had already removed the cause: it
stopped `konstruct()` from reading `result.json` unconditionally, which is what threw
`FileNotFoundException` when the tests ran. **That hypothesis was wrong, and the re-land still
failed** — on the very first full-suite run, unloaded:

```
FailedRenderRetryTest.reRequestRetriesATargetThatFailedAfterAlreadyBuilding
java.util.NoSuchElementException: Collection contains no element matching the predicate.
```

#121 turned the crash into the honest `Nothing was built: this konstruction has no render yet.`
FAILURE, which the test's `assertEquals(FAILURE, first.status)` accepts — so the run now walks one
line further before dying on an empty target list. Same defect, new mask.

### What is actually happening (#122)

Instrumenting `konstruct()` and reproducing under four CPU spinners:

```
DBG104 konstruct req=[bad] info=KonstructionInfo(..., dirtyState=NEEDS_COMPILE, targets=[])
DBG104 first=TaskResult(taskArguments=[], status=FAILURE,
        messages=[Nothing was built: this konstruction has no render yet. Compile it first.])
DBG104 info=KonstructionInfo(..., dirtyState=NEEDS_EXEC, targets=[])
DBG104 compileResult={"status":"SUCCESS"}
```

`compile()` had returned SUCCESS and published `NEEDS_EXEC`. `konstruct()` then read
`NEEDS_COMPILE`, failed the dirty-state guard and **skipped `render()`**. Immediately afterwards
the same field read `NEEDS_EXEC` again.

`NEEDS_EXEC → NEEDS_COMPILE → NEEDS_EXEC` with no intervening `set()` has one possible source.
`KonstructionControllerImpl.info`'s setter published the value synchronously and then ended its
async persist with `infoImpl = value` — *the value that job captured*:

```kotlin
infoImpl = value                                   // published here
GlobalScope.launch(saveContext + callSign) {
    runCatching { contentFileLock.withLock { writeInfo(...) } }...
    // Always set one more time after write to settle out any race conditions.
    infoImpl = value                               // ← resurrects it
}
```

`compile()` holds `contentFileLock` for the whole kotlinc run and the persist needs the same lock,
so the job launched by the preceding `set()` was still queued behind the compile when `compile()`
published `NEEDS_EXEC` — and then overwrote it with `NEEDS_COMPILE`. `infoImpl` has exactly two
writers (`ensureLoaded()`, which returns early once initialised, and this setter), so nothing else
can produce that sequence.

The line is redundant as well as wrong — the value is already published before the launch — so it
just goes. File writes stay correctly ordered: the jobs queue on one thread in setter order.

This is a **production** bug, not a test artefact: any user whose `compile()` is slow enough can
hit build-does-nothing, reported as a render that never ran.

### It also settles the open question on #104

The previous comment there recorded "class alone passes 5/5, full suite fails 2 of 3" and named
cross-test interference as the live hypothesis. **Disproved.** No state crosses between tests; the
window is "the single-threaded save dispatcher hasn't been scheduled yet", which widens with load
and with how much else the JVM is doing. That is also the exact mechanism behind the original #113
red — same skipped render, and pre-#121 `konstruct()` read `result.json` on that path.

## Coverage

`KonstructionControllerConcurrencyTest.aQueuedPersistDoesNotResurrectOlderInfo` drains the save
dispatcher by hand, so the interleaving is exact rather than load-dependent. Red-first verified: with
the offending line restored it fails with

```
A persist must never publish the info it captured over a newer one. Saw: [NEEDS_COMPILE, NEEDS_EXEC]
```

## Verification

`spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true`, test XML deleted before each
run, on this exact tree:

| run | CPU load | result |
| --- | --- | --- |
| 1 | none | 172 tests, 0 failures, 0 errors, 6 skipped |
| 2 | none | 172 tests, 0 failures, 0 errors, 6 skipped |
| 3 | 4 spinners | 172 tests, 0 failures, 0 errors, 6 skipped |
| 4 | 4 spinners | 172 tests, 0 failures, 0 errors, 6 skipped |
| 5 | 4 spinners | 172 tests, 0 failures, 0 errors, 6 skipped |
| 6 | 4 spinners | 172 tests, 0 failures, 0 errors, 6 skipped |

6/6 green, four of them loaded, each ~11m with `:backend:jvmTest` and `:protocol:jvmTest` actually
executing (not UP-TO-DATE) and all 34 result XMLs freshly written. For contrast, the same protocol
on the re-land **without** the #122 fix failed 1/1 unloaded and 1 of 2 loaded.

The 6 skips are the ordinary no-kotlin-lsp-binary path and the 4
`This build of intellij-server has expired` lines are the deliberate fake engine in
`KotlinLspProcessExpiredEngineTest` (1 + the 3-spawn cap), both unchanged from `main`.

## Not in scope

Issues #115, #116 and #117 live in the same function and are deliberately sequenced after this. No
third `TaskStatus` state — the owner ruled all-or-nothing.

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionController.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionController.kt
@@ -44,6 +44,15 @@ import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.json.decodeFromStream
 import kotlinx.serialization.json.encodeToStream
 
+/**
+ * What a render touched.
+ *
+ * [attemptedTargets] is the subset of [allTargets] the render actually tried to build — the
+ * script's exports plus any requested extras that exist. Which of those succeeded is reported
+ * separately, in the render's [TaskResult.taskArguments].
+ */
+data class RenderedTargets(val allTargets: List<String>, val attemptedTargets: List<String>)
+
 interface KonstructionController {
     val callSign: CallSign
     val paths: PathController.Paths
@@ -55,7 +64,7 @@ interface KonstructionController {
     suspend fun write(content: ByteReadChannel)
     suspend fun compile()
     suspend fun lastCompileResult(): TaskResult
-    suspend fun render(targets: List<String>): List<String>
+    suspend fun render(targets: List<String>): RenderedTargets
 
     /** Whether a render has ever completed, i.e. whether [lastRenderResult] has one to read. */
     suspend fun hasRenderResult(): Boolean
@@ -127,7 +136,7 @@ class KonstructionControllerImpl(
         }
     }
 
-    override suspend fun render(targets: List<String>): List<String> {
+    override suspend fun render(targets: List<String>): RenderedTargets {
         contentFileLock.withLock {
             for (target in targets) {
                 val targetFile = File(paths.renderOutput, "$target.stl")
@@ -144,11 +153,11 @@ class KonstructionControllerImpl(
                 extraTargets = targets,
                 subprocessExit = scriptExit
             )
-            val (result, executedTargets) = executeTask.execute()
+            val (result, renderedTargets) = executeTask.execute()
             paths.renderResultFile.outputStream().use { output ->
                 config.json.encodeToStream(result, output)
             }
-            return executedTargets
+            return renderedTargets
         }
     }
 
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionController.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionController.kt
@@ -88,7 +88,15 @@ class KonstructionControllerImpl(
         }
         set(value) {
             if (infoImpl == value) return
-            // Optimistically set now, to have the info available immediately.
+            // The in-memory value is published HERE, synchronously, and nowhere else. The
+            // persist below used to repeat `infoImpl = value` after its write "to settle out
+            // any race conditions", which did the opposite: the job captures the value it was
+            // launched with, so a slow write resurrected stale info over a newer assignment.
+            // `compile()` holds contentFileLock for the whole compile, so the job launched by
+            // the preceding `set()` was still queued on that lock when compile() published
+            // NEEDS_EXEC — and then overwrote it with NEEDS_COMPILE. A konstruct() landing in
+            // that window read the stale state, skipped render() and reported a render that
+            // never happened (#122).
             infoImpl = value
             GlobalScope.launch(saveContext + callSign) {
                 runCatching {
@@ -100,8 +108,6 @@ class KonstructionControllerImpl(
                         writeInfo(paths.infoFile, config.json, value)
                     }
                 }.onFailure { hauler.error("Failed to persist info for $workspaceId/$id", it) }
-                // Always set one more time after write to settle out any race conditions.
-                infoImpl = value
             }
         }
     override val scriptLock: Mutex = Mutex()
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionServiceImpl.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionServiceImpl.kt
@@ -269,17 +269,24 @@ class KonstructionServiceImpl(
             if (info.dirtyState == NEEDS_EXEC ||
                 info.targets.any { it.name in targets && it.state == NEEDS_EXEC }
             ) {
-                val allTargets = konstructionController.render(targets)
+                val rendered = konstructionController.render(targets)
                 val latestBuilt = konstructionController.lastRenderResult().taskArguments
                 val existingTargets = info.targets.associateBy { it.name }
                 val changedTargets = mutableListOf<KonstructionTarget>()
-                val targets = allTargets.map { target ->
-                    val dirtyState =
-                        if (target in latestBuilt) {
-                            CLEAN
-                        } else {
-                            existingTargets[target]?.state ?: NEEDS_EXEC
-                        }
+                val targets = rendered.allTargets.map { target ->
+                    val dirtyState = when {
+                        target in latestBuilt -> CLEAN
+
+                        // Attempted but not built: leave it dirty so the guard above lets a
+                        // later request re-enter render(). Keeping its previous state would
+                        // record a target that had already built as CLEAN even though its
+                        // render just failed, and the retry would be skipped in favour of
+                        // re-broadcasting the stale failure (#104).
+                        target in rendered.attemptedTargets -> NEEDS_EXEC
+
+                        // Untouched by this render, so its state is still whatever it was.
+                        else -> existingTargets[target]?.state ?: NEEDS_EXEC
+                    }
                     existingTargets[target]?.let {
                         if (it.state != dirtyState) {
                             it.copy(state = dirtyState).also(changedTargets::add)
@@ -295,12 +302,16 @@ class KonstructionServiceImpl(
                 konstructionController.info = newInfo
                 broadcast {
                     onInfoChanged(newInfo, changedTargets)
-                    for (rendered in latestBuilt) {
+                    // Every attempted target, not only the built ones: render() deleted the
+                    // STL of each target it was asked for, so one that then failed to build
+                    // must broadcast a null render (konstructed() finds no file) to clear the
+                    // geometry the editor is still showing for it.
+                    for (attempted in rendered.attemptedTargets) {
                         onRenderChanged(
                             KonstructionRender(
                                 info.konstruction,
-                                rendered,
-                                konstructed(rendered)
+                                attempted,
+                                konstructed(attempted)
                             )
                         )
                     }
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/tasks/ExecuteTask.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/tasks/ExecuteTask.kt
@@ -20,6 +20,7 @@ import com.monkopedia.hauler.error
 import com.monkopedia.hauler.hauler
 import com.monkopedia.hauler.info
 import com.monkopedia.konstructor.Config
+import com.monkopedia.konstructor.RenderedTargets
 import com.monkopedia.konstructor.common.MessageImportance
 import com.monkopedia.konstructor.common.TaskMessage
 import com.monkopedia.konstructor.common.TaskResult
@@ -58,14 +59,15 @@ class ExecuteTask(
     private val subprocessExit: Deferred<Int>? = null
 ) {
     private val hauler by lazy { hauler() }
-    suspend fun execute(): Pair<TaskResult, List<String>> {
+    suspend fun execute(): Pair<TaskResult, RenderedTargets> {
         val messages = mutableListOf<TaskMessage>()
         var isSuccessful = true
         val exports = script.listTargets(onlyExports = true)
         val allTargets = script.listTargets(onlyExports = false).map { it.name }
         val validExtraTargets = extraTargets.filter { it in allTargets }
-        val builtTargets = (exports.map { it.name } + validExtraTargets).distinct()
-        for (export in builtTargets) {
+        val attemptedTargets = (exports.map { it.name } + validExtraTargets).distinct()
+        val builtTargets = mutableListOf<String>()
+        for (export in attemptedTargets) {
             hauler.debug("Starting building $export")
             val exportService = script.buildTarget(export)
             val status = awaitTerminalStatus(export, exportService)
@@ -73,10 +75,14 @@ class ExecuteTask(
             val targetBuilt = status == BUILT
             // Accumulate — one failed target must fail the whole render. This previously
             // ASSIGNED (`isSuccessful = status == BUILT`), i.e. last-target-wins: with a live
-            // subprocess, target A failing then target B succeeding reported SUCCESS and the
-            // caller marked the failed target CLEAN, so it never retried (#81).
+            // subprocess, target A failing then target B succeeding reported SUCCESS (#81).
             isSuccessful = isSuccessful && targetBuilt
-            if (!targetBuilt) {
+            if (targetBuilt) {
+                // Only targets that actually built are reported. The caller derives dirty
+                // state from this list, so a failed target must be left out of it to stay
+                // NEEDS_EXEC and be retried by a later render (#104).
+                builtTargets += export
+            } else {
                 val trace = runCatching { exportService.getErrorTrace() }.getOrNull()
                 hauler.error("Error: $trace")
                 // Surface the execute-phase error to the UI. The trace is a runtime error,
@@ -111,7 +117,7 @@ class ExecuteTask(
             builtTargets,
             if (isSuccessful) SUCCESS else FAILURE,
             messages
-        ) to allTargets
+        ) to RenderedTargets(allTargets, attemptedTargets)
     }
 
     /**
--- a/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/KonstructionControllerConcurrencyTest.kt
+++ b/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/KonstructionControllerConcurrencyTest.kt
@@ -22,9 +22,11 @@ import com.monkopedia.konstructor.common.Konstruction
 import com.monkopedia.konstructor.common.KonstructionInfo
 import com.monkopedia.konstructor.testutil.TestEnvironment
 import java.io.File
+import kotlin.coroutines.CoroutineContext
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertTrue
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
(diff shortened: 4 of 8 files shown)


monkopedia-reviewer (approved):
## Independent review — approve

Reviewed at `f74a8e1` in a clean worktree off that SHA. I did not write this PR. Everything below
is my own measurement, not a re-reading of the PR body.

The title says "Re-land" and that framing is incomplete — this is a re-land **plus** an independent
production fix. I treated the two halves separately.

---

### 1. The #122 mechanism is real, and I verified it in the code rather than from the comment

`infoImpl` has exactly two writers, confirmed by `git grep infoImpl`: `ensureLoaded()` (guarded by
`hasInitialized`, so it writes at most once) and the `info` setter. There is no flow, no listener and
no other assignment, so a `NEEDS_EXEC → NEEDS_COMPILE → NEEDS_EXEC` sequence with no intervening
`set()` has one possible source.

The sequence lines up exactly as claimed:

| step | effect |
| --- | --- |
| `KonstructionServiceImpl.set()` | publishes `NEEDS_COMPILE`, launches job A capturing it |
| `KonstructionControllerImpl.compile()` | takes `contentFileLock` for the whole kotlinc run |
| job A | dispatched on the single-thread `saveContext`, suspends on `contentFileLock` |
| `KonstructionServiceImpl.compile()` | releases the lock, publishes `NEEDS_EXEC`, launches job B |
| job A | acquires the lock, writes, then `infoImpl = NEEDS_COMPILE` ← the lost update |
| `konstruct()` | reads `NEEDS_COMPILE`; both clauses of the dirty-state guard fail; `render()` skipped |
| job B | writes, restores `NEEDS_EXEC` |

The `set()` path also rewrites every target to `NEEDS_COMPILE`, so the guard's second clause
(`info.targets.any { it.name in targets && it.state == NEEDS_EXEC }`) fails too — which is why the
skip is total rather than partial, and why the observed failure is
`NoSuchElementException` out of `getInfo(Unit).targets.single { … }` in `FailedRenderRetryTest`
rather than a wrong dirty flag. That matches the reported red exactly.

### 2. Removing the line is safe, including the failed-write path — I checked this specifically

The old trailing assignment ran **after** `runCatching{…}.onFailure{…}`, i.e. unconditionally,
success or failure. So on a failed write it re-asserted a value that had already been published
synchronously at the top of the setter. Two cases:

- No newer `set()` in between → the re-assignment is a no-op. Nothing is lost by deleting it.
- A newer `set()` in between → the old code **reverted in-memory state to the value whose write had
  just failed**. That is strictly worse than the new behaviour.

So nothing depended on it, and the failed-write path in particular is improved, not regressed. After
a failed write, in-memory state is now the newest assigned value (correct) and `info.json` keeps its
previous contents — `writeInfo` is a write-to-temp + `ATOMIC_MOVE`, and it deletes the temp when the
move did not happen, so a failure never leaves a torn or partial file. Nothing in the process reads
`info.json` back after startup (`ensureLoaded()` runs once), so the only consequence of a failed
persist is a lost durable update, logged via `hauler.error`. That was true before this PR and is
unchanged by it.

Disk write ordering also survives: `saveContext` is single-threaded and `kotlinx.coroutines.Mutex` is
documented first-come-first-served, so job A (already waiting) is granted the lock before job B
(which asks later) and the writes land in setter order.

### 3. Red-first, verified by me, and 100% deterministic in both directions

I took a second worktree at the same SHA and **disabled only the fix** — restored the two removed
lines, nothing else (`git diff` in that tree was exactly `+2`).

```
:backend:jvmTest --tests "*KonstructionControllerConcurrencyTest*"
```

| tree | iterations | result |
| --- | --- | --- |
| fix disabled | 5 | **5/5 FAILED**, `tests="4" failures="1"` every time |
| PR as-is | 5 | **5/5 passed**, `tests="4" failures="0"` every time |

Every failing iteration produced the same assertion, verbatim:

```
A persist must never publish the info it captured over a newer one. Saw: [NEEDS_COMPILE, NEEDS_EXEC]
```

Fresh XML on every iteration (deleted before each run; mtimes 04:09:00 / :02 / :03 / :05 / :06, each
matching the clock at that iteration), `:backend:jvmTest FAILED` rather than `UP-TO-DATE`. That is a
genuine 10-out-of-10, not an 8-out-of-10 — which is what the test's design predicts, since it hand-drains
a queueing `CoroutineDispatcher` instead of racing. Reading it: both `set()`s complete before any
persist runs, the dispatcher executes nothing on its own, and the uncontended `Mutex` fast path means
each job runs to completion inside one `Runnable.run()`. Nothing is timing-dependent, and the
`guard++ < 100` loop would re-drain a job that did suspend and re-dispatch. The
`observed.isNotEmpty()` assertion correctly stops a silently-empty drain from passing.

### 4. The re-land half is byte-identical to #113 — mechanically proven

Isolating the second commit (`git diff 3e09c39 f74a8e1`) and comparing its `+`/`-` lines against
`git diff 332c472^ 332c472`:

```
diff <(grep -E '^[+-]' orig113.patch | grep -vE '^(\+\+\+|---)') \
     <(grep -E '^[+-]' reland.patch  | grep -vE '^(\+\+\+|---)')
→ IDENTICAL
```

The full patches differ only in blob hashes, hunk headers, and one context window — the
`hasRenderResult()` declaration #121 added next to `render()`. No changed line differs. Same 7 files,
405/25.

I re-read the restored logic anyway rather than trusting the identity check alone. The three-way
decision is intact and unmodified: `target in latestBuilt → CLEAN`,
`target in rendered.attemptedTargets → NEEDS_EXEC`, `else → existingTargets[target]?.state ?: NEEDS_EXEC`.
No one improved anything in passing. `TaskStatus` is still exactly `{SUCCESS, FAILURE}` on this head —
**no third state**, as ruled.

### 5. Protocol and on-disk

`KonstructionServiceTypes.kt` is still a KDoc-only change — the entire `protocol/` diff against the
merge base is five comment lines above `TaskResult.taskArguments`. `KonstructorTypes.kt`, which
actually defines `KonstructionInfo` / `KonstructionTarget` / `DirtyState`, is **untouched**, and so is
`InfoStore.kt`, which contains `writeInfo`. So `info.json`'s shape is byte-identical to `main`'s and
there is nothing for a round-trip to catch: the only values that change are `targets[].state`, which
now records an existing enum member (`NEEDS_EXEC`) where it used to record a stale one. I re-checked
this at this SHA rather than inheriting the earlier round.

One on-disk *semantic* change worth naming, non-blocking: `result.json`'s `taskArguments` now means
"targets that built" rather than "targets attempted". A `result.json` written by the current
production build and read by the new code will therefore mark a previously-failed target CLEAN once.
It self-corrects on the next render, the field has exactly one consumer
(`KonstructionServiceImpl.kt:273` — it is not read by the frontend), and the alternative is keeping
the bug. Fine as is; flagging it so it is a known property rather than a surprise.

### 6. Does the combination actually fix the red? Five runs, three loaded

`spotlessCheck :backend:jvmTest :protocol:jvmTest -Dintegration=true`, all test XML deleted before
each run, all through the fleet build slot, on this exact tree:

| run | CPU load | wall | result |
| --- | --- | --- | --- |
| 1 | none | 12m43s | BUILD SUCCESSFUL |
| 2 | none | 11m06s | **172 tests, 0 failures, 0 errors, 6 skipped** (34 XMLs) |
| 3 | 4 spinners | 11m08s | **172 / 0 / 0 / 6** (34 XMLs) |
| 4 | 4 spinners | 11m07s | **172 / 0 / 0 / 6** (34 XMLs) |
| 5 | 4 spinners | 11m07s | **172 / 0 / 0 / 6** (34 XMLs) |

5/5 green, three of them loaded (load average measured at 6.22 mid-run on run 3). On every run
`:backend:jvmTest` and `:protocol:jvmTest` are listed as executed, never `UP-TO-DATE`; run 1 reports
"75 actionable tasks: 75 executed". Every run's newest XML mtime equals that run's own end timestamp
to the second (e.g. run 5: `1787472422` vs `end epoch=1787472422`), so none of these are stale
artefacts from a denied slot. `KonstructionControllerConcurrencyTest`'s XML is present each time.

The 6 skips are the ordinary no-kotlin-lsp-binary path and the 4
`This build of intellij-server has expired` lines are the deliberate fake engine in
`KotlinLspProcessExpiredEngineTest` (1 + the 3-spawn cap) — I checked the count is exactly 4, matching
`main`.

**And the control that makes those five runs mean something.** Green runs alone cannot distinguish
"the fix worked" from "I got lucky five times", so I ran the same full suite twice more on the
fix-disabled tree, both under four spinners:

| control run (fix disabled) | result |
| --- | --- |
| 1 | BUILD FAILED — `FailedRenderRetryTest.alreadyCleanTargetThatFailsBecomesDirtyAgain` → `java.util.NoSuchElementException: Collection contains no element matching the predicate`, plus the new unit test |
| 2 | BUILD FAILED — the new unit test only; the integration test happened to pass |

Control run 1 reproduced **the exact original #104 red signature**, from a tree that differs from this
PR by two lines. That is the causal link stated as a hypothesis in the description, measured
independently. Control run 2 shows the integration half really is intermittent — 1 of my 2 loaded
controls, against the ~2-in-3 previously reported — which is precisely why the deterministic unit
test matters: it was red in 7 of 7 fix-disabled runs (5 targeted + both controls) and green in 10 of
10 with the fix. (Both controls report 155 tests rather than 172 because `:backend:jvmTest` failed
before `:protocol:jvmTest` ran.)

CI on the PR is green: `build-and-test` pass, 21m59s.

### 7. Should it be split? No — and I looked for a reason to say yes

The case for splitting is that a verbatim re-land is easier to trust when it is alone, and that each
fix gets its own bisect point. Neither survives contact:

- The verbatim property is already **mechanically provable** and I proved it above, in one command
  anyone can rerun. Splitting would buy a property the diff already has.
- The bisect points already exist — the branch is two commits, correctly separated, with the
  prerequisite first. (The one thing that would destroy that is a squash merge collapsing them. If
  you care about the two bisect points, this is the PR to merge rather than squash. Not a blocker,
  and I have no strong view.)
- Landing #113 alone is not an option: it cannot be green without #122, and that is what reddened
  `main` twice already.
- Landing #122 alone first *is* possible — it is independently testable, and I demonstrated exactly
  that by toggling one line and getting 5/5 red then 5/5 green. But it would cost a full extra review
  cycle to establish something the byte-identity check establishes for free.

I went looking for the thing a bundle would hide — a quiet edit to the reverted logic riding along
with the new fix — and there is none. The `+`/`-` sets are disjoint and each half is separately
verifiable. Bundle is the right call.

### 8. Non-blocking follow-ups

None of these block; all are pre-existing and none are made worse by this PR.

1. **`infoImpl` is not `@Volatile`** while being read and written from arbitrary dispatcher threads.
   This PR narrows the set of writers (the save thread no longer writes it), which helps, but the
   field is still unsynchronised. Cheap to harden.
2. **`ensureLoaded()` can clobber a `set()` that lands before the first `get()`** — the setter never
   sets `hasInitialized`, so a later getter re-reads `info.json` over it. Unreachable today only
   because `callContext(baseCallSign = konstructionController.callSign)` forces the getter on every
   service entry point. A `hasInitialized = true` in the setter would remove the trap.
3. **The write-ordering argument now rests on two implicit properties** — `saveContext` being a single
   thread and `Mutex` being fair. Both hold, but a one-line note next to `saveContext` would stop a
   future widening of that dispatcher from silently reintroducing a disk-level lost update.
4. `konstruct()` still publishes top-level `dirtyState = CLEAN` even when a target failed; retry
   depends entirely on the per-target `NEEDS_EXEC`. That is the design reviewed on #113 and it works,
   but it is load-bearing in a way that is easy to break. #115/#116/#117 are already sequenced here.

---

**Verdict: approve.** The #122 diagnosis is correct, I reproduced it, and the fix is safe on every
path including the failed write. The re-land is provably the same code that was reviewed twice. The
combination is green 5/5 including 3 under load, which is the thing the whole exercise was for.

PR #129, opened 2026-08-25
fix(backend): drain exec pipes before waiting so a noisy kotlinc can't deadlock (#125)

### LINKED ISSUE #125: ExecUtil.executeAndWait never drains kotlinc's pipes before waitFor(), so a compile emitting >64 KiB of output hangs forever (demonstrated)
**Category:** complexity / correctness (java-ism with a demonstrated hang)

**Where:** `backend/src/jvmMain/kotlin/com/monkopedia/konstructor/tasks/ExecUtil.kt:53-69` (`ExecUtil.executeAndWait`), sole caller `backend/src/jvmMain/kotlin/com/monkopedia/konstructor/tasks/CompileTask.kt:51`

**Current:**

```kotlin
fun executeAndWait(command: String): ExecResult {
    val rt = Runtime.getRuntime()
    hauler.debug("Executing: $command")
    val commands = arrayOf("bash", "-c", command)
    val proc = rt.exec(commands)
    return ExecResult(
        BufferedReader(InputStreamReader(proc.inputStream)),   // constructed — reads nothing
        BufferedReader(InputStreamReader(proc.errorStream)),   // constructed — reads nothing
        proc.waitFor()                                          // blocks with both pipes undrained
    ).also { ... }
}
```

Kotlin evaluates constructor arguments left to right, so the two `BufferedReader`s are merely *constructed* — wrapping a stream reads nothing — and then `waitFor()` blocks. **Nothing drains the child's stdout or stderr while it runs.** When the child fills the OS pipe buffer it blocks on `write()`, and `waitFor()` therefore blocks forever: classic deadlock.

The single caller is the user-script compile path:

```kotlin
// CompileTask.kt:50-51
val command = "kotlinc $opts -d ${output.absolutePath} ${input.absolutePath}"
val (stdOut, stdError, result) = executeAndWait(command)
```

`kotlinc` on a user script emits warnings and notes to stderr; the *only* bound on that volume is how noisy the user's script is. Errors are then parsed out of these same readers (`CompileTask.kt:56, 62`) — the reading happens strictly **after** `waitFor()` has already returned, which it may never do.

**Demonstrated, with a positive control** (JDK 21, this machine). A 25-line Java repro mirroring the exact shape above — construct both readers, then `waitFor()` — with an 8 s watchdog:

```
RESULT: RETURNED rc=0 for 1000 bytes            ← control: the probe CAN report the other answer
RESULT: RETURNED rc=0 for 65000 bytes
RESULT: HUNG (waitFor did not return in 8s) for 66000 bytes
RESULT: HUNG (waitFor did not return in 8s) for 70000 bytes
RESULT: HUNG (waitFor did not return in 8s) for 200000 bytes
```

The threshold sits between 65 000 and 66 000 bytes — i.e. exactly Linux's 65 536-byte default pipe capacity. Below it the child's output fits in the buffer and everything works, which is why this has never been hit in the test suite (the fixtures compile tiny, warning-free scripts).

**Note the file already disagrees with itself:** `ExecUtil.ExecProcess` (same file, `:71+`) *does* pump the child's stderr on a dedicated coroutine — that pump is the subject of the closed #101. So one exec path in this object handles the problem and the other does not.

**Proposed:** replace `Runtime.exec` + bare `waitFor()` with either

- `ProcessBuilder(...).redirectErrorStream(true)` plus reading the merged stream to completion *before* `waitFor()` (simplest; note `CompileTask` currently distinguishes stdout from stderr via `parseErrors(..., isError = true)`, so this changes error tagging), or
- read both streams into strings concurrently (two `launch`es on `Dispatchers.IO`, or `ProcessBuilder.redirectOutput/redirectError` to temp files) and only then `waitFor()`, keeping the two streams distinct.

Either way `ExecResult` should carry the already-read text rather than two live `BufferedReader`s over a finished process.

**Acceptance evidence:** a test that compiles a script producing >64 KiB of `kotlinc` output (e.g. a script with a few thousand deprecation warnings, or simply `bash -c` echoing that much in a unit test of `executeAndWait` itself) must complete rather than hang. Confirm the test *fails* against the current implementation first — the control above shows the shape of that RED.

**Effort:** small refactor. Contained to `ExecUtil.kt` plus the `ExecResult` consumers in `CompileTask.kt`. (`#65`, open, also touches `CompileTask.execute` — worth sequencing so the two don't collide.)

monkopedia-coder:
`ExecUtil.executeAndWait` constructed two `BufferedReader`s over the child's stdout and stderr and then called `waitFor()`. Wrapping a stream reads nothing, so **neither pipe was drained while the process ran**: once the child wrote past the OS pipe buffer (65536 bytes on Linux) it blocked in `write()` and `waitFor()` never returned. The sole caller is the user-script compile path, where the volume is bounded only by how many warnings the user's script provokes out of `kotlinc`.

## The change

- `executeAndWait` drains both streams on dedicated threads and waits afterwards — the same shape `ExecProcess`, in the same file, already uses for its stderr pump.
- `ExecResult` carries the text that was read rather than two live readers over a finished process, and `CompileTask.parseErrors` now takes that text. Stdout and stderr stay distinct, so the `isError = true` tagging is unchanged.

Kept deliberately small: no `ProcessBuilder`/lifecycle change, no temp files, no merged stream.

## Red/green — both measured

Regression test `ExecUtilPipeBufferTest` straddles the buffer at 65000 / 66000 bytes on stdout, on stderr, and on both at once, and runs the call on a throwaway thread with a 30s watchdog so a deadlock **fails** the test instead of hanging the suite. Output is measured, never printed, and tops out at ~200 KB total, so nothing accumulates the way the `copyTo(System.err)` writer in PR `#105` did.

**RED** — surgical mutation only (`waitFor()` moved back ahead of the two drains, API untouched, so the test still compiles):

```
./gradlew :backend:jvmTest --tests "*ExecUtilPipeBufferTest*"
4 tests completed, 3 failed          BUILD FAILED in 3m 37s (53 tasks executed)
  testOutputAbovePipeBufferIsReturned        FAILED  "did not return within 30000ms"
  testErrorOutputAbovePipeBufferIsReturned   FAILED  "did not return within 30000ms"
  testBothStreamsAbovePipeBufferStayDistinct FAILED  "did not return within 30000ms"
  testOutputBelowPipeBufferIsReturned        PASSED  <- 65000 bytes fits the buffer
```

The one passing case is the control: the probe can report the other answer, and the split lands exactly where the issue said it would.

**GREEN** — with the fix in place, re-run on the exact committed tree after deleting every result XML first:

```
./gradlew :backend:jvmTest      BUILD SUCCESSFUL in 10m 16s
159 tests, 0 failures, 0 errors, 19 skipped   (33 fresh result XMLs; the 19 skips are the
                                               integration tests that need -Dintegration=true)
  all four ExecUtilPipeBufferTest cases pass, each in single-digit ms
./gradlew spotlessCheck         BUILD SUCCESSFUL (20 tasks executed, 0 up-to-date)
```

Everything above ran under the fleet build slot; no run reported exit 75.

Fixes #125

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/tasks/CompileTask.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/tasks/CompileTask.kt
@@ -30,7 +30,6 @@ import com.monkopedia.konstructor.common.TaskResult
 import com.monkopedia.konstructor.common.TaskStatus.FAILURE
 import com.monkopedia.konstructor.common.TaskStatus.SUCCESS
 import com.monkopedia.konstructor.tasks.ExecUtil.executeAndWait
-import java.io.BufferedReader
 import java.io.File
 import kotlin.coroutines.EmptyCoroutineContext
 import kotlin.coroutines.coroutineContext
@@ -83,8 +82,8 @@ class CompileTask(private val config: Config, private val input: File, private v
         private val headerLines = KcsgScript.HEADER.split("\n").size
         private val footerLines = KcsgScript.FOOTER.split("\n").size
 
-        fun parseErrors(stdOut: BufferedReader, isError: Boolean = false): List<TaskMessage> =
-            stdOut.lineSequence().onEach {
+        fun parseErrors(output: String, isError: Boolean = false): List<TaskMessage> =
+            output.lineSequence().onEach {
                 if (isError) {
                     hauler.error(it)
                 } else {
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/tasks/ExecUtil.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/tasks/ExecUtil.kt
@@ -27,7 +27,10 @@ import com.monkopedia.ksrpc.ksrpcEnvironment
 import com.monkopedia.ksrpc.sockets.asConnection
 import java.io.BufferedReader
 import java.io.IOException
-import java.io.InputStreamReader
+import java.io.InputStream
+import java.util.concurrent.Callable
+import java.util.concurrent.FutureTask
+import kotlin.concurrent.thread
 import kotlin.coroutines.EmptyCoroutineContext
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CompletableDeferred
@@ -44,11 +47,7 @@ object ExecUtil {
     @OptIn(DelicateCoroutinesApi::class)
     private val hauler by lazy { hauler().asAsync(GlobalScope) }
 
-    data class ExecResult(
-        val stdOut: BufferedReader,
-        val stdErr: BufferedReader,
-        val returnCode: Int
-    )
+    data class ExecResult(val stdOut: String, val stdErr: String, val returnCode: Int)
 
     fun executeAndWait(command: String): ExecResult {
         val rt = Runtime.getRuntime()
@@ -59,14 +58,21 @@ object ExecUtil {
             command
         )
         val proc = rt.exec(commands)
-        return ExecResult(
-            BufferedReader(InputStreamReader(proc.inputStream)),
-            BufferedReader(InputStreamReader(proc.errorStream)),
-            proc.waitFor()
-        ).also {
+        // Both pipes have to be drained while the child is still running. Once either one
+        // fills the OS pipe buffer (65536 bytes on Linux) the child blocks in write() and
+        // never exits, so waiting first deadlocks for anything but a quiet command
+        // (konstructor#125). ExecProcess below pumps its stderr for the same reason.
+        val stdOut = proc.inputStream.drainAsync("stdout")
+        val stdErr = proc.errorStream.drainAsync("stderr")
+        return ExecResult(stdOut.get(), stdErr.get(), proc.waitFor()).also {
             hauler.debug("Done executing: $command (${it.returnCode})")
         }
     }
+
+    private fun InputStream.drainAsync(name: String): FutureTask<String> =
+        FutureTask(Callable { bufferedReader().use(BufferedReader::readText) }).also { drain ->
+            thread(isDaemon = true, name = "exec-$name", block = drain::run)
+        }
 
     class ExecProcess(private val proc: Process) {
         private val parentJob = SupervisorJob()
--- a/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/tasks/CompileTaskParseErrorsTest.kt
+++ b/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/tasks/CompileTaskParseErrorsTest.kt
@@ -28,15 +28,15 @@ class CompileTaskParseErrorsTest {
 
     @Test
     fun testParseEmptyOutput() {
-        val reader = "".reader().buffered()
-        val result = CompileTask.parseErrors(reader)
+        val output = ""
+        val result = CompileTask.parseErrors(output)
         assertTrue(result.isEmpty())
     }
 
     @Test
     fun testParseErrorPrefixed() {
-        val reader = "error: something went wrong".reader().buffered()
-        val result = CompileTask.parseErrors(reader)
+        val output = "error: something went wrong"
+        val result = CompileTask.parseErrors(output)
         assertEquals(1, result.size)
         assertEquals("error: something went wrong", result[0].message)
         assertEquals(ERROR, result[0].importance)
@@ -44,8 +44,8 @@ class CompileTaskParseErrorsTest {
 
     @Test
     fun testParseWarningPrefixed() {
-        val reader = "warning: deprecated".reader().buffered()
-        val result = CompileTask.parseErrors(reader)
+        val output = "warning: deprecated"
+        val result = CompileTask.parseErrors(output)
         assertEquals(1, result.size)
         assertEquals("warning: deprecated", result[0].message)
         assertEquals(WARNING, result[0].importance)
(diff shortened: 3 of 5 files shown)


monkopedia-reviewer (approved):
**Approve.** I reproduced both directions myself in a clean worktree off `5ee0a7c`, and the central risk for a deadlock fix — a test that is green because it never filled the pipe — does not apply here.

## The RED is a hang, and I measured it

Surgical mutation only: `waitFor()` back ahead of any draining, reads afterwards, `ExecResult(String, String, Int)` untouched so `ExecUtilPipeBufferTest` still compiles against the shipped API. `:backend:compileKotlinJvm` shows `executed`, not `UP-TO-DATE`, so the mutation was in the bytecode under test.

```
:backend:jvmTest --tests "*ExecUtilPipeBufferTest*"      BUILD FAILED in 1m 33s
tests=4 failures=3 errors=0        (fresh XML, result XML deleted first)

testOutputAbovePipeBufferIsReturned         30.001s  "did not return within 30000ms"
testErrorOutputAbovePipeBufferIsReturned    30.001s  "did not return within 30000ms"
testBothStreamsAbovePipeBufferStayDistinct  30.018s  "did not return within 30000ms"
testOutputBelowPipeBufferIsReturned          0.012s  PASSED
```

Three failures at exactly 30.0s each — timeout, not assertion. The 65 000-byte case passing in 12 ms is the positive control: the probe can report the other answer, and the split lands where #125 said it would.

## The volume is real, not a comment

Measured on the host rather than trusting the constant name:

```
bash -c "head -c 66000 /dev/zero | tr '\0' 'x'" | wc -c   ->  66000
fcntl(F_GETPIPE_SZ) on a fresh pipe                        ->  65536
```

66 000 emitted against a 65 536-byte capacity: 464 bytes over, enough to block `write()`, and the RED above proves it does.

## The length assertion is present — that is the part that matters

`assertEquals(ABOVE_PIPE_BUFFER, result.stdOut.length)` (and the stderr and both-streams variants) asserts the observed volume alongside "it returned". So the test cannot silently degrade into "returned quickly because nothing filled the pipe" — if the fixture shrinks, the assertion moves with it and the constants sit two lines apart under a name that says what they mean.

The one residual: nothing ties `ABOVE_PIPE_BUFFER` to the *actual* OS pipe capacity. On a kernel with 64 KiB pages the default pipe is 16 pages = 1 MiB, and this test would pass unfixed. CI is `ubuntu-latest` (x86_64, 4 KiB pages, 65 536), so this is theory today — but a runtime `F_GETPIPE_SZ` probe, or simply an order of magnitude more output, would make it self-checking. Non-blocking.

**Timeout calibration** is fine and specifically not load-sensitive: the failure is an unconditional deadlock, not a slow path, so 30 s buys robustness at zero flake risk. Green runs took 3–5 ms per case — a ~6000x margin.

## Convergence with `ExecProcess` — checked, and they should not merge

`ExecProcess` is not a second copy of the same job: its stdout is the live ksrpc duplex channel and its stderr is tee'd to `System.err`, neither captured. `executeAndWait` captures both to strings. There is nothing to unify beyond the invariant, and the new comment states it and points at `ExecProcess` explicitly, which is what stops the next occurrence. The mechanisms differ (daemon threads + `FutureTask` vs. the `SupervisorJob` scope), and that is defensible — `executeAndWait` is a plain blocking `fun`, so raw threads avoid dragging a coroutine scope into it, and `FutureTask` propagates a drain failure out of `get()` rather than dropping it.

Also worth recording: the fix no longer depends on argument evaluation order for correctness. Both drains are already running before the `ExecResult(...)` argument list is evaluated, so even if `waitFor()` ran first nothing deadlocks — the exact subtlety that caused #125 is no longer load-bearing.

## Collateral — nothing was weakened

Diffed the two existing test files line by line. Every edit is `"...".reader().buffered()` -> the string itself; **every assertion is byte-identical**. No expectation loosened, none deleted. Only nit: `val output = input` in two places is a pointless alias.

## Full suite, both forms — no #105-style OOM

```
:backend:jvmTest                                    BUILD SUCCESSFUL 12m
  159 tests, 0 failures, 0 errors, 19 skipped   (33 fresh XMLs, 53 tasks: 53 executed)

:backend:jvmTest :protocol:jvmTest spotlessCheck -Dintegration=true   BUILD SUCCESSFUL 11m 11s
  176 tests, 0 failures, 0 errors, 6 skipped    (35 fresh XMLs)
  CompileAndExecuteIntegrationTest 3/3, SubprocessFailureTest 6/6, FailedRenderRetryTest 4/4
  spotlessKotlinCheck executed in every module
```

The integration form is the one that matters here: it drives real `kotlinc` through the rewritten `executeAndWait` end to end. Result XML deleted before each run; no run reported exit 75.

## Minor notes, none blocking

1. **`BufferedReader.lineSequence()` != `String.lineSequence()` on a trailing newline.** `kotlinc` output ends in `\n`, so `parseErrors` now sees one extra empty trailing line. It is filtered before `map`, so no `TaskMessage` changes — but `onEach` logs it first, which means one extra empty `hauler.error("")` per compile. Cosmetic.
2. **Output is now fully materialized** rather than streamed line-by-line. Bounded by `kotlinc`'s noise, and the old `List<TaskMessage>` was already O(matching lines), so this is fine — just no longer constant-space.
3. **Drain failure leaves the process unreaped.** If `stdOut.get()` throws, `waitFor()` never runs and the sibling drain thread keeps going. Both threads are daemons so nothing wedges JVM exit, and the path needs an I/O failure on a process pipe. Worth a `runCatching`/`finally` if this file is touched again.
4. Streams are now closed (`use`), which the old code never did — a leak fixed in passing.

Scope is tight, the API change is minimal and fully propagated (`executeAndWait` has exactly one caller), and `isError` tagging is preserved with stdout and stderr still distinct — asserted by `testBothStreamsAbovePipeBufferStayDistinct`.

PR #133, opened 2026-08-26
Cache KonstructionControllers strongly so their locks can't be swapped

### LINKED ISSUE #126: KonstructorManager caches controllers in a WeakHashMap whose keys nothing holds, so a konstruction's save/script Mutexes can silently be replaced mid-flight
**Category:** complexity / correctness (non-idiomatic collection choice with a real concurrency consequence)

**Where:** `backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructorManager.kt:19, 22, 27-33`

**Current:**

```kotlin
import java.util.WeakHashMap

class KonstructorManager private constructor(private val config: Config) {
    private val controllers = WeakHashMap<Pair<String, String>, KonstructionController>()

    fun controllerFor(workspaceId: String, id: String): KonstructionController {
        synchronized(controllers) {
            return controllers.getOrPut(workspaceId to id) {
                KonstructionControllerImpl(config, workspaceId, id)
            }
        }
    }
    ...
}
```

`WeakHashMap` holds its **keys** weakly. The key here is `workspaceId to id` — a `Pair` **freshly allocated on every call**. The only strong reference to that `Pair` is the map entry itself, which is precisely what a weak key does not keep alive, and the value (`KonstructionControllerImpl`) holds no back-reference to its key. **Every entry is weakly reachable the instant it is inserted, and may be cleared at any GC.**

So the cache is not a cache with a lifetime policy — it is a cache that may or may not hold anything, nondeterministically.

**Why that matters here rather than being merely wasteful:** controller *identity* is load-bearing, because the mutual-exclusion primitives are per-instance fields of `KonstructionControllerImpl`:

- `contentFileLock` — `KonstructionController.kt:90`
- `scriptLock` — `KonstructionController.kt:112`

and both are reached **through this same cache**:

```
backend/src/jvmMain/kotlin/com/monkopedia/konstructor/hostservices/ScriptManager.kt:
    val lock = KonstructorManager(config).controllerFor(paths.workspaceId, paths.konstructionId).scriptLock
```

If the entry for a konstruction is collected between two `controllerFor` calls, two coroutines operating on the *same* konstruction receive *different* `Mutex` instances and the mutual exclusion silently evaporates — no exception, no log, just two writers in the critical section. The window is small and GC-dependent, which is the worst shape for a bug: it will not reproduce on demand and will look like a mystery corrupt `info.json` or a torn script write. Also, every collected entry means a fresh `KonstructionControllerImpl`, which allocates a dedicated thread (`KonstructionController.kt:82`, `newSingleThreadContext`) that nothing ever shuts down — so eviction leaks threads too.

**The one test that pins this invariant cannot fail:**

```kotlin
// backend/src/jvmTest/kotlin/com/monkopedia/konstructor/KonstructorManagerTest.kt
val controller1 = manager.controllerFor("ws1", "k1")
val controller2 = manager.controllerFor("ws1", "k1")
assertSame(controller1, controller2)
```

Two statements with no allocation pressure between them: no GC occurs, so it is green today for reasons unrelated to the code being correct. It reports "identity is guaranteed" while the collection type guarantees the opposite.

**No sign this was deliberate.** `git log` shows no commit message or comment explaining the weak reference, and the three sibling per-`Config` caches in this codebase all use strong maps: `PathController.kt` (`mutableMapOf`), `ScriptManager.kt` (`mutableMapOf`), `KotlinLspProcess.kt` (`getOrPut` on a strong map).

**Proposed:**
- Switch `controllers` to a strong `mutableMapOf` (matching its three siblings) — or, if unbounded growth per konstruction is a concern, key a `ConcurrentHashMap<String, KonstructionController>` on `"$workspaceId/$id"` and add an explicit eviction that also shuts the controller down. **A weak *key* is the wrong tool either way**; only a weak *value* map (or a key the value strongly references) would even express the intended idea.
- Give `KonstructorManagerTest` a test that can actually fail: acquire a controller, drop the reference, force a GC (`System.gc()` + allocation pressure), re-acquire, `assertSame`. Confirm it goes RED against `WeakHashMap` before landing the fix — otherwise the replacement gate is as inert as the current one.

**Effort:** mechanical for the map change; small for the test.

monkopedia-coder:
## What

`KonstructorManager.controllers` was a `java.util.WeakHashMap` keyed by `workspaceId to id`.

`WeakHashMap` holds its **keys** weakly. That key is a `Pair` allocated fresh on every call, and nothing outside the map ever holds it — the value keeps the two `String`s but never the `Pair`. So every entry was weakly reachable the instant it was inserted, and any GC could clear it.

Controller *identity* is load-bearing, because the mutual-exclusion primitives are per-instance fields of `KonstructionControllerImpl` (`contentFileLock`, `scriptLock`) and both are reached **through this cache** (`hostservices/ScriptManager.kt:68`, `hostservices/ScriptHostImpl.kt:58/72/89`). Once the entry is dropped, the next lookup constructs a second controller with its own Mutexes, and two callers working on the same konstruction lock different objects — no exception, no log.

It is also wider than "a GC landing between two adjacent statements". `KonstructionServiceImpl.kt:79` holds its controller in a field for the service's lifetime. That keeps the *value* alive but not the *entry*, because the entry's key is what is weak. From the first GC after the service is constructed, the script path and the service path use different mutexes for that konstruction, permanently.

## Change

`controllers` becomes a strong `mutableMapOf`, matching the sibling per-`Config` caches in `PathController.kt` and `hostservices/ScriptManager.kt`.

Entries are never evicted, deliberately. Because the locks are per-instance fields, **adding an eviction policy to this cache reintroduces exactly this defect** — there is a comment on the field saying so. The cost is one retained controller per konstruction ever touched, which is accepted: nothing in the codebase is scoped to a controller's lifecycle (`KonstructionController` has no `close`/`shutdown` at all, so there is nothing that release would trigger).

**No `ConcurrentHashMap` needed, and I checked rather than assumed.** `controllers` is private and its only access is inside `controllerFor`, wholly within `synchronized(controllers)` — the get / construct / put sequence is already atomic against other callers. A bare `getOrPut` outside a lock would indeed have reintroduced a narrower version of the same "two different Mutexes" bug; it is not outside a lock. The `Pair` key is fine under a strong map: `kotlin.Pair` is a data class over two `String`s, and `testSameIdReturnsSameController` exercises exactly that (two calls allocate two distinct-but-equal `Pair`s and must hit the same entry).

## Test — and the RED it was proved against

The existing invariant test could not fail:

```kotlin
val controller1 = manager.controllerFor("ws1", "k1")
val controller2 = manager.controllerFor("ws1", "k1")
assertSame(controller1, controller2)
```

Two adjacent statements, no allocation pressure, so no GC — green for reasons unrelated to the code being correct. That inert gate is why the defect survived.

The replacement, `testSameControllerAfterGarbageCollection`:

- holds `controller1` strongly for the whole test — the point is that the **entry** must survive, not merely the value;
- calls `forceWeakKeyCollection()`, which builds a canary entry with the *same* reachability shape (a weakly held key nothing else references, allocated in its own frame so no stack slot keeps it live) and drives `System.gc()` plus allocation pressure until a `WeakReference` to it is observably cleared **and** the canary `WeakHashMap` has expunged it;
- **fails the test outright** if that never happens inside a 30s budget. `System.gc()` is advisory, so the only pass path through the helper is one where collection was *confirmed*. A green here cannot mean "the GC simply never ran";
- then asserts both controller identity and `scriptLock` identity.

**Measured, full suite, no `--tests` filter, XML deleted before each run:**

| | tests | failed | errors | skipped |
|---|---|---|---|---|
| RED (`WeakHashMap`, new test only) | 160 | **1** | 0 | 19 |
| GREEN (strong map) | 160 | 0 | 0 | 19 |

RED failure, verbatim:

```
Controller cache dropped its entry across a GC; per-konstruction locks
(contentFileLock/scriptLock) are per-instance, so callers would now be locking
different Mutex instances for the same konstruction.
expected same:<...KonstructionControllerImpl@7c211fd0> was not:<...@7004e3d>
```

Two genuinely different instances — and note the test reached that assertion, i.e. the GC positive control was satisfied rather than the helper bailing out. It reproduced on the first `System.gc()` attempt (0.055s), so this is not a slow or marginal test.

`./gradlew :backend:jvmTest spotlessCheck` → `BUILD SUCCESSFUL`, `:backend:jvmTest` executed (not `UP-TO-DATE`), spotless clean across all five modules.

## Known cost, not fixed here

`KonstructionControllerImpl` takes `saveContext = newSingleThreadContext("KonstructionController")` typed as `CoroutineContext`, so it cannot be closed, and there is no shutdown path in `KonstructionController.kt`. **That thread was already never reclaimed** — under the old weak map it leaked one per GC cycle, i.e. per re-creation. A strong map makes it one thread per konstruction instead of one per GC cycle, which is strictly better than today, but the thread is still never released and this PR does not change that. Recording it so the cost of the never-evicting cache is on the record; out of scope here.

The lock-registry alternative (hoist `contentFileLock`/`scriptLock` off the cache so identity stops mattering) was considered on the issue and priced as a wider change. Not attempted here.

Fixes #126

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructorManager.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructorManager.kt
@@ -16,10 +16,27 @@
 package com.monkopedia.konstructor
 
 import com.monkopedia.konstructor.common.Konstruction
-import java.util.WeakHashMap
 
 class KonstructorManager private constructor(private val config: Config) {
-    private val controllers = WeakHashMap<Pair<String, String>, KonstructionController>()
+    // Strong map, entries never evicted. This used to be a WeakHashMap, which holds its
+    // KEYS weakly -- and the key here is `workspaceId to id`, a Pair allocated fresh on
+    // every call that nothing outside the map ever references. Every entry was therefore
+    // weakly reachable the instant it was inserted, so any GC could drop it even while the
+    // controller itself was strongly held by KonstructionServiceImpl. The next lookup then
+    // built a second KonstructionControllerImpl with its own contentFileLock/scriptLock,
+    // and the script path and service path locked different Mutexes for the same
+    // konstruction -- silently, permanently, from the first GC onward (#126).
+    //
+    // Controller identity is load-bearing precisely because those Mutexes are per-instance
+    // fields, so this cache must not evict: adding an eviction policy here reintroduces the
+    // same defect. Nothing is scoped to a controller's lifecycle (there is no close/shutdown
+    // on KonstructionController), so retaining one per konstruction for the process lifetime
+    // is the accepted cost. Matches the strong per-Config caches in PathController and
+    // ScriptManager.
+    //
+    // Mutation is confined to controllerFor below, entirely under `synchronized(controllers)`,
+    // which is what makes the get-construct-put sequence atomic.
+    private val controllers = mutableMapOf<Pair<String, String>, KonstructionController>()
 
     fun controllerFor(konstruction: Konstruction): KonstructionController =
         controllerFor(konstruction.workspaceId, konstruction.id)
--- a/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/KonstructorManagerTest.kt
+++ b/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/KonstructorManagerTest.kt
@@ -22,9 +22,13 @@ import com.monkopedia.konstructor.common.Konstruction
 import com.monkopedia.konstructor.common.KonstructionInfo
 import com.monkopedia.konstructor.testutil.TestEnvironment
 import java.io.File
+import java.lang.ref.WeakReference
+import java.util.WeakHashMap
+import java.util.concurrent.TimeUnit
 import kotlin.test.Test
 import kotlin.test.assertNotSame
 import kotlin.test.assertSame
+import kotlin.test.fail
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.json.encodeToStream
 import org.junit.After
@@ -67,6 +71,88 @@ class KonstructorManagerTest {
         assertSame(controller1, controller2)
     }
 
+    /**
+     * The controller cache must survive garbage collection.
+     *
+     * `contentFileLock` and `scriptLock` are per-instance fields of
+     * `KonstructionControllerImpl`, and callers reach them *through* this cache
+     * (`ScriptManager`, `ScriptHostImpl`). If the cache entry can disappear, two callers
+     * working on the same konstruction get two different `Mutex` instances and the mutual
+     * exclusion silently evaporates.
+     *
+     * Note `controller1` is held strongly for the whole test: the point is that the
+     * *entry* must survive, not merely the value. A weak-**key** map loses the entry even
+     * while the value is strongly reachable, because nothing outside the map ever holds
+     * the freshly allocated key.
+     *
+     * [forceWeakKeyCollection] fails the test if the JVM never actually collects, so a
+     * pass here cannot mean "the GC simply never ran" — the flaw that made
+     * [testSameIdReturnsSameController] unable to fail.
+     */
+    @Test
+    fun testSameControllerAfterGarbageCollection() {
+        env.createWorkspaceDir("ws1", "Workspace 1")
+        createKonstructionInfo("ws1", "k1", "test-k")
+        val controller1 = manager.controllerFor("ws1", "k1")
+
+        forceWeakKeyCollection()
+
+        val controller2 = manager.controllerFor("ws1", "k1")
+        assertSame(
+            controller1,
+            controller2,
+            "Controller cache dropped its entry across a GC; per-konstruction locks " +
+                "(contentFileLock/scriptLock) are per-instance, so callers would now be " +
+                "locking different Mutex instances for the same konstruction."
+        )
+        assertSame(
+            controller1.scriptLock,
+            controller2.scriptLock,
+            "scriptLock identity is the invariant that actually matters."
+        )
+    }
+
+    /**
+     * Positive control for the GC assertion above.
+     *
+     * Builds a canary entry with exactly the reachability shape of the cache under test —
+     * a weakly held key that nothing outside the map references — and drives GC until it
+     * is observably cleared. Returning normally means weak keys *were* collected in this
+     * JVM at this moment; if that never happens within the budget the test fails rather
+     * than proceeding to an assertion that could only pass vacuously.
+     */
+    private fun forceWeakKeyCollection() {
+        val canaryMap = WeakHashMap<Any, Any>()
+        val canaryRef = allocateCanary(canaryMap)
+        val deadline = System.nanoTime() + TimeUnit.SECONDS.toNanos(GC_BUDGET_SECONDS)
+        var attempts = 0
+        while (System.nanoTime() < deadline) {
+            attempts++
+            System.gc()
+            if (canaryRef.get() == null && canaryMap.isEmpty()) return
+            // Allocation pressure: System.gc() is advisory, so give the collector a
+            // reason as well as a request.
+            @Suppress("UNUSED_VARIABLE")
+            val ballast = ByteArray(BALLAST_BYTES)
+            Thread.sleep(GC_POLL_MILLIS)
+        }
+        fail(
+            "GC never cleared a weakly-keyed canary entry after $attempts attempts in " +
+                "${GC_BUDGET_SECONDS}s; the identity assertion would have been vacuous."
+        )
+    }
+
+    /**
+     * Allocated in its own frame so the only reference to the canary dies with that frame
+     * (a local in the calling method can otherwise stay live in a stack slot). The value
+     * is a constant String, never anything that refers back to the key.
+     */
+    private fun allocateCanary(map: WeakHashMap<Any, Any>): WeakReference<Any> {
+        val canary = Any()
+        map[canary] = CANARY_VALUE
+        return WeakReference(canary)
+    }
+
     @Test
     fun testDifferentIdReturnsDifferentController() {
         env.createWorkspaceDir("ws1", "Workspace 1")
@@ -75,5 +161,12 @@ class KonstructorManagerTest {
         val controller1 = manager.controllerFor("ws1", "k1")
         val controller2 = manager.controllerFor("ws1", "k2")
         assertNotSame(controller1, controller2)
+    }
+
+    companion object {
+        private const val CANARY_VALUE = "canary"
+        private const val GC_BUDGET_SECONDS = 30L
+        private const val GC_POLL_MILLIS = 20L
+        private const val BALLAST_BYTES = 4 * 1024 * 1024
     }
 }


monkopedia-reviewer (approved):
## Approve — the defect is real, the gate is real, and I reproduced both myself

Independent review from a worktree at `a90533a` (`/tmp/wt-133-rev`), off `origin/main` — not from any pre-existing checkout. Everything below is something I ran or read, not something I took from the PR body.

**On CI:** GitHub Actions was in a major outage while I worked, so `gh pr checks` read "no checks reported" for part of it (it now shows `build-and-test pending`, run `32988054560`). That was the outage, not a repo fact — `.github/workflows/ci.yaml` does trigger on `pull_request: branches: [main]`. **My verdict rests entirely on local runs, not on absent or pending checks.**

---

### 1. The RED reproduces, with the same shape and different addresses

I reverted the field surgically to `WeakHashMap` (import + declaration only, API and test untouched) and ran the **full** `:backend:jvmTest` — no `--tests` filter, results XML deleted first, XML mtimes checked against the clock, task confirmed `executed` not `UP-TO-DATE`.

| run | map | tests | failures | errors | skipped | build |
|---|---|---|---|---|---|---|
| GREEN #1 | strong | 160 | **0** | 0 | 19 | SUCCESSFUL (10m14s, 7 executed) |
| GREEN #2 (+`spotlessCheck`) | strong | 160 | **0** | 0 | 19 | SUCCESSFUL (10m16s, 7 executed) |
| RED | `WeakHashMap` | 160 | **1** | 0 | 19 | FAILED (10m15s, 9 executed) |

The instances I observed — different from yours, as they should be, and genuinely two objects:

```
expected same:<com.monkopedia.konstructor.KonstructionControllerImpl@59712875>
        was not:<com.monkopedia.konstructor.KonstructionControllerImpl@47b2e9e1>
```

RED test time `0.05s` — it reached the assertion, so the GC positive control was satisfied rather than the helper bailing out. Your reported numbers match mine exactly.

### 2. The GC control is sound — and I proved its failure branch is reachable

This was the place a hollow instrument would hide, so I did not take the claim on trust. Reading `forceWeakKeyCollection`: the **only** non-exceptional exit is the `return` guarded by `canaryRef.get() == null && canaryMap.isEmpty()`. `java.util.WeakHashMap.isEmpty()` delegates to `size()`, which calls `expungeStaleEntries()` — so the conjunction confirms *both* that the referent was cleared and that the map performed the very expunge that drops the real entry. Every other path falls out of the loop into `fail()`. **There is no vacuous pass.**

I then measured it, replicating the helper and the cache's reachability shape verbatim in a standalone JVM harness:

| JVM | n | control failed | same instance | different instance | attempts to confirm GC |
|---|---|---|---|---|---|
| default, strong map | 200 | 0 | **200** | 0 | ≤2 |
| default, `WeakHashMap` | 200 | 0 | 0 | **200** | ≤2 |
| `-XX:+DisableExplicitGC` (`System.gc()` is a no-op) | 1 | 0 | 1 | 0 | 77 (~1.6s) |
| `-XX:+UseEpsilonGC` (no collector at all), budget cut to 3s | 1 | **1 → `fail()`** | — | — | — |

Three things fall out of that:

- **The `fail()` branch is reachable.** Under EpsilonGC the loop exhausts its budget and takes it. A JVM that never collects makes this test go **red**, not green — which is exactly the property the old two-adjacent-statements test lacked.
- **It does not need `System.gc()` to be honoured.** With explicit GC disabled the allocation pressure still confirmed collection in 1.6s, so the helper degrades gracefully rather than either passing falsely or failing spuriously.
- **The gate's sensitivity is not marginal.** 200/200 detection against `WeakHashMap`, worst case 2 attempts. The 30s budget is ~400x headroom over the worst case I could produce with a working collector.

### 3. Not flaky

The new test executed **7 times on PR head: 7 passes, 0 failures.** Five filtered runs of `KonstructorManagerTest` (`--rerun`, results dir deleted between each) at 0.215–0.224s, plus the two full-suite runs above at 0.045s and 0.078s. Plus 200/200 in the harness. I found no run in which it flapped.

### 4. The `ConcurrentHashMap`-is-unnecessary claim holds — checked, not assumed

`git grep -n controllers -- '*.kt'` returns exactly three code references, all in `KonstructorManager.kt`: the declaration, the `synchronized(controllers)`, and the `getOrPut` inside it. `controllers` is `private`, has no accessor, and no other file touches it. The get-construct-put is therefore atomic against other callers, and a `ConcurrentHashMap` would buy nothing here (and a bare `getOrPut` outside the monitor would indeed have reintroduced a narrower two-Mutexes bug — it is not outside the monitor).

### 5. The retained cost is stated correctly, and is not worse than claimed

`KonstructionController.kt:82` — `private val saveContext: CoroutineContext = newSingleThreadContext("KonstructionController")`, typed as `CoroutineContext`, so `close()` is not reachable without a cast; `grep` for `close|shutdown` in that file returns nothing. So: one controller **and one never-reclaimed thread** per konstruction ever touched, unbounded. Your "strictly better than today" is right rather than a rationalisation: under the weak map an evicted entry produced a *fresh* controller with a *fresh* thread, so the thread count grew per GC cycle per konstruction, not per konstruction. This change bounds it to one. It is correctly recorded in the PR body as a known, unfixed cost.

One gap, not a blocker: **there is no open issue tracking that thread.** It exists only in this PR body, which is not where a future reader will look. Worth a one-line follow-up issue so it is on the backlog rather than in a merged description.

### 6. Non-blocking: the comment records the premise but not whose ruling it was

This is the one thing I would change, and it is the cheapest protection in the diff. The comment says *"Nothing is scoped to a controller's lifecycle … so retaining one per konstruction for the process lifetime is the accepted cost"* — the premise **is** in the code, which is the important half, and the "adding an eviction policy here reintroduces the same defect" warning is stated well. What is missing is that this was an **owner decision on a date**, and that the premise is the thing that can expire:

```
// Accepted by Jason, 2026-08-26 (#126): "we don't do anything scoped on the
// objects lifecycle, so ... the cost of it existing is pretty minimal."
// That premise is load-bearing: if anything ever DOES become scoped to a
// controller's lifecycle, this decision needs revisiting -- and the answer
// still is not an eviction policy on this map.
```

Without the attribution, a future reader sees an assertion they may feel free to re-litigate; with it, they see a ruling and the exact condition under which it stops applying. Land it or not — I am not blocking on it.

### 7. Two small notes

- `assertSame(controller1.scriptLock, controller2.scriptLock)` is strictly implied by the identity assertion above it. Harmless, and it documents *why* identity matters, so I would keep it.
- The canary is an `Any` key in a separate `WeakHashMap` while the real key is a `Pair`; in principle the two could sit in different generations, so "canary cleared" is not a formal proof that the cache entry would have cleared. In practice `System.gc()` is a full collection and I measured 200/200 detection, so this is a theoretical note, not a defect.

### 8. On the missing label

I checked rather than defaulting: **no label, deliberately.** All 12 most recently merged PRs in this repo (#100 through #129) carry zero labels, and the repo's custom labels (`agent-workable`, `needs-decision`, `health-suggestion`, `under-investigation`) are issue-triage machinery that drives dispatch — putting one on a PR would misfile it into that pipeline. `bug` would also be a stretch for a change whose linked issue is already labelled. Unlabelled is the correct state here.

---

**Verdict: approve.** The fix is minimal and correct, the invariant it protects is the one that actually matters, the test that guards it can fail for the right reason and I made it fail for the right reason, and the analysis in the PR body survived every check I ran against it. `spotlessCheck` clean on `:backend` — verified by execution, with an injected violation as a positive control to confirm the gate can go red.

Issue #135, opened 2026-08-26
ServiceHolder reconnect cadence: flat 3s retry with no backoff, jitter, or cap


monkopedia-coder:
**Split out of #127** (the connection-leak fix) rather than folded into it: closing a client that is already being discarded is mechanical, but *changing when we retry* is a policy call with product-visible consequences, and the triage note on #127 explicitly said to stage it separately.

The leak itself is fixed by the PR for #127; the cadence is untouched by that PR (the literal is now named `RECONNECT_DELAY_MS` but its value is still `3_000`).

**Where:** `frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/viewmodel/ServiceHolder.kt`

```kotlin
private const val RECONNECT_DELAY_MS = 3_000L
...
delay(RECONNECT_DELAY_MS)
```

**The behaviour:** while the backend is down, every open tab attempts a websocket handshake every 3s, forever — 20/minute, 1200/hour per tab — and every tab reconnects in lockstep the instant the backend comes back. There is no backoff, no jitter, and no attempt cap.

**The decision this needs from the owner**, because each option trades something real:

1. **Leave it at a flat 3s.** Konstructor is a small self-hosted tool; a handful of tabs polling a local server every 3s costs nothing, and a fast flat retry gives the best possible reconnect latency after a backend restart — which is the common case during development. Backoff makes that case *worse*: after a 60s outage an exponential schedule can leave the tab waiting tens of seconds after the server is already back.
2. **Exponential backoff with a ceiling** (e.g. 1s → 30s). Kinder to the server and to laptop batteries on a long outage, at the cost of reconnect latency.
3. **Backoff plus jitter.** Only matters if multiple tabs/clients are expected; it de-synchronises the thundering herd on recovery.
4. **An attempt cap** with a manual "Reconnect" affordance in the UI. This is the biggest change — it needs UI work, and it means a tab left open overnight stops reconnecting on its own.

Whichever is chosen, "reconnect latency after a backend restart" is the thing being traded away, and that is the developer-facing behaviour Konstructor is used for most.

**Related (also deliberately left alone in the #127 PR):** when the keep-alive ping fails, the loop `break`s without clearing `_service`/`_connected`, so for the length of the retry delay the UI still reports itself connected and still holds a stub pointing at a dead websocket. Clearing them there would surface a "Connection lost" state during the gap — a product-visible change that belongs with whatever cadence is chosen.

PR #136, opened 2026-08-26
Close the discarded HttpClient/websocket on every reconnect attempt (#127)

### LINKED ISSUE #127: ServiceHolder.connect() allocates a new HttpClient + websocket on every reconnect attempt and never closes any of them, retrying silently every 3s forever
**Category:** complexity / correctness (resource leak on the app's connection lifecycle)

**Where:** `frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/viewmodel/ServiceHolder.kt:47-76`

**Current — the whole method:**

```kotlin
private suspend fun connect() {
    while (true) {
        try {
            val hostname = window.location.hostname
            val port = window.location.port
            val protocol = if (window.location.protocol == "https:") "wss" else "ws"
            val url = "$protocol://$hostname:$port/konstructor"
            val env = ksrpcEnvironment { }
            val client = HttpClient { install(WebSockets) }        // ← allocated per attempt
            val conn = client.asWebsocketConnection(url, env)
            val stub = conn.defaultChannel().toStub<Konstructor, String>()
            stub.ping()
            _service.value = stub
            _connected.value = true
            // Keep-alive loop
            while (true) {
                delay(10_000)
                try { stub.ping() } catch (e: Exception) { break }
            }
        } catch (e: Exception) {                                   // ← `e` bound, never used
            _service.value = null
            _connected.value = false
        }
        delay(3_000)
    }
}
```

Four distinct problems in 30 lines, and this is the only path by which the app talks to the backend.

**1. A leaked `HttpClient` (and websocket connection) per attempt — the main one.**
`git grep -n 'HttpClient\|\.close()' -- 'frontend/src/**/*.kt'` returns **3 hits across 40 tracked frontend `.kt` files**: the import here, this construction, and one unrelated `serverForDispose?.close()` in `ui/editor/EditorPane.kt`. **Nothing ever closes this client or this connection.** Each `HttpClient` owns an engine and its own coroutine scope; each `conn` is a live websocket.

The loop reaches the construction on *every* iteration, so the leak occurs whether the attempt succeeds or fails:
- backend down → a new client every 3 s, all leaked (20/minute, 1200/hour, for as long as the tab stays open);
- backend restarts → the keep-alive `ping()` throws, `break`, `delay(3_000)`, and the **old, still-open** client is abandoned in favour of a new one.

`WebSocketReconnectionTest` exists in the e2e suite, so this path is exercised — but nothing there counts clients, so the leak is invisible to it.

**2. Fixed 3 s retry, no backoff, no jitter, no cap.** Every open tab hammers the server every 3 s indefinitely while it is down, and they all reconnect in lockstep the moment it comes back.

**3. Every failure is completely silent.** Both `catch (e: Exception)` blocks (`:66`, `:70`) bind `e` and never use it — nothing logged, nothing in the console. A user seeing "Connection lost" has no way to learn why, and neither does anyone debugging it. The codebase already has the right shape one directory over (`threejs/ThreeJsRenderer.kt:181-184`: rethrow `CancellationException`, then `consoleError`); this is the only other place that matters as much and it has neither half.

**4. `CancellationException` is caught.** `catch (e: Exception)` at `:66` and `:70` swallows cancellation. It happens to escape via the `delay(3_000)` at `:74`, which sits outside the `try` — so cancellation does eventually propagate today, but only by accident of statement placement. Any future edit that moves that `delay` inside the `try` turns this into an uncancellable loop.

**Proposed:**
- Hoist the `HttpClient` to a single instance created once in `ServiceHolder` (it is a `single` in `di/AppModule.kt` and lives for the page's lifetime), or keep it per-attempt and close it — plus the previous `conn` — in a `finally` before retrying.
- Exponential backoff with a ceiling and jitter instead of the flat `delay(3_000)`; name the two literals (`KEEP_ALIVE_INTERVAL_MS`, `RECONNECT_BASE_DELAY_MS`) — `10_000` at `:63` in particular is a tuning knob buried in a nested loop.
- Log the caught exception (`consoleError`) and rethrow `CancellationException` explicitly at both sites.

**Acceptance evidence:** an e2e/manual check that stops the backend for ~30 s and confirms the number of live websocket connections / `HttpClient`s stays at 1 rather than growing by one per 3 s. Confirm the check goes RED against the current code first.

**Effort:** small refactor, one file.

monkopedia-coder:
`ServiceHolder.connect()` allocated a new `HttpClient` — which owns an engine and its own coroutine scope — plus a live websocket on **every** pass of the reconnect loop, and closed none of them.

- backend down → one leaked client every 3s for as long as the tab is open (20/min, 1200/hr);
- backend restarts → the keep-alive `ping()` throws, the loop `break`s, and the **still-open** client is abandoned in favour of a new one.

## What changed

- A `ServiceConnection` / `ConnectionFactory` seam owns one attempt's resources (`HttpClient` + ksrpc `Connection` + stub). The loop closes it in a `finally`, so an attempt that is being discarded is released rather than abandoned — whether it failed outright or its keep-alive dropped out.
- Cancellation now closes the live connection too (`withContext(NonCancellable)`) instead of stranding it.
- The real factory closes its `HttpClient` if the handshake or stub lookup throws after the client is already allocated.
- Both `catch (e: Exception)` sites bound `e` and logged **nothing** — a reconnect storm was completely invisible. Both now `consoleError` the exception and rethrow `CancellationException` explicitly, matching `threejs/ThreeJsRenderer.kt`. Previously cancellation escaped only by the accident of `delay(3_000)` sitting outside the `try`; any edit moving that delay inside would have made the loop uncancellable.
- The two timing literals are named (`KEEP_ALIVE_INTERVAL_MS`, `RECONNECT_DELAY_MS`). **Values unchanged.**

## Deliberately not in this PR

The **retry cadence**. Flat 3s, no backoff/jitter/cap is a policy question, not a leak, so it is split out as #135 (`needs-decision`) rather than folded in — as the triage note asked. Also left alone there: the loop `break`s out of keep-alive without clearing `_service`/`_connected`, so the UI keeps reporting "connected" for the retry gap. Changing that is product-visible and belongs with whatever cadence is chosen.

## Test infrastructure

`:frontend` had **no test source set at all** — the module was entirely untested. This PR adds `wasmJsTest` (Karma + headless Chrome, the same machinery `:protocol` already configures) and wires `:frontend:wasmJsBrowserTest` into CI. The CI step resolves `CHROME_BIN` from whichever chrome/chromium is on the runner.

Before writing the leak test I ran a deliberately-failing probe test through the new source set to confirm the instrument *can* go red: `2 tests completed, 1 failed`. It can.

## RED / GREEN

The leak is resource retention, which passes silently. Three tests substitute a counting `ConnectionFactory` and assert on connections **allocated and never closed**.

RED came from a surgical mutation — dropping only the `finally { connection?.closeQuietly() }` line, leaving the API the tests compile against completely intact (a `BUILD FAILED` with `tests=0` would have been a compile failure, not a red test):

| test | RED (`finally` removed) | GREEN (this PR) |
|---|---|---|
| `failedAttemptsDoNotAccumulateOpenConnections` | 11 created, **0 closed** | 11 created, 11 closed, 0 live |
| `droppedKeepAliveDoesNotAccumulateOpenConnections` | **11 of 11** allocated and never closed | ≤1 live (only the currently-established one) |
| `cancellingTheScopeClosesTheLiveConnection` | **1 stranded** | 0 stranded |

`3 tests completed, 3 failed` → `tests="3" ... failures="0"`.

Commands (result XML deleted first, tasks confirmed `executed`):

```
CHROME_BIN=/usr/bin/chromium ./gradlew :frontend:wasmJsBrowserTest
./gradlew spotlessCheck :frontend:wasmJsBrowserDistribution :frontend:wasmJsBrowserTest
```

`spotlessCheck` clean; `:frontend:wasmJsBrowserDistribution` still builds with the new source set present.

## Not verified

GitHub Actions is mid-outage, so the new CI step has **not** been observed running on a runner — only locally against `/usr/bin/chromium`. It should be watched on its first green CI run.

Fixes #127

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -54,6 +54,15 @@ jobs:
         # a real kotlin-lsp engine, absent on the runner); soak stays off.
         run: ./gradlew test :backend:jvmTest :protocol:jvmTest -Dintegration=true
 
+      - name: Run frontend wasmJs unit tests (Karma + headless Chrome)
+        run: |
+          export CHROME_BIN="${CHROME_BIN:-$(command -v google-chrome \
+            || command -v google-chrome-stable \
+            || command -v chromium-browser \
+            || command -v chromium)}"
+          echo "Using CHROME_BIN=$CHROME_BIN"
+          ./gradlew :frontend:wasmJsBrowserTest
+
       - name: Run e2e tests under Xvfb
         run: xvfb-run --auto-servernum --server-args='-screen 0 1280x720x24' ./gradlew :e2e:test -Pe2e
 
--- a/frontend/build.gradle.kts
+++ b/frontend/build.gradle.kts
@@ -29,11 +29,21 @@ repositories {
 
 kotlin {
     wasmJs {
-        browser()
+        browser {
+            testTask {
+                useKarma {
+                    useChromeHeadless()
+                }
+            }
+        }
         binaries.executable()
     }
     sourceSets["commonMain"].dependencies {
         implementation(compose.components.resources)
+    }
+    sourceSets["wasmJsTest"].dependencies {
+        implementation(kotlin("test"))
+        implementation(libs.kotlinx.coroutines.test)
     }
     sourceSets["wasmJsMain"].dependencies {
         // Compose
--- a/frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/viewmodel/ServiceHolder.kt
+++ b/frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/viewmodel/ServiceHolder.kt
@@ -16,21 +16,66 @@
 package com.monkopedia.konstructor.frontend.viewmodel
 
 import com.monkopedia.konstructor.common.Konstructor
+import com.monkopedia.konstructor.frontend.threejs.consoleError
+import com.monkopedia.ksrpc.channels.Connection
 import com.monkopedia.ksrpc.ksrpcEnvironment
 import com.monkopedia.ksrpc.ktor.websocket.asWebsocketConnection
 import com.monkopedia.ksrpc.toStub
 import io.ktor.client.HttpClient
 import io.ktor.client.plugins.websocket.WebSockets
 import kotlinx.browser.window
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.MainScope
+import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 
-class ServiceHolder {
+/**
+ * How long to wait between keep-alive pings on an established connection.
+ */
+private const val KEEP_ALIVE_INTERVAL_MS = 10_000L
 
-    private val scope = kotlinx.coroutines.MainScope()
+/**
+ * How long to wait before starting the next connection attempt. Deliberately a
+ * flat interval — changing the retry cadence (backoff/jitter/caps) is a policy
+ * question tracked separately from the resource handling here.
+ */
+private const val RECONNECT_DELAY_MS = 3_000L
+
+/**
+ * One attempt's worth of connection state: the stub callers talk to, plus every
+ * resource that had to be allocated to produce it.
+ *
+ * [close] exists so the reconnect loop can release an attempt's resources
+ * instead of abandoning them; a [ServiceConnection] that is dropped without
+ * being closed is a leak of an [HttpClient] (which owns an engine and its own
+ * coroutine scope) and of a live websocket.
+ */
+internal interface ServiceConnection {
+    val stub: Konstructor
+
+    suspend fun close()
+}
+
+/**
+ * Establishes a [ServiceConnection]. Exists as a seam so the reconnect loop's
+ * resource handling can be observed in tests without a real websocket.
+ */
+internal fun interface ConnectionFactory {
+    suspend fun connect(): ServiceConnection
+}
+
+class ServiceHolder internal constructor(
+    private val connectionFactory: ConnectionFactory,
+    private val scope: CoroutineScope
+) {
+
+    constructor() : this(WebSocketConnectionFactory(), MainScope())
 
     private val _service = MutableStateFlow<Konstructor?>(null)
     val service: StateFlow<Konstructor?> = _service.asStateFlow()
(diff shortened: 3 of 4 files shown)


monkopedia-reviewer (approved):
## Independent review — APPROVE

Reviewed at head `fc5a248` in a clean worktree off the PR head. I did not write this and took nothing in the description as established; every number below is one I measured myself.

---

### 1. The leak fix is verified by measurement, and the measurement scales

"`close()` is now called" is a claim about source, so I reproduced the author's mutation independently: I dropped **only** `connection?.closeQuietly()` from the `finally` (API left intact, so the tests still compiled — `41/42 actionable tasks executed`, not a `tests=0` compile failure), and I added my own throwaway probe that reports the raw counters at several horizons instead of asserting on them.

The critical question was whether the leak count actually **tracks N** — a fixed "11" that did not would mean the test was measuring something other than accumulation. It tracks:

| scenario | virtual ms | RED `live` | GREEN `live` |
|---|---|---|---|
| every attempt fails its first ping | 30 001 | **11** | 0 |
| " | 60 001 | **21** | 0 |
| " | 120 001 | **41** | 0 |
| connect, then keep-alive drops | 130 001 | **11** | 1 |
| " | 260 001 | **21** | 1 |

RED grows exactly one connection per `RECONNECT_DELAY_MS` (3 s) on the failure path and one per 13 s on the keep-alive path — linear in N, `closed=0` throughout. GREEN is **flat regardless of N**: 0 stranded on the failure path, and exactly 1 live on the keep-alive path (the currently-established connection, which is correct). That is a genuine accumulation measurement, not a snapshot.

The three shipped tests reproduce the author's table exactly under the same mutation:

- `failedAttemptsDoNotAccumulateOpenConnections` — `every allocated connection should have been closed. Expected <11>, actual <0>.`
- `droppedKeepAliveDoesNotAccumulateOpenConnections` — `expected <= 1 open connection, 11 of 11 were allocated and never closed`
- `cancellingTheScopeClosesTheLiveConnection` — `cancellation must not strand the open connection. Expected <0>, actual <1>.`

GREEN on the pristine head: `tests="3" skipped="0" failures="0" errors="0"`, ChromeHeadless 151, XML deleted before the run and re-created fresh, `41 actionable tasks: 41 executed`.

### 2. Test soundness

All three tests carry their own positive control, so none of them can pass vacuously: test 1 asserts `created >= 10` before comparing `created`/`closed` (0 == 0 would otherwise pass), test 2 asserts `created >= 5`, test 3 asserts `created == 1 && live == 1` *before* cancelling. `CountingConnection` also `check(!isClosed)`s, so a double close can't inflate `closed` into a false green. The `advanceTimeBy` + `runCurrent` choice over `advanceUntilIdle` is correct for a holder on `backgroundScope` and is documented in a comment at the point it matters.

**The instrument can fail.** I didn't take that on faith either — my probe class produced 5 real failures alongside the 3 mutated ones (8 failures, `BUILD FAILED`, `GRADLE_EXIT=1`), so the new source set genuinely reports red rather than silently reporting nothing.

### 3. Scope — the line held

The cadence values are **unchanged**; the constants are a pure rename:

```
main:  delay(10_000)   ->  KEEP_ALIVE_INTERVAL_MS = 10_000L
main:  delay(3_000)    ->  RECONNECT_DELAY_MS     =  3_000L
```

Nothing cadence-related shipped here. #135 exists, is OPEN and `needs-decision`, and lays out the four options honestly. Points 1, 3 and 4 of #127 are addressed; point 2 is the one deliberately left out — exactly the split the triage note asked for.

The second product-visible finding is real and correctly deferred: `break` on keep-alive failure exits the inner loop **without an exception**, so the `_service.value = null` / `_connected.value = false` at lines 117-118 never run and the UI reports "connected" over a dead stub for the retry gap. It is recorded in #135's "Related" paragraph, not silently dropped.

### 4. Scope expansion — justified and minimal

`:frontend` had no test source set at all, so a regression test for this defect required building the infrastructure first. What was added is the minimum to do that: one `testTask { useKarma { useChromeHeadless() } }` block, two test dependencies, and one CI step. No unrelated refactors rode along. The seam itself (`ServiceConnection` / `ConnectionFactory`) is the smallest thing that makes "was this released?" observable without a real websocket, and the public API is unchanged — `ServiceHolder()` in `di/AppModule.kt` still resolves via the secondary constructor.

### 5. The CI step — the highest-risk element — is now PROVEN on a runner

This was the one thing I would have blocked on. It is no longer unproven. Run **33007545754** on `fc5a248`: `status: completed`, `conclusion: success`, `jobs.total_count: 1`, every step `success` including e2e.

The `CHROME_BIN` resolution that had never executed on a runner resolved on the first alternative:

```
Using CHROME_BIN=/usr/bin/google-chrome
```

And it was not a vacuous pass — the JUnit XML **uploaded from the runner** carries a real count:

```
testsuite name="...ServiceHolderConnectionLeakTest" tests="3" skipped="0" failures="0" errors="0"
```

`shadowJar` also succeeded on the runner, which independently confirms `:frontend:wasmJsBrowserDistribution` still builds with the new source set present. The last six CI runs on `main` are all green, so this run is a clean signal rather than noise.

---

### Non-blocking nits

1. **Seven opt-in warnings** in the new test file (`advanceTimeBy`, `runCurrent`, `backgroundScope` need `@OptIn(ExperimentalCoroutinesApi::class)`). Harmless, but this repo files backlog issues for warning accumulation, so worth a class-level `@OptIn` at some point.
2. **`./gradlew check` / `build` now needs a browser locally.** `wasmJsBrowserTest` attaches to `check`, so a dev without Chrome/Chromium gets a new failure mode on a task that used to pass. CI is unaffected (it calls the explicit tasks) and `shadowJar` is unaffected, so this is a developer-ergonomics note only.
3. **`WebSocketConnectionFactory.connect()`** closes `client` if `toStub()` throws but does not close `conn` on that path. `client.close()` tears the engine and its websocket down, so this is not a leak in practice — mentioning it only because the rest of the file is scrupulous about exactly this.
4. Closing the discarded client slightly **sharpens** the pre-existing stale-stub bug: during the retry gap the stub in `_service` is now backed by a definitively closed client rather than an open-but-dead one, so calls fail fast instead of hanging. That is arguably the better behaviour, and it is the same window #135 covers — but it is a small behaviour change worth being aware of when #135 is decided.

### Could not verify

Nothing material. The only claim I did not independently re-derive is the author's original "2 tests completed, 1 failed" probe — my own probe demonstrated the same property (the source set can go red) more thoroughly, so the claim is superseded rather than unchecked.

**Approving.** The defect is real, the fix is minimal and correct, the RED is genuine and scales with the number of reconnect attempts, the scope line on the cadence was respected, and the one element that carried real risk to `main` has now run green on a real runner with a non-zero test count.

PR #138, opened 2026-09-04
Report a failed render as RENDER_FAILED rather than CLEAN (#117)

### LINKED ISSUE #117: info.dirtyState is set CLEAN after every render, even one that failed

monkopedia-coder:
## What

`KonstructionServiceImpl.konstruct()` wrote `dirtyState = CLEAN` after every render, including one that failed (`KonstructionServiceImpl.kt:299` on main). Since #113 / #123 the **per-target** states are accurate, so nothing was stuck — but the konstruction-level flag claimed "clean" about a konstruction whose last render had just failed, and `ScriptHostImpl.findScript()` reads exactly that flag to decide whether an imported konstruction still needs compiling.

## The two traps the issue named, and how this avoids them

**1. "Just report NEEDS_EXEC" would loop the frontend forever.** `KonstructionViewModel.onInfoChanged()` fires `requestKonstructs(...)` whenever it sees `NEEDS_EXEC`, so a permanently-failing target would drive render → fail → info change → re-request, without end. The unconditional `CLEAN` was load-bearing, not sloppy.

So this adds a **fourth `DirtyState`: `RENDER_FAILED`** — appended last, so every existing entry keeps its ordinal and serial name. It means *"the build ran and did not work"* and asks nobody to run it again. The frontend treats it **exactly** as it treated the `CLEAN` that used to be reported on this path: settle the UI, show whatever did build, do not auto-retry. Frontend behaviour after a failed render is therefore unchanged by this PR; only the backend's honesty about it changes.

**2. "Dirty if any target is dirty" would pin the flag forever** for a script with a declared-but-never-exported target. Avoided by keying on what the render **reported** (`lastRenderResult().status`, which `ExecuteTask` fails as soon as any *attempted* target errors) rather than on whether any target is dirty. A never-attempted target cannot make the konstruction `RENDER_FAILED`.

## The per-target logic from #113 / #123 is untouched

No line of the `when` that derives per-target state was changed. The retry route is unchanged and still the per-target `NEEDS_EXEC` that the second clause of the `konstruct()` guard reads (#104) — `KonstructionServiceDirtyStateTest` asserts both explicitly, so a future change that quietly undoes either fails here.

## One intentional behaviour change

`ScriptHostImpl.findScript()` gates on `dirtyState != CLEAN`, so it now sees `RENDER_FAILED` and recompiles an imported konstruction whose last render failed (it previously did not, because the flag lied). That gate is precisely the consumer the issue cited as being misled; recompiling is the conservative answer to "the last build was not good". Flagged here because it is the one place where reading the truth changes what happens rather than only what is reported.

## Test, and the RED control

`backend/src/jvmTest/.../KonstructionServiceDirtyStateTest.kt` drives a render through a fake `KonstructionController` that reproduces `ExecuteTask`'s contract (built targets in `taskArguments`, FAILURE as soon as one attempted target errors), so it runs in the **ordinary `:backend:jvmTest`** rather than behind `-Dintegration=true`.

Every case asserts `renderCalls` as well as the state. The `konstruct()` guard can skip `render()` **silently**, and dirty-state tests in this repo have passed vacuously that way before, so "the render actually ran" is asserted rather than assumed.

Measured, `:backend:jvmTest` in full (no `--tests` filter), on kaladin under the build slot:

| | tests | failed | skipped |
|---|---|---|---|
| **RED** — only `dirtyState = if (...) CLEAN else RENDER_FAILED` reverted to `dirtyState = CLEAN` | 164 | **2** | 19 |
| **GREEN** — fix in place | 164 | **0** | 19 |

The RED failures are `failedRenderDoesNotLeaveTheKonstructionClean` and `aFailedRenderCanStillBeRetried`; the other two cases in the class stay green, so the test is not merely "always fails". The mutation is one expression — the enum, the test, and every signature stayed intact, and the RED run compiled and executed 164 tests rather than dying at compile time.

`:protocol:jvmTest` (17 tests, 0 failed), `:frontend:compileKotlinWasmJs` and `spotlessCheck` all pass on the fix.

## Scope

Deliberately does **not** touch #115 (LSP respawn cap — different file entirely) or #116 (the `onRenderChanged` broadcast loop and `render()`'s STL-deletion set). #116's territory is `KonstructionServiceImpl.kt:303-317` plus `KonstructionController.kt:128`; neither is modified here. The two changes are adjacent inside `konstruct()` but do not overlap.

Not fixed, noted: `compile()` has the same shape of lie — a **compile** failure also writes `dirtyState = CLEAN` (`:206`). Out of scope for this issue, which is about the render path.

Fixes #117

### DIFF
--- a/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionServiceImpl.kt
+++ b/backend/src/jvmMain/kotlin/com/monkopedia/konstructor/KonstructionServiceImpl.kt
@@ -23,6 +23,7 @@ import com.monkopedia.konstructor.common.DirtyState
 import com.monkopedia.konstructor.common.DirtyState.CLEAN
 import com.monkopedia.konstructor.common.DirtyState.NEEDS_COMPILE
 import com.monkopedia.konstructor.common.DirtyState.NEEDS_EXEC
+import com.monkopedia.konstructor.common.DirtyState.RENDER_FAILED
 import com.monkopedia.konstructor.common.KonstructionCallbacks
 import com.monkopedia.konstructor.common.KonstructionCallbacks.CONTENT_CHANGE
 import com.monkopedia.konstructor.common.KonstructionCallbacks.DIRTY_CHANGE
@@ -270,7 +271,8 @@ class KonstructionServiceImpl(
                 info.targets.any { it.name in targets && it.state == NEEDS_EXEC }
             ) {
                 val rendered = konstructionController.render(targets)
-                val latestBuilt = konstructionController.lastRenderResult().taskArguments
+                val lastRender = konstructionController.lastRenderResult()
+                val latestBuilt = lastRender.taskArguments
                 val existingTargets = info.targets.associateBy { it.name }
                 val changedTargets = mutableListOf<KonstructionTarget>()
                 val targets = rendered.allTargets.map { target ->
@@ -296,7 +298,18 @@ class KonstructionServiceImpl(
                     } ?: KonstructionTarget(target, dirtyState).also(changedTargets::add)
                 }
                 val newInfo = info.copy(
-                    dirtyState = CLEAN,
+                    // What the render reported, not an unconditional CLEAN. `ExecuteTask`
+                    // fails the whole render when any attempted target errors, so this says
+                    // "the last build worked" / "it did not" — and a konstruction whose
+                    // render just failed no longer reads CLEAN to `ScriptHostImpl.findScript`
+                    // or to anything else treating the flag as "last build was good" (#117).
+                    //
+                    // Deliberately NOT NEEDS_EXEC: that is the frontend's auto-build trigger
+                    // (`KonstructionViewModel.onInfoChanged`), so a permanently-failing target
+                    // would loop render -> fail -> re-request. RENDER_FAILED carries the
+                    // failure without asking for a retry; the retry route is the per-target
+                    // NEEDS_EXEC above, which the second clause of the guard reads (#104).
+                    dirtyState = if (lastRender.status == SUCCESS) CLEAN else RENDER_FAILED,
                     targets = targets
                 )
                 konstructionController.info = newInfo
--- a/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/KonstructionServiceDirtyStateTest.kt
+++ b/backend/src/jvmTest/kotlin/com/monkopedia/konstructor/KonstructionServiceDirtyStateTest.kt
@@ -1,0 +1,278 @@
+/*
+ * Copyright 2022 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.monkopedia.konstructor
+
+import com.monkopedia.hauler.CallSign
+import com.monkopedia.konstructor.common.DirtyState.CLEAN
+import com.monkopedia.konstructor.common.DirtyState.NEEDS_EXEC
+import com.monkopedia.konstructor.common.DirtyState.RENDER_FAILED
+import com.monkopedia.konstructor.common.Konstruction
+import com.monkopedia.konstructor.common.KonstructionInfo
+import com.monkopedia.konstructor.common.KonstructionTarget
+import com.monkopedia.konstructor.common.TaskResult
+import com.monkopedia.konstructor.common.TaskStatus
+import com.monkopedia.konstructor.logging.WarehouseWrapper
+import com.monkopedia.konstructor.testutil.TestEnvironment
+import io.ktor.utils.io.ByteReadChannel
+import java.io.File
+import kotlin.test.AfterTest
+import kotlin.test.BeforeTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotEquals
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.sync.Mutex
+
+/**
+ * The konstruction-level [KonstructionInfo.dirtyState] after a render (#117).
+ *
+ * `konstruct()` used to write `dirtyState = CLEAN` unconditionally, so a konstruction whose
+ * render had just failed reported itself clean — misleading on its own, and actively wrong for
+ * [com.monkopedia.konstructor.hostservices.ScriptHostImpl.findScript], which reads the flag as
+ * "the last build was good" to decide whether an imported konstruction needs recompiling.
+ *
+ * The render is driven through a [FakeController] rather than kotlinc plus a script
+ * subprocess: the behaviour under test lives entirely in how [KonstructionServiceImpl] turns
+ * what a render reported into dirty state, and the fake reproduces exactly the contract
+ * [com.monkopedia.konstructor.tasks.ExecuteTask] produces (built targets in
+ * [TaskResult.taskArguments], FAILURE as soon as any attempted target errors). That keeps
+ * these in the ordinary `:backend:jvmTest` run instead of behind `-Dintegration=true`.
+ *
+ * Every case asserts [FakeController.renderCalls] as well as the state. Dirty-state tests in
+ * this repo have passed vacuously before by never entering `render()` at all — the `:263`
+ * guard skips it silently — so "the render actually ran" is part of what is being asserted,
+ * not an assumption.
+ */
+class KonstructionServiceDirtyStateTest {
+
+    private lateinit var env: TestEnvironment
+    private lateinit var controller: FakeController
+    private lateinit var service: KonstructionServiceImpl
+
+    @BeforeTest
+    fun setUp() {
+        env = TestEnvironment()
+        env.createWorkspaceDir("ws1", "Test")
+        controller = FakeController(
+            config = env.config,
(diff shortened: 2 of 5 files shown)


monkopedia-reviewer (approved):
**APPROVE.** Independent review at `7ef50e5f`, in a private worktree off the PR head. The fix is correct and minimal, the RED control reproduces exactly as claimed, and the enum addition is safe at every consumer I could find. Two things below are non-blocking but should not be lost: the `ScriptHostImpl` consequence needs to be stated **in that file**, and there is an unmentioned rollback hazard.

## Numbers I measured myself

| run | tests | failed | skipped |
|---|---|---|---|
| **GREEN** `:backend:jvmTest` (full, no filter) | 164 | **0** | 19 |
| **GREEN** `:protocol:jvmTest` | 17 | **0** | 0 |
| **RED** — only `dirtyState = if (lastRender.status == SUCCESS) CLEAN else RENDER_FAILED` → `dirtyState = CLEAN` | 164 | **2** | 19 |

RED failures: `failedRenderDoesNotLeaveTheKonstructionClean`, `aFailedRenderCanStillBeRetried` — the other two cases in the class stayed green, so the test is not merely "always red". The mutation was one expression; the enum, the imports and every signature stayed intact, so RED **compiled and executed 164 tests** rather than dying at compile time. XML deleted before each run; `55 actionable tasks: 55 executed`. `spotlessCheck`: `20 actionable tasks: 20 executed`, clean, on the restored tree (`git status --porcelain` empty first).

The test is not vacuous: every case asserts `FakeController.renderCalls`, so "the render actually ran" is measured, not assumed — which matters given this repo's history of dirty-state tests passing while `konstruct()` silently skipped `render()`.

## Consumer enumeration — 10 files reference `dirtyState`; 4 production read sites

| site | shape | effect of the new case |
|---|---|---|
| `KonstructionServiceImpl:200` | `== NEEDS_COMPILE` (compile guard) | RENDER_FAILED behaves exactly as the CLEAN it replaces — body skipped either way. No change. |
| `KonstructionServiceImpl:270` | `== NEEDS_EXEC` (konstruct guard) | Same. Retry rides the second clause (per-target `NEEDS_EXEC`), asserted by `aFailedRenderCanStillBeRetried`. |
| `KonstructionViewModel:231, :320` | `in SETTLED_STATES` / `when` | Handled by this PR. |
| `ScriptHostImpl:90` | **`!= CLEAN`** | **The one site whose meaning changes.** Below. |

Note for future enum work: the frontend `when` ends in `else -> { /* no-op */ }`, so it would **not** have failed to compile if it had been missed — RENDER_FAILED would have fallen silently into the no-op and the UI would never have called `applyCleanRenders`. The PR got this right, but `DirtyState` has no compile-time exhaustiveness guard anywhere in the tree.

**No untouched test was encoding the bug.** The four untouched `dirtyState` tests (`KonstructionControllerConcurrencyTest:150-163`, `KonstructionServiceListenerTest:109`, `ListenerHandlerTest:101`, `KonstructionLifecycleTest:83`) all assert `NEEDS_COMPILE`/`NEEDS_EXEC` on paths this PR does not touch; none asserts CLEAN after a failed render, so there was nothing to invert. `FailedRenderRetryTest` — the #104/#113/#123 test — never reads konstruction-level `dirtyState` at all, only per-target `state`, which is why the re-landed per-target guarantees are genuinely untouched. The only test edit is `3` → `4` in `testDirtyStateValues` plus an **added** ordinal assertion; nothing was deleted.

I also confirmed the "no dead end" invariant at the source rather than taking the PR's word: `ExecuteTask` sets `isSuccessful = isSuccessful && targetBuilt` and puts only built targets in `builtTargets`, so **FAILURE ⇒ builtTargets ⊊ attemptedTargets ⇒ at least one target is left `NEEDS_EXEC`**. RENDER_FAILED therefore always ships with a live retry route; the issue's "pinned forever" trap is really avoided, not just asserted.

## `ScriptHostImpl` — I measured it rather than reasoning about it

The PR body **does** name this change, which I credit. What follows is what the body does not say.

I installed a fake controller for an imported konstruction and called `findScript` at each state:

- `dirtyState == CLEAN` → **0 compiles** (the old behaviour after a failed render).
- `dirtyState == RENDER_FAILED`, compile succeeds → **1 compile**, gate passes, script returned.
- `dirtyState == RENDER_FAILED`, compile fails → **1 compile**, then `findScript` hard-errors with `does not compile successfully`.

**Verdict: a cost, not a break — and where it does break, the break is the better behaviour.**

To reach `RENDER_FAILED` the source must have compiled successfully and not changed since (`set()` rewrites the state to `NEEDS_COMPILE`, and a failed service `compile()` writes CLEAN), so a recompile of that same source succeeds and the gate passes. I could not construct a non-racy path to the hard error. The one path I could construct is the lost update in `konstruct()`: it writes `info.copy(...)` from a snapshot taken **before** `render()`, so a `set()` landing mid-render is overwritten. In that window the state says RENDER_FAILED while the content on disk no longer compiles — and `findScript` now errors. Before this PR the same race wrote CLEAN, and `findScript` silently handed back a stale build of source that no longer compiles. **Erroring is the better of the two.**

The part worth putting in the code:

**`RENDER_FAILED` is the first `DirtyState` that is both `!= CLEAN` and stable indefinitely.** `NEEDS_COMPILE` and `NEEDS_EXEC` are transient — the frontend drives them to CLEAN. `RENDER_FAILED` is terminal *by design* (that is the whole point of not reusing `NEEDS_EXEC`). And `KonstructionController.compile()` does not touch `info`, which I verified by asserting the state is still `RENDER_FAILED` after `findScript` returns. So the `!= CLEAN` gate acquires a **permanent recompile** steady state it has never had: a konstruction whose render failed pays a full kotlinc compile on **every** cross-script import resolution, for as long as it stays broken. That is defensible — arguably correct, since a stale artifact is worse — but it is a real cost that arrives as a side effect of the enum, and it is invisible to anyone reading `ScriptHostImpl.kt`.

**Please add a comment at `ScriptHostImpl.kt:90`** saying that `RENDER_FAILED` deliberately falls on the recompile side of this gate and that the recompile is permanent while the state holds. A PR description stops being reachable once this merges; the next person to read that line will see a bare `!= CLEAN` and no reason. A test pinning `findScript` under `RENDER_FAILED` would be better still — this is the consumer #117 was filed about, and it is the only consumer this PR changes without covering.

Minor, same area: the `DirtyState.RENDER_FAILED` KDoc's "without asking anyone to run it again" is exactly true of the *render* loop it is about, but `findScript` now does ask for a *compile*. Worth a half-sentence so the KDoc is not read as a broader promise than it makes.

## Wire compatibility — verified, plus one hazard the PR does not mention

Verified by probe rather than by inspection:

- The three existing serial names decode unchanged (`CLEAN`, `NEEDS_COMPILE`, `NEEDS_EXEC`). No `@SerialName` anywhere in `:protocol`, and **no ordinal-dependent code anywhere in the tree** (`grep` for `.ordinal` / `values()[…]` finds nothing outside the count assertion). The "appended last" claim holds.
- No existing persisted `info.json` can contain the new name, so **old data → new code is safe**, which is the direction the PR body checks.

**The direction it does not check is the rollback.** `Config.json` is `Json { ignoreUnknownKeys = true }`, and I measured that this ignores unknown *keys*, not unknown *enum values* — decoding `"dirtyState":"<unknown>"` throws `SerializationException`, and `dirtyState` has no default so no coercion could rescue it. Once a new backend writes `"RENDER_FAILED"` into any `info.json`, a rollback to the previous backend makes `WorkspaceImpl.list()` — which decodes **every** konstruction's `info.json` in the workspace — throw, so the whole workspace listing fails, not just that one konstruction.

Not blocking: it needs a failed render to have happened, and it is recoverable by editing one JSON value. But #113 was reverted once in this exact area, so a rollback here is not hypothetical, and the compatibility note in the KDoc currently states only the safe direction. Please add the forward direction to it.

Same mechanism, much narrower: a browser tab loaded before a redeploy receiving `RENDER_FAILED` over ksrpc will throw in the decode. Self-healing on reload; mentioning it for completeness only.

## What I could not verify

- Nothing about the LSP: the four `This build of intellij-server has expired` lines in every run are the test's own fake, not a real engine.
- I did not run the `-Dintegration=true` suite or e2e; CI's `build-and-test` is green on this head, and no e2e test or the TestBridge snapshot references `DirtyState` at all.
- I did not exercise the `konstruct()` lost-update race directly — that path is stated as constructed-from-the-code, not measured.

PR #141, opened 2026-09-05
e2e: one Playwright driver + browser per test class, not per test method (#124)

### LINKED ISSUE #124: e2e harness leaks a Playwright driver + headed Chromium per test method; the finalize() that was meant to clean them up can never run
**Category:** test-gap (test-infrastructure resource leak)

**Where:** `e2e/src/test/kotlin/com/monkopedia/konstructor/e2e/BaseE2eTest.kt:39-40, 44-56, 68-78`

**Current:**

```kotlin
abstract class BaseE2eTest {
    private var playwright: Playwright? = null     // ← INSTANCE fields
    private var browser: Browser? = null

    @org.junit.Before
    fun setUpBase() {
        server = ServerFixture(); server.start()
        if (playwright == null) {                  // ← therefore ALWAYS true
            playwright = Playwright.create()
            browser = playwright!!.chromium().launch(
                BrowserType.LaunchOptions().setHeadless(false)
            )
        }
        page = browser!!.newPage()
        ...
    }

    @org.junit.After
    fun tearDownBase() {
        if (::page.isInitialized) page.close()
        server.stop()                              // ← browser / playwright NOT closed
    }

    // Clean up Playwright at end (called by JVM shutdown, not per-test)
    protected fun finalize() {
        browser?.close()
        playwright?.close()
    }
}
```

Three things stack up:

1. **`e2e` runs on JUnit 4** (`e2e/build.gradle.kts`: `testImplementation(kotlin("test-junit"))`, `tasks.test { … useJUnit() }`). JUnit 4 constructs a **new instance of the test class for every `@Test` method**, so `playwright` and `browser` are freshly `null` each time and the `if (playwright == null)` guard — clearly written to make them one-per-class — never guards anything. Every test method does a fresh `Playwright.create()` (which spawns the Node driver process) plus a fresh **headed** Chromium launch.

2. **Nothing closes them.** `tearDownBase()` closes only the page and the server.

3. **The comment on the cleanup is factually wrong.** `finalize()` is not "called by JVM shutdown" — the JVM has run no finalizers at shutdown since `runFinalizersOnExit` was removed, and finalization is deprecated for removal (JEP 421). It runs, if ever, at an unspecified GC. So the one mechanism the class relies on for cleanup is inert, and the comment is what stops anyone looking again. (This is the "a justifying comment is not a justification" shape — the comment was load-bearing for the omission.)

**Measured scale** (`origin/main`, denominators included):

```
$ git grep -l 'BaseE2eTest()' -- e2e/src | wc -l                 → 18 subclasses
$ git grep -l 'BaseE2eTest()' -- e2e/src | xargs grep -c '@Test' \
    | awk -F: '{s+=$2} END {print s}'                            → 42 @Test methods
$ git grep -n 'Playwright.create\|chromium()\|finalize' -- e2e/src
    BaseE2eTest.kt:49, BaseE2eTest.kt:52, BaseE2eTest.kt:75      → 3 hits, all in this one file
$ git grep -n '@BeforeClass\|@AfterClass\|ClassRule' -- e2e/src  → 0 hits
```

So a full `:e2e:test -Pe2e` run starts **42 Playwright driver processes and 42 headed Chromium browsers and closes none of them**, holding them until the Gradle test JVM exits.

**Why this is not merely tidiness:** `.github/workflows/ci.yaml` already carries a workaround pointed straight at the symptom —

```yaml
      # The Ktor CIO / Netty server and WebSocket use a lot of file descriptors
      # during the e2e runs; bump the limit.
      - name: Raise file descriptor limit
        run: ulimit -n 4096 || true
```

The comment blames the server. 42 unclosed browsers are at least a co-author, and the leak is a **floor that grows with every e2e test added** — the suite has grown steadily and each new `@Test` adds one more browser. A harness that exhausts fds is a false-green/flake vector, not hygiene: it corrupts the evidence of every other e2e-verified fix.

**Proposed:**
- Move `playwright` / `browser` to a `companion object`, created once in `@BeforeClass` and closed in `@AfterClass` (or a JUnit `ClassRule`, or a single process-wide holder with a shutdown hook if you'd rather share one browser across all classes). Keep `page = browser.newPage()` per test — that part is already right, and `page.close()` in `tearDownBase` already handles it.
- **Delete `finalize()`** and its comment outright rather than repairing it.

**Acceptance evidence worth capturing** (so this doesn't become another inert guard): before the fix, run two `@Test` methods from one class and count `Playwright.create()` calls / driver child processes; after, it must be 1 regardless of method count. A log line or a counter in `setUpBase` is enough to make it visible.

**Effort:** small refactor (one file, ~15 lines).

monkopedia-coder:
## What

`BaseE2eTest` held `playwright` / `browser` in **instance** fields. JUnit 4 constructs a new instance of the test class for every `@Test` method, so the `if (playwright == null)` guard — clearly written to make them one-per-class — was always true, and every test method spawned its own Node driver process plus its own **headed** Chromium. Nothing closed them: `tearDownBase()` closes only the page and the server, and the `finalize()` that carried the comment *"called by JVM shutdown"* is not called by JVM shutdown at all (finalization is deprecated for removal, JEP 421; it runs at an unspecified GC, if ever). The comment was load-bearing for the omission.

- Moved `playwright` / `browser` to the companion object, created in `@BeforeClass`, closed in `@AfterClass`.
- Deleted `finalize()` and its comment.
- `page = browser.newPage()` stays per-method: `Browser.newPage()` creates its own browser context, so **per-test isolation (localStorage etc.) is unchanged**.
- Added `HarnessLifecycleTest` as the regression guard.
- **Deleted the `Raise file descriptor limit` CI step** (see below).

## The control — RED and GREEN, measured

The defect is *resource retention across test methods*, which a single-method check cannot see: it passes identically before and after. So `HarnessLifecycleTest` has three methods that each count **real OS processes** — `ProcessHandle.current().descendants()` filtered to the Playwright `run-driver` node process — and assert exactly one live driver.

Run locally under Xvfb (`:e2e:test -Pe2e --tests "*HarnessLifecycleTest"`), same test file both times, only `BaseE2eTest` differing:

| | method 1 | method 2 | method 3 | result |
|---|---|---|---|---|
| **RED** (harness as on `main`) | drivers **1**, browser procs 8 | drivers **2**, browser procs 14 | drivers **3**, browser procs 20 | `3 tests completed, 2 failed` |
| **GREEN** (this branch) | drivers **1**, browser procs 8 | drivers **1**, browser procs 8 | drivers **1**, browser procs 8 | `tests="3" failures="0"` |

The leak scales exactly with N: **+1 driver and +6 browser processes per test method**, none of them released. Extrapolated over the current 45 `@Test` methods that is ~45 drivers and ~270 browser processes held to the end of the run — but note the 45-method figure is extrapolation; what was measured is the per-method growth rate.

The RED came from running the control against the **unmodified** harness (the API it compiles against is untouched — it only counts processes), so it is a failing test, not a compile failure.

The assertion is `== 1`, never `<= 1`: if the process filter ever stops matching the driver, the count collapses to 0 and the test **fails** rather than passing vacuously. On any mismatch it dumps every JVM descendant command line so the filter can be repaired rather than guessed at.

**Cross-class check** (`--tests "*AppLoadTest" --tests "*HarnessLifecycleTest"`): `HarnessLifecycleTest` ran *after* `AppLoadTest` (timestamps 17:07:32 vs 17:07:27) and still saw exactly 1 driver — so `@AfterClass` really does release the previous class's driver, and the fix is one-at-a-time across the whole run, not merely one-per-class-retained.

**Full suite**, this branch, locally under Xvfb: 20 classes, 48 tests, 9 skipped, **0 failures**, 3m13s.

## The CI step

```yaml
      # The Ktor CIO / Netty server and WebSocket use a lot of file descriptors
      # during the e2e runs; bump the limit.
      - name: Raise file descriptor limit
        run: ulimit -n 4096 || true
```

Each `run:` step is its own shell process, so this raised the limit for a shell that exited immediately — **zero effect on any later step**, including the e2e run it was added for, and `|| true` meant it could never fail visibly. Whatever fd pressure the suite was under has been completely unmitigated the entire time. Deleted rather than repaired: a step asserting a mitigation that does not exist is worse than no step, because the next reader sees fd pressure as handled.

**If a raised limit is actually wanted**, it has to be inside the same step that runs the tests:

```yaml
      - name: Run e2e tests under Xvfb
        run: ulimit -n 4096 && xvfb-run --auto-servernum --server-args='-screen 0 1280x720x24' ./gradlew :e2e:test -Pe2e
```

I did not do that — with the leak fixed the pressure this was aimed at is gone, and inlining it would re-assert a mitigation nobody has measured a need for. Say the word and I will add it.

## Not verified

- Only run on this Linux box (Arch + Mesa EGL override for Xvfb) and not yet in CI; the process-count filter is Linux-specific (`/proc`-backed `ProcessHandle.Info.commandLine()`), which matches the ubuntu-latest runner.
- The ~45-driver whole-suite figure is extrapolated from the measured per-method growth, not counted at the end of a full pre-fix run.

Fixes #124

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,11 +36,6 @@ jobs:
             libhyphen0 \
             libicu74 || true
 
-      # The Ktor CIO / Netty server and WebSocket use a lot of file descriptors
-      # during the e2e runs; bump the limit.
-      - name: Raise file descriptor limit
-        run: ulimit -n 4096 || true
-
       - name: Spotless check
         run: ./gradlew spotlessCheck
 
--- a/e2e/src/test/kotlin/com/monkopedia/konstructor/e2e/BaseE2eTest.kt
+++ b/e2e/src/test/kotlin/com/monkopedia/konstructor/e2e/BaseE2eTest.kt
@@ -36,25 +36,15 @@ abstract class BaseE2eTest {
     protected lateinit var server: ServerFixture
     protected lateinit var page: Page
 
-    private var playwright: Playwright? = null
-    private var browser: Browser? = null
-
     protected val json = Json { ignoreUnknownKeys = true }
 
     @org.junit.Before
     fun setUpBase() {
         server = ServerFixture()
         server.start()
-        if (playwright == null) {
-            playwright = Playwright.create()
-            // Compose/Skiko needs a real display for WebGL rendering.
-            // Use headed mode (Xvfb provides a virtual display in CI).
-            browser = playwright!!.chromium().launch(
-                BrowserType.LaunchOptions()
-                    .setHeadless(false)
-            )
-        }
-        page = browser!!.newPage()
+        // One page (and so one fresh browser context) per test method; the
+        // browser and driver behind it are shared for the whole class.
+        page = browser().newPage()
         page.onConsoleMessage { msg ->
             if (msg.type() == "error" || msg.type() == "warning" || msg.type() == "log") {
                 System.err.println("[browser:${msg.type()}] ${msg.text()}")
@@ -69,12 +59,6 @@ abstract class BaseE2eTest {
     fun tearDownBase() {
         if (::page.isInitialized) page.close()
         server.stop()
-    }
-
-    // Clean up Playwright at end (called by JVM shutdown, not per-test)
-    protected fun finalize() {
-        browser?.close()
-        playwright?.close()
     }
 
     // -- Bridge helpers -------------------------------------------------------
@@ -271,6 +255,45 @@ abstract class BaseE2eTest {
     protected fun waitOpts(timeout: Double) = Page.WaitForSelectorOptions().setTimeout(timeout)
 
     companion object {
+        /**
+         * The Playwright driver and browser are held on the **class**, not on the
+         * instance: JUnit 4 constructs a new test-class instance for every `@Test`
+         * method, so instance fields here meant one Node driver process plus one
+         * headed Chromium per test method, all of them retained for the life of
+         * the test JVM (issue #124). They are created once per test class in
+         * [startBrowser] and closed in [stopBrowser], so at most one of each is
+         * ever alive. [HarnessLifecycleTest] asserts that property.
+         */
+        private var playwright: Playwright? = null
+        private var browser: Browser? = null
+
+        @JvmStatic
+        @org.junit.BeforeClass
+        fun startBrowser() {
+            if (playwright == null) {
+                playwright = Playwright.create()
+                // Compose/Skiko needs a real display for WebGL rendering.
+                // Use headed mode (Xvfb provides a virtual display in CI).
+                browser = playwright!!.chromium().launch(
+                    BrowserType.LaunchOptions()
+                        .setHeadless(false)
+                )
+            }
+        }
+
+        @JvmStatic
+        @org.junit.AfterClass
+        fun stopBrowser() {
+            browser?.close()
+            browser = null
+            playwright?.close()
+            playwright = null
+        }
+
+        private fun browser(): Browser = checkNotNull(browser) {
+            "Browser not started; @BeforeClass startBrowser() did not run"
+        }
+
         /** Default timeout for a generic [waitForState] predicate. */
         const val DEFAULT_STATE_TIMEOUT: Double = 30000.0
 
--- a/e2e/src/test/kotlin/com/monkopedia/konstructor/e2e/HarnessLifecycleTest.kt
+++ b/e2e/src/test/kotlin/com/monkopedia/konstructor/e2e/HarnessLifecycleTest.kt
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2022 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.monkopedia.konstructor.e2e
+
+import java.util.stream.Collectors
+import kotlin.test.assertEquals
+import org.junit.Test
+
+/**
+ * Guards the browser lifecycle of the harness itself (issue #124).
+ *
+ * JUnit 4 builds a **new instance of the test class for every `@Test` method**,
+ * so any Playwright/Browser held in an *instance* field is created once per
+ * method and — unless something closes it — retained for the life of the test
+ * JVM. That leak is invisible to a single-method check: it only shows up as
+ * growth across methods. Hence three methods that each assert the same
+ * invariant.
+ *
+ * The measurement is of **real OS processes**, not of a harness counter: every
+ * `Playwright.create()` spawns a Node driver process as a child of the test JVM,
+ * and the browser it launches is a descendant of that driver. Counting live
+ * descendants is therefore a direct count of leaked drivers/browsers.
+ *
+ * The assertion is `== 1`, never `<= 1`, deliberately: if the command-line
+ * filter below ever stops matching the driver, the count collapses to 0 and the
+ * test fails instead of passing vacuously. A check that cannot fail is not a
+ * check.
+ */
+class HarnessLifecycleTest : BaseE2eTest() {
+
+    @Test
+    fun onlyOneDriverAlive_1() = assertSingleDriverProcess("method 1")
+
+    @Test
+    fun onlyOneDriverAlive_2() = assertSingleDriverProcess("method 2")
+
+    @Test
+    fun onlyOneDriverAlive_3() = assertSingleDriverProcess("method 3")
+
+    private fun assertSingleDriverProcess(label: String) {
+        // Closing a browser/driver is asynchronous; give a leftover process a
+        // moment to actually exit before deciding it leaked. A genuine leak
+        // never drops, so this only removes flake, not the failure.
+        val drivers = awaitSettled { driverCommandLines() }
+        val browsers = browserCommandLines()
+        System.err.println(
+            "[harness-lifecycle] $label: playwright drivers alive=${drivers.size}, " +
+                "browser processes alive=${browsers.size}"
+        )
+        if (drivers.size != 1) {
+            System.err.println("[harness-lifecycle] all JVM descendants:")
+            descendantCommandLines().forEach { System.err.println("[harness-lifecycle]   $it") }
+        }
+        assertEquals(
+            1,
+            drivers.size,
+            "Expected exactly one live Playwright driver process for the whole test " +
+                "class, found ${drivers.size} ($label). More than one means the harness " +
+                "creates a driver per test method and never closes it (issue #124); " +
+                "zero means the driver-detection filter no longer matches."
+        )
+    }
+
+    private fun awaitSettled(timeoutMs: Long = 5000, count: () -> List<String>): List<String> {
+        val deadline = System.currentTimeMillis() + timeoutMs
+        var current = count()
+        while (current.size > 1 && System.currentTimeMillis() < deadline) {
+            Thread.sleep(250)
+            current = count()
+        }
+        return current
+    }
+
+    private fun descendantCommandLines(): List<String> = ProcessHandle.current().descendants()
+        .map { it.info().commandLine().orElse("") }
+        .collect(Collectors.toList())
+
+    /** Node driver processes: `<driver>/node <driver>/package/cli.js run-driver`. */
+    private fun driverCommandLines(): List<String> =
+        descendantCommandLines().filter { it.contains("run-driver") }
+
+    /** Top-level browser processes (renderers/GPU helpers carry `--type=`). */
+    private fun browserCommandLines(): List<String> = descendantCommandLines().filter {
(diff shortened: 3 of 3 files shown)


monkopedia-reviewer (approved):
**APPROVE.** Independent review — I did not write this. I rebuilt the branch in my own worktree at `34733e9` and re-measured every number below rather than taking the PR body's word for it.

## 1. The count actually drops — and it grows without the fix

Every run: worktree at `34733e9`, JDK 21, `xvfb-run` + the local Mesa-EGL override, `./gradlew :e2e:test -Pe2e`. `build/test-results` deleted before each run, `> Task :e2e:test` never `UP-TO-DATE`, counts read from fresh JUnit XML (not the console).

| run | harness | drivers per method | outcome |
|---|---|---|---|
| GREEN | this branch | **1, 1, 1** (chromium procs 8, 8, 8) | `tests="3" failures="0"` |
| RED | `BaseE2eTest.kt` reverted to `b439a48`, test file untouched | **1, 2, 3** (chromium 8, 14, 20) | `tests="3" failures="2"` |

Reproduces the PR's table exactly. The RED **grows across methods**, so it is measuring accumulation, not a wrong constant.

Three points is thin for an extrapolation to ~45, so I added a scratch 10-method class (report-only, no assertions, so nothing short-circuits) and ran it under both harnesses:

```
fixed     drivers  1  1  1  1  1  1  1  1  1  1     chromium  8 ×10
unfixed   drivers  1  2  3  4  5  6  7  8  9 10     chromium  8 14 20 26 32 38 44 50 56 62
```

Exactly `drivers = N`, `chromium = 6N + 2`, with no plateau at N=10. The extrapolation in the PR body is sound: at this suite's 48 `@Test` methods the pre-fix end state is ~45–48 live driver processes and ~290 chromium processes retained to the end of the test JVM. The issue's "42" is the same phenomenon at an earlier method count. Scratch class deleted; my worktree is back to a clean `34733e9`.

**Cross-class, on the fixed harness** (`--tests "*AppLoadTest" --tests "*HarnessLifecycleTest"`): `AppLoadTest` at `17:20:56`, `HarnessLifecycleTest` at `17:21:01` — and still exactly 1 driver in all three methods. `@AfterClass` genuinely releases the previous class's driver; this is one-at-a-time across the run, not one-retained-per-class.

**Full suite, this branch, locally:** `classes=20 tests=48 skipped=9 failures=0 errors=0`, 3m17s, with `spotlessCheck` in the same invocation. `HarnessLifecycleTest` still read 1 driver *inside the full run* — a stronger measurement than any filtered one, since 19 other classes had their turn in the same JVM.

## 2. The anti-vacuity guard holds in practice, not just in the comment

I broke the filter deliberately on the **fixed** harness (`it.contains("run-driver")` → `it.contains("zzz-no-such-driver-zzz")`):

```
[harness-lifecycle] method 1: playwright drivers alive=0, browser processes alive=8
tests="3" skipped="0" failures="3" errors="0"
```

Three failures, not three vacuous passes. `awaitSettled` only loops while `size > 1`, so a 0 returns immediately and `assertEquals(1, 0)` fires. The `== 1`-never-`<= 1` reasoning is correct as written. The mismatch path also dumps every JVM descendant command line, so a future Playwright upgrade that moves the driver argv yields a repairable failure rather than silence. Confirmed the filter matches on this machine: the dump shows `/tmp/playwright-java-*/node /tmp/playwright-java-*/package/cli.js run-driver`.

## 3. No context leak traded for the browser leak

Denominator: all 24 files under `e2e/src/test/kotlin/com/monkopedia/konstructor/e2e/` (21 of them carry `@Test` methods). `newPage` appears **once** in the entire tree (`BaseE2eTest.kt:47`); `newContext` appears **zero** times. No `page.context()`, no `waitForPopup`, no `storageState`, no cookie or service-worker use. One page and one context per test, both released in `tearDownBase`.

That the context is released is verified, not assumed — in `playwright-1.58.0`, `BrowserImpl.newPageImpl` calls `newContext(...)` and stores the result in `PageImpl.ownedContext`, and `PageImpl.close()` invokes `BrowserContextImpl.close()` on it (checked in the bytecode). So `page.close()` per test closes the context too.

## 4. Isolation

`browser.newPage()` gives each test its own context, which is the correct mitigation, and it is genuinely per-test. Belt and braces: `ServerFixture` binds an ephemeral port per test, so each test's page is a *different origin* anyway — the `localStorage` round-trips in `MigrationRegressionStateTest` and `MigrationRegressionTest` could not see each other even if contexts were shared. Nothing in the suite depends on a fresh *browser* rather than a fresh *page*. The full-suite green above is the empirical side of this.

## 5. The `ulimit` deletion — right call, with one correction to the framing

No `ulimit` remains at this head outside the stock `gradlew` wrapper. Deleting an inert mitigation beats keeping it: `|| true` guaranteed it could never report its own failure, and a step naming a protection that does not exist is worse than no step.

But the body's "whatever fd pressure the suite was under has been completely unmitigated the entire time" overstates it, in this PR's favour. `gradlew` itself does `MAX_FD="maximum"` … `ulimit -n $MAX_FD` (`gradlew:50`, `109-115`) — it raises the **soft** limit to the **hard** limit in the shell that launches Gradle, and test workers fork from that tree (524288 on my box). So the deleted step was not merely in the wrong process; it was asking for 4096 inside a tree already running at the hard maximum. Redundant *as well as* misplaced. Worth a line in the body, so the next person who hits fd trouble does not re-add an fd step believing there is none. And the honest consequence still stands: if CI ever does exhaust descriptors it will now fail visibly rather than degrade quietly, which is the right direction.

## Nits — none blocking

1. **`stopBrowser()` is not exception-safe.** If `browser?.close()` throws, `playwright?.close()` is skipped and **both fields stay non-null**, so the next class's `if (playwright == null)` guard silently reuses a half-dead pair and every later class fails confusingly. `try { browser?.close() } finally { browser = null; try { playwright?.close() } finally { playwright = null } }` closes it. (The `@BeforeClass`-throws direction is already fine — `ParentRunner.classBlock` wraps the befores inside the afters, so `@AfterClass` still runs.)

2. **`browserCommandLines()`'s comment is inaccurate on Linux.** It says renderers/GPU helpers "carry `--type=`" and are therefore excluded. Measured here: `ProcessHandle.Info.commandLine()` returns the chromium processes with *no arguments at all* (the dump prints a bare `.../chrome-linux64/chrome`), while `node` and the backend `java` do carry theirs — so the `--type=` exclusion never fires and the reported number is "all chromium processes", not "top-level browsers". Harmless (it is diagnostic, never asserted, and `8 → 14 → 20` is a real per-browser-tree count either way), but the comment describes something that does not happen.

3. Cosmetic: a class whose tests all self-skip via `assumeTrue` (`ManualVerificationTest`, `ScreenshotTest` without `-Pcapture`) now launches a browser in `@BeforeClass` that no test uses. Not a regression — it was launched in `@Before` before — just noting it is not free.

## What I did not verify

- **CI had not reached the e2e step when I published this.** `Spotless check` and `Build shadowJar` are green; the backend/integration test step was still running after ~50 minutes. So the process filter is confirmed on Linux + JDK 21 *on my box only*; ubuntu-latest is still unconfirmed. A mismatch there fails loudly rather than passing vacuously (item 2 above), so the risk is a visible red, not a false green. **Per this repo's merge gate, whoever merges must confirm `gh pr checks 141` is green first — `main` is unprotected, so nothing else will.**
- The ~45-driver whole-suite figure is extrapolated from a measured, strictly linear growth rate at N=10, not counted at the end of a full pre-fix run. I did not run the full suite unfixed.

monkopedia-coder:
**CI is green** — [run 33980291091](https://github.com/Monkopedia/Konstructor/actions/runs/33980291091), `build-and-test` pass in 23m19s, with the `Raise file descriptor limit` step removed.

The control also runs on the runner, which is the part worth checking: from the `junit-xml` artifact, `HarnessLifecycleTest` on ubuntu-latest —

```
tests="3" skipped="0" failures="0" errors="0"
[harness-lifecycle] method 1: playwright drivers alive=1, browser processes alive=8
[harness-lifecycle] method 2: playwright drivers alive=1, browser processes alive=8
[harness-lifecycle] method 3: playwright drivers alive=1, browser processes alive=8
```

The counts are 1 and 8, not 0 — so the `ProcessHandle` command-line filter matches on the runner as well as locally, and the check is live there rather than passing vacuously. Full e2e in CI: 20 classes, 48 tests, 9 skipped, 0 failures.

monkopedia-reviewer (commented):
**Follow-up to my approval: the one caveat I left open is now closed — CI is green, including the e2e step.**

Run [33980291091](https://github.com/Monkopedia/Konstructor/actions/runs/33980291091) completed `success`; every step including `Run e2e tests under Xvfb` passed. I did not stop at the green step — a green step can also mean a task that never ran — so I pulled the `junit-xml` artifact and read the XML:

```
CI e2e: classes=20 tests=48 skipped=9 failures=0 errors=0
```

Byte-for-byte the same shape as my local run (20/48/9/0/0). And `HarnessLifecycleTest` did execute on the runner rather than being filtered out — `tests="3" skipped="0" failures="0" errors="0"`, with its own captured output:

```
[harness-lifecycle] method 1: playwright drivers alive=1, browser processes alive=8
[harness-lifecycle] method 2: playwright drivers alive=1, browser processes alive=8
[harness-lifecycle] method 3: playwright drivers alive=1, browser processes alive=8
```

Two things follow, and the second is the one worth recording:

1. The per-class lifecycle holds on ubuntu-latest exactly as it does locally — flat at 1 driver across methods.
2. **The `run-driver` process filter matches on the runner too, and matched non-vacuously.** This was the portability risk I flagged: a filter tuned to one Playwright/JDK combination could silently stop matching elsewhere. It reported **1**, not 0 — and 0 is the value that would have failed the assertion. So the measurement is pinned on both environments this suite actually runs in, not just on my box.

That was the only outstanding item in my review. Nothing else changes: approval stands, and the three nits (exception-safety in `stopBrowser()`, the inaccurate `--type=` comment, the `gradlew`-already-raises-fds correction to the body's framing) remain non-blocking.

PR #142, opened 2026-09-05
TestBridge: build every snapshot through buildSnapshot (#106)

### LINKED ISSUE #106: TestBridge: refreshKonstructions publishes a snapshot of defaults, so e2e can read state the app is not in
Found during the #103 review. Pre-existing, not introduced by it.

## What happens

`JsBridge.refreshKonstructions` builds an `AppStateSnapshot` from scratch (`JsBridge.kt:458`) instead of from the real view-model state, and it omits most fields:

```kotlin
val snapshot = AppStateSnapshot(
    ready = true,
    connected = ...,
    workspaceCount = ..., workspaceNames = ..., workspaceIds = ...,
    selectedWorkspaceId = wsId,
    codePaneMode = "EDITOR",     // hardcoded
    screen = "main",
    konstructionCount = kons.size,
    konstructionNames = kons.map { it.name }
)
```

`lspEnabled`, `editorTheme` and `keymap` are not passed at all, so they take the **data-class defaults unconditionally** — not just before the settings flow emits, but every time this runs. `codePaneMode` is hardcoded to `"EDITOR"` regardless of the real mode.

The other construction site (`JsBridge.kt:427`) does it correctly, reading `settingsVm.lspEnabled.value` and friends.

## Why this matters more than a stale field

**The bridge is the instrument every e2e test measures through.** A snapshot that reports defaults rather than actual state means an e2e assertion can pass or fail for reasons unrelated to the app: a test that toggles the editor to `SPLIT`, triggers anything calling `refreshKonstructions`, and then asserts `codePaneMode` reads `"EDITOR"` — the value the app is not in.

This is the same family as the assertion-free gates fixed in #82: a check that reports something other than what it claims to measure. It is currently benign only because no test happens to read those fields across a refresh.

Equally wrong before and after #103 — `lspEnabled` reported `false` unconditionally before the default flipped and reports `true` unconditionally now. Confirmed it cannot destabilise the new `testEditorWorksWithLspFlagOff`, which creates its konstruction over ksrpc and reloads, so no `refreshKonstructions` runs after the toggle.

## Direction

Build the snapshot from the same source the `:427` site uses rather than re-listing fields by hand — the bug is the duplication, and any fix that leaves two hand-maintained construction sites will drift again the next time a field is added. A single `buildSnapshot(...)` helper taking the view-models, called from both places, removes the class of defect rather than this instance.

Worth asserting once in e2e that a field NOT related to konstructions (e.g. `keymap` or `editorTheme`) survives a `refreshKonstructions` unchanged — otherwise the next regression here is invisible in exactly the same way.

monkopedia-coder:
`JsBridge.refreshKonstructions` hand-built its own `AppStateSnapshot` instead of going through the existing `buildSnapshot` helper. It hardcoded `codePaneMode = "EDITOR"` and never passed `editorTheme`, `keymap`, `lspEnabled`, `targets` or `diagnostics` at all, so those silently took their data-class defaults on every konstruction create / rename / delete.

## Why this is worth more than its diff size

`globalThis.__konstructor.state` is the **only** channel the Playwright e2e suite has for reading app state — Compose renders to a WebGL canvas, so there are no DOM selectors and the TestBridge *is* the instrument. A snapshot that reports defaults instead of reality lets an e2e assertion pass or fail describing state the app is not in. It is benign today only because no current test reads those fields across a refresh — luck, not safety.

## The fix removes the class, not the instance

Rather than adding the missing fields to the second site (which leaves the next added field free to drift), `refreshKonstructions` now takes `settingsVm` / `konstructionVm` and calls `buildSnapshot`.

Construction-site enumeration, `git grep -n 'AppStateSnapshot('` over the whole repo:

| | before | after |
|---|---|---|
| declaration (`BridgeStateSnapshot.kt:40`) | 1 | 1 |
| construction sites | **2** (`JsBridge.kt:417` correct, `JsBridge.kt:458` hand-built) | **1** (`JsBridge.kt:417`, inside `buildSnapshot`) |

Denominator: 7 `AppStateSnapshot` mentions repo-wide before the change (declaration, 2 constructions, 2 `serializer()` calls, 1 KDoc, 1 return type); no third construction site exists in any module. A new snapshot field now has exactly one place it can be added.

Two smaller behaviours also become correct, both previously wrong at the hand-built site: `screen` is computed rather than hardcoded `"main"`, and an unloaded workspace list reports `workspaceCount = -1` (the documented "unknown" value) instead of `0`.

## RED / GREEN control

New `JsBridgeRefreshSnapshotTest` in `:frontend`'s `wasmJsTest` source set (added in #136, wired into CI). It drives all four fields off **both** their `SettingsViewModel` defaults and their `AppStateSnapshot` defaults, triggers the refresh, and reads the published JSON back out of `globalThis` — the same channel e2e reads, not an internal return value.

Measured on ChromeHeadless 151, `:frontend:wasmJsBrowserTest`, XML deleted before each run and the task confirmed `executed`:

- **GREEN** (fix in place): `6 tests completed, 0 failed` — `JsBridgeRefreshSnapshotTest tests="3" failures="0"`, `ServiceHolderConnectionLeakTest tests="3" failures="0"`. `spotlessCheck` green in the same invocation.
- **RED** (surgical mutation: only the snapshot-construction body reverted to the pre-fix hand-built form, the new signature left intact so it still compiles): `6 tests completed, 2 failed` —
  ```
  refreshKonstructions must publish real view-model state, not defaults.
  Expected <[codePaneMode=SETTINGS, editorTheme=TOMORROW, keymap=EMACS, lspEnabled=false]>,
   actual <[codePaneMode=EDITOR,   editorTheme=DRACULA,   keymap=VIM,    lspEnabled=true]>.
  ```

Note the RED actual: `editorTheme=DRACULA` is not even `SettingsViewModel`'s default (`ONE_DARK`) — it is the `AppStateSnapshot` default. The old snapshot reported a theme the app had never been in.

**Would the RED have passed vacuously?** Yes, had the fields not been driven off their defaults — that is exactly the failure mode this issue is about. Two guards against it:
1. `theInstrumentWouldDetectAFieldReportingItsDefault` asserts each chosen value differs from the default it would otherwise have been reported as, so the behavioural assertions cannot quietly go vacuous if someone changes a constant.
2. An instrument check on `konstructionNames` proves the refresh actually published and reached the konstruction list — so "all defaults" can never be mistaken for "no publish happened". That check **passes in RED**, which is what makes the RED a real red rather than a no-op.

The four fields are compared as one list so a regression reports every field that drifted rather than only the first (the first RED run stopped at `codePaneMode` and never proved the other three were live; the list form does).

## Not done

The issue also suggests an e2e assertion that a konstruction-unrelated field survives `refreshKonstructions`. Deliberately skipped: it would need a full `shadowJar` + Playwright run to verify, and shipping an unrun e2e test is worse than none. The unit test above observes the same value through the same `globalThis.__konstructor.state` channel, runs in ~6s in CI, and gives a measured RED. Happy to add the e2e assertion if you'd rather have it there too.

Fixes #106

### DIFF
--- a/frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/JsBridge.kt
+++ b/frontend/src/wasmJsMain/kotlin/com/monkopedia/konstructor/frontend/JsBridge.kt
@@ -148,7 +148,7 @@ object JsBridge {
                     val workspaceId = args["workspaceId"]!!.jsonPrimitive.content
                     mutations.createKonstruction(workspaceId, name)
                     // Refresh state so konstructions list updates
-                    refreshKonstructions(serviceHolder, spaceListVm)
+                    refreshKonstructions(serviceHolder, spaceListVm, settingsVm, konstructionVm)
                 } catch (e: Exception) {
                     setError("createKonstruction failed: ${e.message}")
                 }
@@ -163,7 +163,7 @@ object JsBridge {
                     val kon = mutations.findKonstruction(wsId, konId)
                     if (kon != null) {
                         mutations.deleteKonstruction(kon)
-                        refreshKonstructions(serviceHolder, spaceListVm)
+                        refreshKonstructions(serviceHolder, spaceListVm, settingsVm, konstructionVm)
                     }
                 } catch (e: Exception) {
                     setError("deleteKonstruction failed: ${e.message}")
@@ -178,7 +178,7 @@ object JsBridge {
                     val konId = args["konId"]!!.jsonPrimitive.content
                     val name = args["name"]!!.jsonPrimitive.content
                     mutations.renameKonstruction(wsId, konId, name)
-                    refreshKonstructions(serviceHolder, spaceListVm)
+                    refreshKonstructions(serviceHolder, spaceListVm, settingsVm, konstructionVm)
                 } catch (e: Exception) {
                     setError("renameKonstruction failed: ${e.message}")
                 }
@@ -441,9 +441,23 @@ object JsBridge {
 
     private var refreshTriggerRef: kotlinx.coroutines.flow.MutableStateFlow<Int>? = null
 
-    private suspend fun refreshKonstructions(
+    /**
+     * Publish a snapshot after a konstruction-list mutation, without waiting for
+     * the reactive `combine` in [install] to catch up.
+     *
+     * Every field comes from [buildSnapshot] (issue #106). This used to hand-build
+     * its own [AppStateSnapshot], which hardcoded `codePaneMode = "EDITOR"` and
+     * left `lspEnabled`/`editorTheme`/`keymap`/targets/diagnostics on their
+     * data-class defaults — so the bridge, which is the only instrument e2e has
+     * for reading app state, reported state the app was not in. Keep the single
+     * construction site: a new snapshot field must only ever be added inside
+     * [buildSnapshot].
+     */
+    internal suspend fun refreshKonstructions(
         serviceHolder: ServiceHolder,
-        spaceListVm: SpaceListViewModel
+        spaceListVm: SpaceListViewModel,
+        settingsVm: SettingsViewModel,
+        konstructionVm: KonstructionViewModel?
     ) {
         // Directly update the state snapshot with current konstruction list
         spaceListVm.refreshWorkspaces()
@@ -454,18 +468,13 @@ object JsBridge {
             if (service != null && wsId != null) {
                 val ws = service.get(wsId)
                 val kons = ws.list()
-                val workspaces = spaceListVm.workspaces.value
-                val snapshot = AppStateSnapshot(
-                    ready = true,
+                val snapshot = buildSnapshot(
                     connected = serviceHolder.connected.value,
-                    workspaceCount = workspaces?.size ?: 0,
-                    workspaceNames = workspaces?.map { it.name } ?: emptyList(),
-                    workspaceIds = workspaces?.map { it.id } ?: emptyList(),
-                    selectedWorkspaceId = wsId,
-                    codePaneMode = "EDITOR",
-                    screen = "main",
-                    konstructionCount = kons.size,
-                    konstructionNames = kons.map { it.name }
+                    workspaces = spaceListVm.workspaces.value,
+                    selectedWsId = wsId,
+                    konNames = kons.map { it.name },
+                    settingsVm = settingsVm,
+                    konstructionVm = konstructionVm
                 )
                 val stateJson = json.encodeToString(AppStateSnapshot.serializer(), snapshot)
                 setState(stateJson)
--- a/frontend/src/wasmJsTest/kotlin/com/monkopedia/konstructor/frontend/JsBridgeRefreshSnapshotTest.kt
+++ b/frontend/src/wasmJsTest/kotlin/com/monkopedia/konstructor/frontend/JsBridgeRefreshSnapshotTest.kt
@@ -1,0 +1,233 @@
+/*
+ * Copyright 2022 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.monkopedia.konstructor.frontend
+
+import com.monkopedia.hauler.Shipper
+import com.monkopedia.konstructor.common.Konstruction
+import com.monkopedia.konstructor.common.KonstructionService
+import com.monkopedia.konstructor.common.Konstructor
+import com.monkopedia.konstructor.common.Space
+import com.monkopedia.konstructor.common.Workspace
+import com.monkopedia.konstructor.frontend.viewmodel.CodePaneMode
+import com.monkopedia.konstructor.frontend.viewmodel.ConnectionFactory
+import com.monkopedia.konstructor.frontend.viewmodel.EditorThemeName
+import com.monkopedia.konstructor.frontend.viewmodel.KeymapName
+import com.monkopedia.konstructor.frontend.viewmodel.ServiceConnection
+import com.monkopedia.konstructor.frontend.viewmodel.ServiceHolder
+import com.monkopedia.konstructor.frontend.viewmodel.SettingsViewModel
+import com.monkopedia.konstructor.frontend.viewmodel.SpaceListViewModel
+import com.monkopedia.konstructor.frontend.viewmodel.WorkspaceMutations
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotEquals
+import kotlin.test.assertNotNull
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.test.advanceUntilIdle
+import kotlinx.coroutines.test.runCurrent
+import kotlinx.coroutines.test.runTest
+import kotlinx.serialization.json.Json
+
+/**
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
## Approve — independently verified

Reviewed at head `40244ea4` in a detached worktree off the PR head (not the author's checkout). Everything below is **measured**, not read off the PR body; where I only reasoned, I say so.

---

### 1. The universal claim in the title — verified

`git grep -c 'AppStateSnapshot('` over tracked files (note: the *type* is `AppStateSnapshot`, the *file* is `BridgeStateSnapshot.kt` — grepping the filename returns a misleading zero):

| | base `6b04a58` | head `40244ea4` |
|---|---|---|
| identifier mentions, repo-wide | 7 | 13 (7 prod + 6 in the new test) |
| `AppStateSnapshot(` occurrences | 3 | 3 |
| — declaration | 1 (`BridgeStateSnapshot.kt:40`) | 1 |
| — **production construction sites** | **2** (`JsBridge.kt:417`, `JsBridge.kt:458`) | **1** (`JsBridge.kt:417`, inside `buildSnapshot`) |
| — test (`AppStateSnapshot()` defaults probe) | 0 | 1 (`…Test.kt:81`) |

`buildSnapshot`: 1 declaration (`JsBridge.kt:393`), **2 call sites** (`:369` reactive combine, `:471` refresh path). No third construction site in any module. The title holds.

### 2. Does a test fail if the hardcode returns? — yes, two, by name

Three runs, `:frontend:wasmJsBrowserTest`, ChromeHeadless 151, `frontend/build/test-results` deleted before each run, task confirmed `executed`.

**GREEN** — `6 tests completed, 0 failed` (`JsBridgeRefreshSnapshotTest tests=3 failures=0`, `ServiceHolderConnectionLeakTest tests=3 failures=0`); 61 tasks executed; `spotlessCheck` green in the same invocation.

**RED-A — the literal mutation**, one line, API untouched (`buildSnapshot(...)` → `buildSnapshot(...).copy(codePaneMode = "EDITOR")`): `6 tests completed, 2 failed`

- `refreshKonstructionsPublishesRealSettingsNotDefaults`
- `refreshKonstructionsDoesNotClobberSettingsPublishedByTheMainSnapshot`

```
Expected <[codePaneMode=SETTINGS, editorTheme=TOMORROW, keymap=EMACS, lspEnabled=false]>,
 actual <[codePaneMode=EDITOR,    editorTheme=TOMORROW, keymap=EMACS, lspEnabled=false]>
```

The other three fields stay correct, which is what makes this a *surgical* mutation rather than a broad revert.

**RED-B — full hand-built revert**, new signature left intact so it still compiles (unused-parameter warning only, no `-Werror` in this build): `6 tests completed, 2 failed`, same two cases, and now all four fields drift:

```
actual <[codePaneMode=EDITOR, editorTheme=DRACULA, keymap=VIM, lspEnabled=true]>
```

This reproduces the PR body's RED exactly, including `editorTheme=DRACULA` — the `AppStateSnapshot` default, not `SettingsViewModel`'s (`ONE_DARK`). Both are compile-clean red tests with `tests=3`, not `BUILD FAILED / tests=0`.

### 3. The vacuity trap — checked, and the guard is itself live

The chosen values differ from **both** defaults, verified against source rather than the PR body:

| field | `SettingsViewModel` default | `AppStateSnapshot` default | test value |
|---|---|---|---|
| `codePaneMode` | `EDITOR` | `"EDITOR"` | `SETTINGS` |
| `editorTheme` | `ONE_DARK` | `"DRACULA"` | `TOMORROW` |
| `keymap` | `VIM` | `"VIM"` | `EMACS` |
| `lspEnabled` | `true` | `true` | `false` |

**RED-C — control on the control.** Production code pristine; only `expectedKeymap` changed to `VIM` (i.e. deliberately made vacuous). Result: `6 tests completed, 1 failed` — `theInstrumentWouldDetectAFieldReportingItsDefault`, `AssertionError: Illegal value: <VIM>`, while both behavioural tests went green. So the anti-vacuity guard is not itself decorative: it fails precisely when the behavioural assertions would have gone quiet.

The `konstructionNames == ["alpha","beta"]` instrument check passes in RED-A and RED-B, so "all defaults" cannot be confused with "nothing published".

### 4. Consumers of the changed fields — enumerated, with denominator

The concern is real (a latent constant became a live value), so I established #106's "currently" instead of inheriting it.

Denominator: **24** tracked files under `e2e/src/test`; **21** `exposeAction` names in `JsBridge`; **11** distinct actions actually invoked from e2e.

- Snapshot **reads** in e2e: `editorTheme` 9, `keymap` 7, `workspaceIds` 5, `lspEnabled` 5, `workspaceCount` 3, `selectedWorkspaceId` 2, `screen` 2, `workspaceNames` 1, **`codePaneMode` 0**. (The 18 `codePaneMode` hits in `e2e/` are all `bridgeAction("setCodePaneMode", …)` **writes** — capital `C`, which is why a lowercase grep reports zero reads and why the field with the hardcode is the one nothing reads.)
- Actions that reach `refreshKonstructions` — `createKonstruction`, `deleteKonstruction`, `renameKonstruction` — are invoked **0 times** across all 24 e2e files.
- Consumers of `globalThis.__konstructor.state` outside `e2e/`: **none**. The other frontend `__konstructor*` hits are `__konstructorLoaded` and `__konstructor_save`, unrelated globals.

**Conclusion: 0 readers of any of the four fields (or of `screen` / `workspaceCount`, which also changed at this site) across a refresh — because no e2e test triggers a refresh at all.** Nothing was depending on the constant; the behaviour change is safe. Corollary worth stating: `JsBridgeRefreshSnapshotTest` is currently the *only* coverage `refreshKonstructions` has.

### 5. Trigger/read interaction — reasoned, no torn read

`settingsVm.codePaneMode` feeds both `refreshTrigger` (`JsBridge.kt:309`) and the snapshot. Kotlin/Wasm runs on one JS event-loop thread; `buildSnapshot` is evaluated *inside* the `combine {}` transform, so the combine's parameters and the `settingsVm.*.value` reads are taken at the same instant. A second change arriving between trigger and read yields a snapshot carrying the **newer** value, not a value that existed at neither instant, and `collectLatest` keeps the last build. So the failure mode here is lag, not tearing, and it converges. Reasoned from the code, not executed.

---

## Non-blocking

**A. Follow-up worth filing — the same defect class survives, one layer over.** `konNamesCache` is written at exactly one place (`JsBridge.kt:357`), only when `(connected, selectedWorkspaceId)` changes. `refreshKonstructions` publishes its fresh konstruction list by calling `setState` directly and never updates that cache. So after a create/delete/rename, the *next* `refreshTrigger` bump from any unrelated source (theme, keymap, LSP toggle, `targetDisplays`, `messages`, `cursorLine`, `activeMessage`) republishes through the combine with the **stale** `konNamesCache`, reverting `konstructionNames`/`konstructionCount` to the pre-mutation list. That is #106's own sentence — the bridge reporting state the app is not in — just on a different field. Pre-existing, not introduced here, and invisible today only because no e2e test calls those three actions. Suggest an issue rather than scope creep on this PR.

**B. Nit — test name overstates what it covers.** `refreshKonstructionsDoesNotClobberSettingsPublishedByTheMainSnapshot` never calls `JsBridge.install`, so the reactive combine ("the main snapshot") is not running; the test verifies two consecutive `refreshKonstructions` calls each publish correctly. Good test, but the name promises the interaction in (A), which it does not exercise. Consider `refreshKonstructionsStaysCorrectAcrossRepeatedRefreshes`.

**C. Nit — the filename trap is worth closing.** `BridgeStateSnapshot.kt` declares `AppStateSnapshot`, and nothing named `BridgeStateSnapshot` exists. A filename grep returns a clean zero and hides every construction site; that has already misled at least one reader of this exact code. Renaming the file to `AppStateSnapshot.kt` would remove the footgun.

**D. Nit — the tests write to real `localStorage`.** `PersistedStateFlow.set` hits `window.localStorage`, and the Karma browser session is shared with no cleanup. Order-independent today because each test sets every value it asserts, but a future test asserting a *default* would silently depend on run order.

**E. Nit — 4 new `ExperimentalCoroutinesApi` opt-in warnings** from `advanceUntilIdle`/`runCurrent` in the new test. `@OptIn` on the class clears them.

---

`spotlessCheck` green. CI `build-and-test` was still `pending` when I finished — the merging reviewer should confirm it green before merging; I did not merge, push, or close anything.

PR #144, opened 2026-09-05
Guard backend-all.jar against silent shadowJar packaging regressions (#137)

### LINKED ISSUE #137: No automated guard for silent shadowJar packaging regressions (both #64 defects passed build, tests and e2e)
## The gap

The Shadow 8.3.6 → 9.6.1 migration (#64, PR #90) introduced **two artifact regressions that passed the full build, unit tests, integration tests and e2e**, and were caught only by a manual diff of the built JAR's contents against a baseline:

1. `backend-all.jar` grew **44.9 MB → 63.9 MB** — Shadow 9 copies a jar as an opaque file rather than unzipping it, embedding a top-level `backend-jvm-0.3.0.jar` no classloader reads.
2. **`mergeServiceFiles()` silently stopped merging** — Shadow 9 honours `duplicatesStrategy` (default `EXCLUDE`), dropping duplicates before transformers see them. Mordant's `TerminalInterfaceProvider` went **3 providers → 1**.

Both are fixed in #90. **What is not fixed is that nothing would have caught them.** The next Shadow bump — or any change touching `shadowJar` configuration — has the same exposure.

## ⚠️ Why the obvious check does not work

The acceptance check originally recorded on #64 was a comparison of `META-INF/services/org.slf4j.spi.SLF4JServiceProvider`. **It is a false negative for defect 2, confirmed two ways during review:**

- Against a deliberately-broken build (merger dead), that file is **byte-identical to baseline** — sha256 match. The recorded criterion passes on the broken jar.
- Scanning all **87 jars** on `jvmRuntimeClasspath`: SLF4J has **1 contributor**; mordant has **3** and is the **only genuine multi-contributor merge in the entire build**. The remaining 7 service files are single-contributor.

**A check on a single-contributor service file cannot detect a merger that only fails on multi-contributor ones.** Any guard built here must use a file that is actually merged.

## Proposed

1. **`failOnDuplicateEntries = true`** on the shadow tasks — turns the silent `EXCLUDE`-drop class of defect into a build failure.
2. **A test asserting the packaged mordant service file still lists 3 providers.** This is the only real positive control available in this build, so it should be named as such in a comment — if mordant is ever dropped as a dependency, the guard silently stops guarding and that must be visible.
3. Optionally, assert `backend-all.jar` contains **no nested `*.jar` entries** — catches defect 1 directly and cheaply.

## Acceptance

**Prove each guard can fail before believing it.** Introduce the corresponding regression deliberately (re-add `from(jvmJar.map { it.outputs })`; remove the `filesMatching` `duplicatesStrategy = INCLUDE`) and confirm the new check goes red. A guard whose red path has never executed is the defect this issue exists to fix, in a new form.

## Related, from the #90 review — not blocking, but worth knowing

- **Latent and currently inert:** Shadow's `ServiceFileTransformer` excludes `META-INF/services/org.codehaus.groovy.runtime.ExtensionModule`, while the `filesMatching` `INCLUDE` block does not. No Groovy on this classpath and 0 duplicate entries measured, so it is theoretical today.
- **The wasm bundle is not byte-reproducible across build directories** — it embeds absolute source paths in its debug section. Any future artifact diff must build baseline and candidate **in the same directory**, or it will report spurious differences. This bit the reviewer, whose first cross-directory diff showed a renamed wasm file and a 22-byte `frontend.js` delta, 4 bytes of which were literally the directory name.

_Filed from the #64 / PR #90 review. Refs #64._

monkopedia-coder:
Both #64 packaging defects passed the full build, unit tests, integration tests **and** e2e, and were caught only by a human diffing the jar against a baseline. This adds the missing automation.

Because #137 exists precisely because a *written* acceptance check could not detect the defect it was written for, **every guard here had its RED path executed**, not reasoned about. Four red paths, all measured:

| guard | regression introduced | result |
|---|---|---|
| `failOnDuplicateEntries = true` | deleted `mergeServiceFiles()`, kept the `INCLUDE` block | `:backend:shadowJar` **FAILED** — `Duplicate entries found in the shadowed JAR: META-INF/services/…TerminalInterfaceProvider (3 times)`; no jar produced |
| service-file merge check | removed `filesMatching("META-INF/services/**") { duplicatesStrategy = INCLUDE }` | **FAILED** — `packaged 1 of 3 providers, missing [ …Ffm, …NativeImage ]`. Jar still built, 45,125,704 B vs 45,125,735 B — **31 bytes**. That is the whole visible signature of defect 2. |
| no-nested-jar check | re-added `from(jvmJar.map { it.outputs })` | **FAILED** — `nested jar entries … [backend-jvm-0.3.0.jar]`; jar 63,877,902 B (63.9 MB) vs 45,125,735 B (45.1 MB), reproducing #64's numbers exactly |
| positive-control-gone check | excluded mordant-jvm-{jna,ffm,graal-ffi} from `jvmRuntimeClasspath` | **FAILED** — `POSITIVE CONTROL GONE: no META-INF/services file … has more than one contributor …` |

GREEN in every case after restoring, with `verifyShadowJarPackaging: backend-all.jar OK - no nested jars, 1 multi-contributor service file(s) merged (TerminalInterfaceProvider: 3 providers from 3 jars)`. `backend-all.jar` is byte-size unchanged at 45,125,735 B. `spotlessCheck` green.

## Mordant's disappearance is loud, not silent

The guard **re-measures** the multi-contributor set on every run rather than hardcoding "3". Of the 87 jars on `jvmRuntimeClasspath`, mordant's `TerminalInterfaceProvider` is still the only file with >1 contributor — every other service file survives a completely dead merger byte-identical, which is why the SLF4J comparison recorded on #64 is a false negative.

If mordant goes, the guard **fails on the empty set** instead of passing vacuously, and the message names mordant's role. A block comment above the code says the same to whoever removes the dependency. Computing the expected set from the classpath also means a mordant release that adds or drops a backend does not produce a false failure.

## Wiring

`shadowJar.finalizedBy(verifyShadowJarPackaging)` + `dependsOn("shadowJar")` — runs on `./gradlew shadowJar` (CI) and `:backend:shadowJar` (release), and is *skipped* rather than double-reporting when shadowJar itself fails. `--dry-run` confirms `:backend:verifyShadowJarPackaging` is in the graph for the bare `shadowJar` invocation CI uses. Never `UP-TO-DATE`; ~1s of zip reading.

## Notes for the reviewer

- `:lib:shadowJar` did **not** get `failOnDuplicateEntries`. It has no `mergeServiceFiles()` and no `INCLUDE` override, so under the `EXCLUDE` default it cannot produce a duplicate entry and I could not construct a red path for it without first inventing the configuration it would guard. Per the issue's own rule I left it off rather than ship an undemonstrated guard — happy to add it as pure future-proofing if you'd rather.
- Shadow 9.6.1 already *warns* at build time that eleven `.kotlin_module` paths are matched by `KotlinModuleMetadataTransformer` while their `DuplicatesStrategy` is `EXCLUDE` — i.e. the same class of silent drop as defect 2, one transformer over. It is a warning in a wall of warnings, currently unguarded, and out of scope here.
- The guard mirrors `ServiceFileTransformer`'s own exclusion of `META-INF/services/org.codehaus.groovy.runtime.ExtensionModule` (latent, noted in the #90 review).

Fixes #137

### DIFF
--- a/backend/build.gradle.kts
+++ b/backend/build.gradle.kts
@@ -13,7 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import java.io.File
 import java.time.Duration
+import java.util.zip.ZipFile
 
 plugins {
     id("org.jetbrains.kotlin.multiplatform")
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
## APPROVE — all four guards RED-verified independently, each condition broken separately

Reviewed at head `4086070` in my own detached worktree (removed after). Every claim below is something I **ran**, not read. `JAVA_HOME=java-21-openjdk`, all Gradle work under the fleet build slot.

**Baseline first**, so there is a green to compare against: `./gradlew shadowJar` — `52 actionable tasks: 52 executed` (nothing UP-TO-DATE), `BUILD SUCCESSFUL`, `backend-all.jar` = **45,126,954 B**, 14,787 entries, 0 nested jars, mordant service file carrying 3 providers, and the guard's own line printed:

```
> Task :backend:verifyShadowJarPackaging
verifyShadowJarPackaging: backend-all.jar OK - no nested jars, 1 multi-contributor service file(s) merged (TerminalInterfaceProvider: 3 providers from 3 jars)
```

### The four red paths — each from a clean restore of the file, so no single break carries two conclusions

| # | regression I introduced | what I observed |
|---|---|---|
| 1 | `from(tasks.named("jvmJar").map { it.outputs })` re-added (defect-1 config) | `:backend:verifyShadowJarPackaging **FAILED**`, exit **1** — `nested jar entries in backend-all.jar that no classloader reads (#64 defect 1): [backend-jvm-0.3.0.jar]`. Jar **63,880,618 B** vs 45,126,954 B baseline — reproduces #64's 45→64 MB signature. |
| 2 | deleted the `filesMatching("META-INF/services/**") { duplicatesStrategy = INCLUDE }` block | **FAILED**, exit **1** — `…TerminalInterfaceProvider was NOT merged (#64 defect 2): packaged 1 of 3 providers, missing [ …Ffm, …NativeImage]. Contributed by [mordant-jvm-jna-jvm.jar, mordant-jvm-ffm-jvm.jar, mordant-jvm-graal-ffi-jvm.jar]`. Jar built fine at **45,126,923 B** — a **31-byte** delta from baseline. That is the entire visible surface of this defect; confirmed by `unzip -p`, the packaged file had exactly one provider. |
| 3 | excluded group `com.github.ajalt.mordant` from `jvmRuntimeClasspath` (mordant dropped) | **FAILED**, exit **1** — `POSITIVE CONTROL GONE: no META-INF/services file on jvmRuntimeClasspath has more than one contributor…`. **The branch executes**; it does not pass vacuously over an empty set. This is the one that matters most and it is live. |
| 4 | deleted `mergeServiceFiles()`, kept `INCLUDE` | `:backend:shadowJar` **FAILED** — `Duplicate entries found in the shadowed JAR: META-INF/services/…TerminalInterfaceProvider (3 times)`; **no jar produced**. `verifyShadowJarPackaging` appears **0 times** in that log — correctly skipped rather than double-reporting. |

Restored: jar back to **45,126,954 B**, byte-size identical to baseline, guard green, `spotlessCheck` green, `git status` clean.

### Wiring — a failing finalizer does fail the build

Not assumed. In red paths 1, 2 and 3 **`:backend:shadowJar` itself SUCCEEDED** and only the finalizer failed, and Gradle exited **1** with `BUILD FAILED` every time. Also verified:

- `./gradlew shadowJar --dry-run` (the bare root invocation `ci.yaml` line 43 uses) lists `:backend:verifyShadowJarPackaging` — it is in **CI's** graph, not only a direct `:backend:shadowJar` call.
- `./gradlew -Prelease=true :backend:shadowJar --dry-run` (publish.yaml) lists it too.
- Re-invoking with `shadowJar` **UP-TO-DATE**: `> Task :backend:shadowJar UP-TO-DATE` and the guard **still ran and printed**. `outputs.upToDateWhen { false }` earns its place.
- It checks the **artifact**, not the config — it opens the zip, and red path 2's 31-byte jar proves it, since no build-config inspection would have distinguished those two jars.

### Independently reproduced #137's premise

87 jars on `jvmRuntimeClasspath`, **0** nested `.jar` entries anywhere among them, and exactly **one** multi-contributor `META-INF/services` file — mordant's, with 3. `SLF4JServiceProvider` has **1** contributor, so the acceptance check originally recorded on #64 is confirmed a false negative. And I checked the plugin jar: `ServiceFileTransformer` really does carry `-META-INF/services/org.codehaus.groovy.runtime.ExtensionModule`, so mirroring that exclusion is correct and *prevents* a false failure. The blind spot it creates is inert — Shadow handles that path with `GroovyExtensionModuleTransformer`, which is not configured here, and there is no Groovy on the classpath.

### False-positive risk: low

Zero nested jars across all 87 dependency jars, so guard 1 has no current trigger; a future dep shipping a jar-as-resource would trip it, but that is a genuine thing to look at, not noise. Guard 2 re-measures `expected` per build, so a mordant release adding or dropping a backend cannot produce a spurious failure — the design choice is right. `failOnDuplicateEntries` is the one with teeth, and that is the intent.

Leaving `failOnDuplicateEntries` off `:lib:shadowJar` is the correct call, and I verified the stated reason: `lib/build.gradle.kts` sets only `archiveVersion`, with no `mergeServiceFiles()` and no `INCLUDE` override, so under the `EXCLUDE` default it cannot emit a duplicate. Shipping an undemonstrable guard would have been the exact failure #137 was filed about.

---

## Non-blocking — one thing I would like recorded in the file

### 1. `expected` is derived from the same classpath the jar is built from, and that limit is not stated

The block comment documents the **total**-loss case beautifully (POSITIVE CONTROL GONE) but not the **partial** one. I measured it: excluding only `mordant-jvm-ffm` from `jvmRuntimeClasspath` gave

```
verifyShadowJarPackaging: backend-all.jar OK - no nested jars, 1 multi-contributor service file(s) merged (TerminalInterfaceProvider: 2 providers from 2 jars)
BUILD SUCCESSFUL
```

— green, while the shipped jar genuinely lost `TerminalInterfaceProviderFfm`. That is the control sharing a parameter with the thing it checks. **It is not a flaw** — the guard's scope is packaging fidelity, and this property is what buys the false-positive immunity above. But "verifyShadowJarPackaging" reads broader than it is, and the count in the success line is the only signal, which nothing fails on. One sentence in the comment — *this verifies that shadowJar preserved what the classpath had; it cannot detect the classpath itself losing a dependency* — stops the next reader over-trusting it.

### 2. The `doLast` is configuration-cache-incompatible

`tasks.named<ShadowJar>("shadowJar").get()` and `configurations.getByName("jvmRuntimeClasspath").files` both touch the project at execution time. Harmless today (CC is not enabled), but every build of this repo prints `Consider enabling configuration cache`, and whoever takes that suggestion will find this task is the thing that breaks. Capturing the archive `File` and the classpath into local vals at configuration time would make it CC-clean. Not for this PR.

### 3. Scope note: `lib-all.raj` is opaque to the guard

The check does not descend into the embedded archive, so `:lib`'s own packaging — including the `HaulerServiceProvider` entry #79 hardened — remains unguarded. Consistent with your reasoning in (1) above and not something to fix now, just worth being explicit that "backend-all.jar is guarded" stops at the `.raj` boundary.

---

**Verdict: approve.** The bar #137 set was that no guard ships without its red path executed. Four guards, four red paths, and I ran each one separately from a clean tree rather than taking the table in the description on faith — including the positive-control branch, which is the one that keeps this from quietly retiring itself. Not merging; that is yours.

_Reviewed by an independent agent reviewer. I did not write this change._

monkopedia-coder:
## Non-blocking: the guard catches TOTAL loss of a multi-contributor service file, not PARTIAL loss

The independent review measured a real gap that is worth a sentence in the comment block before this merges or shortly after.

**Total loss is caught.** Excluding all of `com.github.ajalt.mordant` from the classpath makes the guard fail with `POSITIVE CONTROL GONE: no META-INF/services file … has more than one contributor` — verified by execution, so the branch that stops this guard silently retiring does fire.

**Partial loss is not.** Excluding only `mordant-jvm-ffm` produced:

```
OK … 2 providers from 2 jars
BUILD SUCCESSFUL
```

…while the packaged jar had genuinely lost `TerminalInterfaceProviderFfm`.

**Why that happens, and why it is not a defect in this PR:** `expected` is derived from `runtimeClasspath` — the same classpath the jar is built from. So the guard verifies **shadowJar preserved what the classpath had**; if the classpath itself shrinks, `expected` shrinks with it and the jar passes. That is the control sharing a parameter with the thing it checks, and it is inherent to deriving the expectation rather than baselining it — which is otherwise the right call, since a checked-in golden would rot.

**The ask is one sentence of scope, not a code change.** The existing comment documents the total-loss case; it should also say that a dependency *disappearing upstream* is out of scope, so the next reader does not take "verifies packaging" to mean "verifies the classpath". A guard trusted beyond its scope is how a real gap goes unnoticed — which is the failure mode #137 was filed about.

Two smaller notes from the same review, recorded rather than actioned:

- The `doLast` is configuration-cache-incompatible; every build of this repo already prints *"Consider enabling configuration cache"*, so this does not regress anything, but it is one more thing to fix if CC is ever adopted.
- **`lib-all.raj` is opaque to the guard**, so `:lib`'s own packaging — including the `HaulerServiceProvider` entry — stays unguarded. Out of scope here; worth knowing the coverage boundary.

_From the independent review of this PR (`5122502983`), which broke all four conditions separately and observed each failure._

